PR #17, opened 2015-11-22
Circle CI integration

dennybaa:
st2-packages migrated from drone to Circle CI:

Current features:
 * Just a single build and test of debian wheezy.
 * Clear and easy CI scripts.
 * Usage of build-in services: rabbitmq, mongo, postgress
 * Pip and wheels cache volume pass through into docker-compose pipeline.


circle.yml
```diff
@@ -0,0 +1,18 @@
+dependencies: 
+  cache_directories: 
+    - ~/.cache/pip
+    - ~/wheelhouse
+  post: 
+    - .circle/services.sh
+    - ".circle/compose.sh run wheezy /bin/true"
```
armab: Not clear what's happening here via `/bin/true`

circle.yml
```diff
@@ -0,0 +1,18 @@
+dependencies: 
+  cache_directories: 
+    - ~/.cache/pip
+    - ~/wheelhouse
+  post: 
+    - .circle/services.sh
```
armab: `database` configuration directive is suitable for this, see: https://circleci.com/docs/configuration#database

```yaml
database:
  override:
    - ...
```

circle.yml
```diff
@@ -0,0 +1,18 @@
...
+  pre: 
+    - .circle/cache-permissions.sh
+
+machine: 
+  services: 
+    - docker
+
+test:
+  override:
+    - ".circle/compose.sh run wheezy bash -c 'cp /root/Gemfile* ./ && bundle exec rspec'"
```
armab: It could be more useful to have something like: 
`docker-compose -f compose.yml -f compose.circle.yml run wheezy` without wrapper wrapper script.

Is it possible to move `cp /root/Gemfile* ./ && bundle exec rspec` logic into `compose.circle.yml` ?

circle.yml
```diff
@@ -0,0 +1,18 @@
+dependencies: 
+  cache_directories: 
+    - ~/.cache/pip
+    - ~/wheelhouse
+  post: 
+    - .circle/services.sh
+    - ".circle/compose.sh run wheezy /bin/true"
+    - ".circle/compose.sh run wheezy"
```
armab: One `compose.sh` script doing 3 different things and launched 3 times is confusing.

The usual dependencies CI config that is more clear for someone who looks into it: 
```yaml
dependencies:
  cache_directories:
    - ~/.cache/pip
    - ~/wheelhouse
  override:
    - sudo pip install wheel
    - sudo pip install docker-compose
```

Easier to follow the logic, instead of digging into many bash scripts. Also once you divide them as separate commands - CircleCI shows time for every command, what's good to know for debugging.

That's probably the reason why you can see something like: 
https://github.com/ruby/ruby/blob/trunk/.travis.yml

armab:
Also I remember there is special GitHub "administrative" account st2stanley or something like that.

@jfryman can we connect this repo `stackstorm/st2-packages` to CircleCI via `st2stanley`?

circle.yml
dennybaa: Btw it's already in there just have a look inside [compose.yml](compose.yml). This is required for spliting build and test stages, though it's not really important at all (I mean this cp string :) ).

circle.yml
dennybaa: It's simple it's used for pull stage separation.

circle.yml
dennybaa: I doubt about this. Basically the lines such as  
```
    - sudo pip install wheel
    - sudo pip install docker-compose
```
Don't have any importance, and what they do is just cluttering UI of CircleCI. Less commands in `circle.yml` more easy to navigate between its steps.

PR #21, opened 2015-11-25
Advanced Circleci

armab:
Pull changes from https://github.com/armab/st2/blob/master/circle.yml

Make sure it still works.


circle.yml
```diff
@@ -1,18 +1,71 @@
...
+    DEPLOY_DOCKER: 0
+    DEPLOY_PACKAGES: 0
+  pre:
+    # Paswordless SSH between parallel build nodes
+    - sudo cp /home/ubuntu/.ssh/config /root/.ssh/config
+    # Need latest Docker version for some features to work (CircleCI by default works with outdated version)
+    - |
+      sudo curl -L -o /usr/bin/docker 'http://s3-external-1.amazonaws.com/circle-downloads/docker-1.9.0-circleci'
+      sudo chmod 0755 /usr/bin/docker
+      sudo service docker start
```
dennybaa: What's the goal of the newer docker, do have any required feature?

circle.yml
```diff
@@ -1,18 +1,71 @@
...
+  environment:
+    ST2_GITURL: https://github.com/StackStorm/st2
+    ST2_GITREV: master
+    ST2_DOCKERFILES_REPO: https://github.com/armab/st2-dockerfiles
+    BUILD_DOCKER: 0
+    DEPLOY_DOCKER: 0
+    DEPLOY_PACKAGES: 0
+  pre:
+    # Paswordless SSH between parallel build nodes
+    - sudo cp /home/ubuntu/.ssh/config /root/.ssh/config
```
dennybaa: Why do we need interconnect between these nodes?

circle.yml
armab: I need latest Docker for
> docker build `--build-arg`

to work.

circle.yml
armab: Each group of artifacts for distros are available at `0` `1` `2` `3` machines separately.
`deployment` section works only once at `0` machine.

We'll need to get artifacts from `1` `2` `3` machines to upload them from `0` machine.

dennybaa: Is not that cheaper to use volumes?

armab: What volumes do you mean?

dennybaa: docker volumes

armab: `0` `1` `2` `3` machines in parallel matrix builds - they're different nodes that have connection between each other only via SSH.

.circle/vars.sh
```diff
@@ -0,0 +1,23 @@
...
+set +x
+echo "export DISTROS=(${DISTROS[*]})" >> ~/.circlerc
+echo "export DISTRO=${DISTRO}" >> ~/.circlerc
+echo "export ST2PKG_VERSION=${ST2PKG_VERSION}" >> ~/.circlerc
+echo "export ST2PKG_RELEASE=${ST2PKG_RELEASE}" >> ~/.circlerc
+#    - echo ST2PKG_VERSION=$(python -c 'execfile("../st2common/st2common/__init__.py"); print __version__') >> ~/.circlerc
+#    - echo "ST2PKG_RELEASE=$(scripts/bintray.sh next-revision ${DISTRO} ${ST2PKG_VERSION})" >> ~/.circlerc
+
+set -x
+cat ~/.circlerc
```
dennybaa: Here I would suggest not to use ~/.circlecrc at all.
The more clean approach is to use `~/.circle/env` file, which is gonna be sourced by our scripts.

circle.yml
```diff
@@ -1,18 +1,71 @@
...
+  artifacts:
+    - /tmp/st2-packages
+
+machine:
+  # Overwrite these ENV variables in parametrized (manual/API) builds
+  environment:
+    ST2_GITURL: https://github.com/StackStorm/st2
+    ST2_GITREV: master
+    ST2_DOCKERFILES_REPO: https://github.com/armab/st2-dockerfiles
+    BUILD_DOCKER: 0
```
armab: This temporary disables:
* `st2*` Docker images builds from `st2-dockerfiles`
* deploying created `st2*` Docker containers to Docker Hub
* deploying created packages to Bintray

I'll need all that later.
For now focus on builds.

.circle/vars.sh
armab: More logic to handle.

I think `.circlerc` does that job well. Also it makes possible clear & easy 1 command:

```yaml
- .circle/docker-compose.sh test ${DISTRO}
```
 without sourcing anything.

armab:
Ok, polished everything & squashed all 30+ testing CircleCI commits into 2 :)

The workflow looks much more clear/readable from single `circle.yml` file,
without digging too much into bash files, as you @dennybaa requested.

Let me know if you have any suggestions.

circle.yml
```diff
@@ -1,18 +1,71 @@
...
+    - .circle/docker-compose.sh test ${DISTRO}:
+        parallel: true
+  post:
+    - .circle/docker.sh build st2bundle
+    - .circle/docker.sh build st2actionrunner st2api st2auth st2exporter st2notifier st2resultstracker st2rulesengine st2sensorcontainer
+    - .circle/docker.sh run st2api
+    - .circle/docker.sh test st2api 'st2 --version'
+
+deployment:
+  bintray:
```
armab: bintray `deployment` section as-is for now/disabled.
We'll need to decide about repository naming first, but that's later.

At least I have everything synced.

.circle/vars.sh
armab: Also since 
```yaml
machine:
  # Overwrite these ENV variables in parametrized (manual/API) builds
  environment:
    ST2_GITREV: master
    ST2_DOCKERFILES_REPO: https://github.com/armab/st2-dockerfiles
    BUILD_DOCKER: 0
    DEPLOY_DOCKER: 0
    DEPLOY_PACKAGES: 0

Comment (dennybaa) [inline .circle/vars.sh]: it's not absolutely. it's logic separation, you clearly see what your pipeline uses, and not messing it up with that env from .circlerc.
And then doing all that inspection to stdout, it's more simple to look through 5 vars in one file rather then listing through tons of useless output.

Comment (armab) [inline .circle/vars.sh]: This is totally fine: 
kssfzu3 1

CircleCI env variables: https://circleci.com/docs/environment-variables and our variables combined together and the most important *sourced by default*.

The fact is that they're already combined, eg all we define in `machine:environment` already goes in that `.circlerc`.

-----

This is nice and works by default: 
```yaml
dependencies:
  override:
    - .circle/docker-compose.sh pull ${DISTRO}:
        parallel: true
  post:
    - .circle/docker-compose.sh build ${DISTRO}:
        parallel: true

test:
  override:
    - .circle/docker-compose.sh test ${DISTRO}:
        parallel: true
```

comparing to ugly way with separated file for custom variables (how I actually did before): 
```yaml
dependencies:
  override:
    - . ~/env; .circle/docker-compose.sh pull ${DISTRO}:
        parallel: true
  post:
    - . ~/env; .circle/docker-compose.sh build ${DISTRO}:
        parallel: true

test:
  override:
    - . ~/env; .circle/docker-compose.sh test ${DISTRO}:
        parallel: true
```

And don't forget that we're not only two guys who will work with this in future.
Less details and corner cases to note/remember - easier for others to use/read.

circle.yml
armab: @dennybaa
Yeah man, it was a new Docker.

This is rspec tests stage: 
[ubl52ls 1](https://circleci.com/gh/StackStorm/st2-packages/82)

Doesn't looks like hard to fix, probably adjusting IP/networking related to https://circleci.com/docs/docker#connecting-to-services-outside-of-the-container

Will need time to spend/debug it later when we'll reach stage with docker build/deploy.
Reverting original outdated CircleCI Docker version for now, since deploy stage is inactive for now.

armab:
Finally all is green, merging! :green_heart: 

A few notes:
* Added better insight about https://github.com/StackStorm/st2-packages/issues/23 bug.
* I've adjusted rake parallelism to `x9` (number of packages we build). By launching more processes hopefully we'll get better use of I/O. Shouldn't hurt, and *seems like* `0.5min` faster build.

PR #24, opened 2015-11-26
CircleCI Cosmetic Adjustments

armab:
Cosmetic changes

* [x] Remove `parallel: true` for non-test section
* [x] Show CircleCI services that we use explicitly (mongo, rabbit, postgres)
* [x] Tests with rake `j4`/`j9` parallel flag
* [x] Pass ENV variables from CI to Dockerized environment

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -43,11 +43,9 @@ dependencies:
     - docker-compose version
     - docker version
   override:
-    - .circle/docker-compose.sh pull ${DISTRO}:
-        parallel: true
+    - .circle/docker-compose.sh pull ${DISTRO}
   post:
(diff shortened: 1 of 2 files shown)


dennybaa:
Yeah that's better. Don't merge it please so far, maybe you can play with -j4 and without it at all.
It will be interesting to look at the result. I guess 4 is enough with nproc=2, this 2 proc excess might give enough space for performance increase, and if not removing -j will give excess of 4 procs.

But `-j9` is definitely too much for this scenario.

dennybaa:
Nice job @armab. Thank you. Merged!

PR #25, opened 2015-11-26
Mistral fixes, --no-index fixes and various improvements

dennybaa:


dennybaa:
This fixes resolve issues with centos7, so I'm merging it. However systemd container is not operational under CircleCI #31 .

Issue #26, opened 2015-11-26
Using is wheelhouse with later `--no-index` is dangerous


dennybaa:
When we do pre build of  wheels in some shared particular directory we can run into race condition.
Such as we do with **pecan** dependency.
When we populate the shared wheelhouse, both mistral and another st2package create wheels one for version `1.0.3` and another `0.7.0`.

Later on when package is built without `.cache/pip` using **--no-index** option pip doesn't rely on full dependency scanning (it can't parse version for git+https dependencies), so it chooses the latest...

dennybaa:
@Kami Hi there, have a look at https://circleci.com/gh/StackStorm/st2-packages/118 please.
Should we add logshipper dependency into st2reactor `in-requirements.txt`?

And the travis build as well, strange `git clone -q https://github.com/Kami/logshipper.git` failer.
I think it's a github issue (sometimes happens :) So would you please re-run travis.

Thanks!

dennybaa:
Also would you please help with fixing this:

```
[package: st2actions]        Could not find a version that satisfies the requirement logshipper (from st2actions==1.2.dev0) (from versions: )
[package: st2actions]        Some externally hosted files were ignored as access to them may be unreliable (use --allow-external logshipper to allow).
[package: st2actions]      No matching distribution found for logshipper (from st2actions==1.2.dev0)
```

I've got no idea why it complains, and asks to use `--alow-external`.

dennybaa:
There has just been errors in build mechanism which resulted in these kind of problems. Namely `--no-index` is just not used anymore, so dependencies are handled correctly.

Issue #12, opened 2015-11-03
Pass $ST2_RELEASE to docker-compose from CI


armab:
Add ability to pass revision number aka release number from CI, to control package versioning.
> st2api_0.14dev-**20**_amd64.deb

So the version number `0.14dev` is parsed from [`st2common`](https://github.com/StackStorm/st2/blob/master/st2common/st2common/__init__.py#L16) automatically and revision number `20` is injected from outside via ENV.

Same as we have for mistral [in docker-compose](https://github.com/dennybaa/st2-packages/blob/4c356bf7b908fd27bdfdfa865cb93b326985f94a/compose-suite.yml#L19).

`docker-compose` should pick up env via https://github.com/docker/compose/pull/1765 when using:
```sh
export ST2_RELEASE=20
docker-compose up debian
```
--- 

^^ I'm happy that feature with ENV variables was introduced in docker-compose and released in [1.5 version just *hours ago*](https://github.com/docker/compose/releases/tag/1.5.0). We're lucky here :four_leaf_clover: 

Docs: https://github.com/docker/compose/blob/8cc8e61/docs/compose-file.md#variable-substitution

dennybaa:
:+1:

armab:
Worked flawlessly in https://github.com/armab/st2/blob/5736f16a58dbfdc9b30eafb5dc0b959c98f096a9/circle.yml#L51

Thanks to Docker :+1:

PR #27, opened 2015-11-28
Changing to use default ~/.cache/pip instead of wheelhouse with --no-index

dennybaa:


compose.yml
```diff
@@ -9,10 +9,12 @@ suite:
               bundle exec rspec"
   environment:
     - DEBUG_LEVEL=0
-    - ST2_GITURL=${ST2_GITURL}
-    - ST2_GITREV=${ST2_GITREV}
-    - ST2PKG_VERSION=${ST2PKG_VERSION}
-    - ST2PKG_RELEASE=${ST2PKG_RELEASE}
+    ## For the following variables default values are used
+    #  unless explicitly defined:
+    # - ST2_GITURL=
```
armab: Why commented?
I need these vars to be passed from CI to Docker environment.

See: 
https://github.com/StackStorm/st2-packages/issues/12
https://github.com/StackStorm/st2-packages/pull/24#commitcomment-14646393

dennybaa: I know. I've commented them because they generate too many unnecessary warnings. At the moment you don't need it, cuz there's no full implementation yet. So for now we live them commented.

dennybaa: There are many ways to pass them in, also see `docker-compose run --help` for `-e` option. Unless decided what's the best approach, I suggest to leave them commented, since they bloat output with warnings(

dennybaa:
resolvs #26

dennybaa:
this won't be resolved we stay with wheelhouse

PR #28, opened 2015-11-30
Better defaults for non-CI docker-compose run

armab:
* [x] Pass env variables from CI to `docker-compose` via CLI (avoid warnings for users)
* [x] Better defaults for non-CI run

From user side just run `docker-compose run`, instead of hard to write and remember:
`docker-compose -f compose.yml -f docker-compose.yml run`.

See https://docs.docker.com/compose/reference/docker-compose/

### DIFF
--- a/.circle/docker-compose.sh
+++ b/.circle/docker-compose.sh
@@ -19,7 +19,7 @@ case "$1" in
   ;;
   build)
     echo Starting Packages Build for $2 ...
-    docker-compose -f compose.yml -f docker-compose.circle.yml run $2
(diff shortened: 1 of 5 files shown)


dennybaa:
@armab Awesome thank you, I'll look at a bit later. You got my concern about warnings)
Now concurrency is 4, centos7 is not working, so don't take it account build will fail because of it...

armab:
It's building in my account, so the tests should pass in PR https://circleci.com/gh/armab/st2-packages/249, however they will fail after merge (pretty strange behvior for CI though).

armab:
FYI, we'll switch back to variables in yaml with defaults after this get implemented: https://github.com/docker/compose/issues/2441

dennybaa:
If this gets implemented... No one knows when) This originating discussion is more than half year old, and it was closed... Personally I like the idea to support bash variable interpolation syntax. Let's wait and hope, but not solely rely on this.

PR #29, opened 2015-11-30
Use Docker 1.9.0 in CircleCI

armab:
Allows using latest Docker `1.9.0` in CircleCI.
Needed for generating Docker images with `--build-arg` parameter, which appeared in 1.9 version.

See also: https://discuss.circleci.com/t/docker-1-9-0-is-available/366/3


dennybaa:
:+1:

Issue #31, opened 2015-12-02
CircleCI systemd (centos7/xenial) container is not operational


dennybaa:
https://discuss.circleci.com/t/docker-with-centos7-systemd-container/885

dennybaa:
well we know. cleaning the issue board though!

armab:
This is still an issue that someone will probably hit and I think it's better to keep it for informational purposes rather then closing it, so someone won't open duplicate.

I guess we can gather more information about possible infra limitations users can face depending on environment: https://github.com/StackStorm/st2-packages/issues/154 and keep more useful info in this one.

dennybaa:
just pollution, no outcome.

armab:
What I mean - we're not only 2 guys here ;)

dennybaa:
I mean pollution is pollution, if it pops up again - okay, if not keeping it for grandsons... ) Hoping they will read it one day) Well it can be named as pollution, ambiguous pre-caution you can find tons of things to name this useless kind of tickets.

Well if you forgot there's such thing like github search https://github.com/StackStorm/st2-packages/search?q=systemd&type=Issues&utf8=%E2%9C%93,  usually people use it dig through  problem first and not trying to read through the whole list which is the on-deck...

@armab Don't pollute, keep things clean and simple...

armab:
@dennybaa closing this issue is kind of [hiding the dust under the rug](https://github.com/auchenberg/volkswagen).

Grandsons? We already had valid report: `Build fails for EL7`: https://github.com/StackStorm/st2-packages/issues/154
Which is exactly 100% the same infra limitation which exists.

Closed issue usually means `resolved`.
If we keep it open - it means `known`. And that's much useful.

dennybaa:
What are you talking about...

1. #154 what connection with this issue? It's absolutely different issue, what infra?
2. **Can you technically fix it**? I.e. implement startup of systemd containers on Circle CI. If yes please reopen and solve it indeed.
3. If you can not do the 2 (^^), what's you goal to keep it visible? 

> Closed issue usually means resolved.
Absolutely not... Haven't you met *wontfix* in whatever projects?

armab:
- Q: What these 2 issues have in common?
- A: centos7 not running properly


1.) You can't run Docker with systemd under Windows and MacOS X and probably somewhere else because of infra limitation (that issue #154)
2.) You can't run Docker with systemd under CircleCI because of security limitation (this issue #31)

---

[...]

PR #34, opened 2015-12-02
Test choosing

dennybaa:
Ability to choose where tests happen. Disabled docker pushes on master.
@armab If you still need it for testing. Just make a commit and enable them back.


.circle/vars.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 DISTROS=(wheezy jessie trusty centos7)
+TESTING=(wheezy jessie trusty)
 DISTRO=${DISTROS[$CIRCLE_NODE_INDEX]}
```
armab: Why don't just change DISTROS above?
It's easiest way.

dennybaa: Because we need packages for the required set. But don't need tests for specific distros... details here #31

PR #36, opened 2015-12-03
Show DISTRO we build and test in CircleCI explicitly

armab:
> Reverting some logic that was created before and is part of https://github.com/StackStorm/st2/blob/master/circle.yml

Show DISTRO we build and test in CircleCI explicitly (we talked about that before If you remember),
**especially** after moving `DISTROS: "wheezy jessie trusty centos7"` inside `circle.yml` config.

It's good to have understanding which commands are unrelated to distro: `sudo pip install wheel docker-compose` and which are really important.

Also as you remember there are *real* docker-compose commands like:
`docker-compose pull wheezy` `docker-compose build wheezy` `docker-compose run wheezy`.

-----
Also I need to iterate through all `$DISTROS` (previous logic, before pushes), so this variable should be an array.

### DIFF
--- a/.circle/docker-compose.sh
+++ b/.circle/docker-compose.sh
@@ -10,11 +10,11 @@
 set -e
 # Source the build environment defintion (details in buildenv.sh)
 . ~/.buildenv
-set -x
 
 # Used for `RABBITMQHOST` `POSTGRESHOST` `MONGODBHOST`, see docker-compose.override.yml
 HOST_IP=$(ifconfig docker0 | grep 'inet addr' | awk -F: '{print $2}' | awk '{print $1}')
 
+set -x
 case "$1" in
   # Perform fake command invocation, technically provides images "pull" phase.
   pull)
--- a/.circle/docker-compose.sh
+++ b/.circle/docker-compose.sh
@@ -18,7 +18,7 @@ HOST_IP=$(ifconfig docker0 | grep 'inet addr' | awk -F: '{print $2}' | awk '{pri
 case "$1" in
   # Perform fake command invocation, technically provides images "pull" phase.
   pull)
-    echo Pulling dependent Docker images for $DISTRO ...
+    echo Pulling dependent Docker images for $2 ...
     docker-compose -f docker-compose.circle.yml run \
         -e ST2_GITURL=${ST2_GITURL} \
         -e ST2_GITREV=${ST2_GITREV} \
@@ -27,10 +27,10 @@ case "$1" in
         -e RABBITMQHOST=${HOST_IP} \
         -e POSTGRESHOST=${HOST_IP} \
         -e MONGODBHOST=${HOST_IP} \
-        $DISTRO /bin/true
+        $2 /bin/true
   ;;
   build)
-    echo Starting Packages Build for $DISTRO ...
+    echo Starting Packages Build for $2 ...
     docker-compose -f docker-compose.circle.yml run \
         -e ST2_GITURL=${ST2_GITURL} \
         -e ST2_GITREV=${ST2_GITREV} \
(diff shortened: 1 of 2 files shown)


dennybaa:
Since ages shell didn't have arrays, but it's lyric) If you want to iterate in shell, it's handled as easy as this:

```shell
list="first second third"
for i in $list; do
  echo $i
done
```

Using array style doesn't scale well, and it's more easy to set string in yaml `DISTROS: "wheezy jessie ..."` it's more efficient way.

Since you are not restricted to which distros you are going to build. DISTROS can be easily passed as a parameter.

.circle/docker-compose.sh
```diff
@@ -10,15 +10,15 @@
 set -e
 # Source the build environment defintion (details in buildenv.sh)
 . ~/.buildenv
-set -x
```
dennybaa: There's already cat step of buildenv. I find it as excessive logging.

circle.yml
```diff
@@ -50,13 +50,13 @@ dependencies:
     - docker-compose version
     - docker version
   override:
-    - .circle/docker-compose.sh pull
+    - .circle/docker-compose.sh pull ${DISTRO}
   post:
-    - .circle/docker-compose.sh build
+    - .circle/docker-compose.sh build ${DISTRO}
```
dennybaa: Also what is advantage of these above? Also not clear. I do understand if the variable required in a bash singlet on a particular step. However docker-compose.sh is self sufficient with it's build env...

.circle/docker-compose.sh
dennybaa: Sorry about this. Looked wrong. You omitied HOST_IP verbosity, it's ok

armab:
```sh
DISTROS="wheezy jessie trusty centos7"
for DISTRO in $DISTROS; do                                             
  echo "element: $DISTRO"
done
```
Result:
```
element: wheezy jessie trusty centos7
```

What I'm talking about and how it was before:
```sh
DISTROS=(wheezy jessie trusty centos7)
for DISTRO in ${DISTROS[@]}; do
  echo "element: $DISTRO"
done
```
Result:
```
element: wheezy
element: jessie
element: trusty
element: centos7
```

dennybaa:
I dun no how it happens, seriously)

```
➜  st2-packages git:(centos6) ✗ cat > /tmp/a.sh
DISTROS="wheezy jessie trusty centos7"
for DISTRO in $DISTROS; do                                             
  echo "element: $DISTRO"
done
➜  st2-packages git:(centos6) ✗ bash /tmp/a.sh 
element: wheezy
element: jessie
element: trusty
element: centos7
```

dennybaa:
and 
```
➜  st2-packages git:(centos6) ✗ cat > /tmp/a.sh
DISTROS=(wheezy jessie trusty centos7)
for DISTRO in ${DISTROS[@]}; do
  echo "element: $DISTRO"
done
➜  st2-packages git:(centos6) ✗ bash /tmp/a.sh 
element: wheezy
element: jessie
element: trusty
element: centos7
```

dennybaa:
I confirm I got your result in zsh... But it's not that we should care about. It's not compatible with bash and dash 100%. 
http://slopjong.de/2012/07/02/compatibility-between-zsh-and-bash/

armab:
Ok, let me try that in Circle.

armab:
@dennybaa That worked :+1: 
Thanks, 1 problem less.

circle.yml
dennybaa: And think about this seriously, where you need to iterate over $DISTROS, np. But excessive var usage like build ${DISTRO}, test ${DISTRO} etc is really not required... again in case of docker-compose.sh. About other things i dun know. We'll see and figure it out as soon as the appear on the horizon.

armab: Yeah, I need this.
It reverts old things before direct push. There is also file renaming "refactoring" coming from that, but I'll synchronize with your other changes in `st2`.

armab:
I'll sync remaining things in `st2`.
If we won't rename files or variables - it should't break anything.

After implementing CircleCI/Bintray + st2 staging, I'll leave it for your rafactorings.

dennybaa:
Yeah, it would be nice if after implementing these features, you refactored it yourself not me :) If you can see any structure or design flaws, it's worth doing.
But actually, this bash code base for circle ci is quite small so far, and imo it would be cooler to keep it with a clean design a priori. (Keep code modular, use less conditionals, make it more readable...)

@armab You know, I have one rule for myself if there's more then ~50 lines of bash code, it shouldn't exist (should be rewritten).

The problem with bash is that as the code base grows it becomes really painful to keep it clean. Post fact refactoring of bash is the thing which personally I wouldn't like to do. The code base might become big and difficult to understand. So refactoring will be pain and quite often it happens that it's easier to re-implement from 0 rather than refactor ((

PR #40, opened 2015-12-07
fix components build regression

dennybaa:
We must not pre-built `st2*` wheel, before installation.

If we do so, wheeled component is being by pip `pip install --find-links=/tmp/wheelhouse`, where scripts shebangs occur to be `#!python`.

This breaks `dh_virtualenv` operation... So our processing order is:
1. First build package for st2 component.
2. After package is built, create bdist wheel.

### DIFF
--- a/rake/build/environment.rb
+++ b/rake/build/environment.rb
@@ -14,7 +14,7 @@ pipeopts do
 
   env     :buildnode
   env     :testnode
-  envpass :testmode, 'bundle' # components || bundle
+  envpass :testmode, 'components' # components || bundle
   envpass :basedir,  '/root'
   envpass :debug_level, 1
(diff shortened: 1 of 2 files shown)


dennybaa:
Okay. This regression is caused by `pip install --find-links=/tmp/wheelhouse .`, when pip installs current package via distutils, it doesn't change `#!python` shebang which comes from the wheel. All scripts from `setup.scripts` section end up with exactly this interrupter.

After installation of dependencies and package itself via pip, `dh_virtualenv` invokes **fix_shebangs** which doesn't expect `#!python` to be there. To fix this, we are using https://github.com/dennybaa/dh-virtualenv/commit/bf43b0ae436c8caa67f24d545df8cfca70d3cb9a.

PR #55, opened 2015-12-22
Add files for the new st2garbagecollector service

Kami:
The title says it all.

On a related note - we need to add some developer docs on how to add a new service.


Kami:
I looked over the existing files and I'm a bit confused in some places.

1. https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/Makefile#L18

Any particular reason why we don't copy all the logging configs here? We only do it for st2auth, st2api, st2actions, but not for st2notifer, st2rulesengine, st2sensorcontainer, etc.

Kami:
Please let me know if I missed some spot.

circle.yml
```diff
@@ -59,7 +59,7 @@ test:
         parallel: true
   post:
     - .circle/docker.sh build st2bundle
-    - .circle/docker.sh build st2actionrunner st2api st2auth st2exporter st2notifier st2resultstracker st2rulesengine st2sensorcontainer
+    - .circle/docker.sh build st2actionrunner st2api st2auth st2exporter st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
```
armab: So it's new service/Docker container?
You need to add new Dockerfile here too: https://github.com/StackStorm/st2-dockerfiles

Also we need new repo in Docker hub (creating manually). I did it for you: https://hub.docker.com/r/stackstorm/st2garbagecollector/

Kami: Yeah, it's a new service which is part of  st2reactor package (for now).

I will have a look, we need to make sure all this gets documented :)

armab: Don't forget to include `st2garbagecollector` here: https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L79

At the same time @Kami keep in mind that `docker build` and `docker push` commands are here temporarily, this logic should be moved to staging box eventually (eg. when all tests passed -> build and push Docker).

Kami: Good catch, added in 1df086ac093d982c4b833c936557d77aef7d5e19

armab:
Regarding packaging @dennybaa please take a look

armab:
@Kami Also one important thing I just remembered:

Everything that's included in any package should be duplicated in `st2bundle` package, see for example all init files are here: 
https://github.com/StackStorm/st2-packages/tree/master/packages/st2bundle/debian


`st2bundle` at the moment is used for building Docker.

Kami:
@armab ah, I totally missed that.

Can we automate this - aka copy the files over in some make step or something like that? Copying it manually for sure means things will get out of sync at some point.

dennybaa:
> I looked over the existing files and I'm a bit confused in some places.

> https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/Makefile#L18
> Any particular reason why we don't copy all the logging configs here? We only do it for st2auth,  st2api, st2actions, but not for st2notifer, st2rulesengine, st2sensorcontainer, etc.

@Kami yeah exactly they are not copied here. makefiles used only for coping files which require renaming, standard `debian/install` handles the operation otherwise. Specifically for this it's here: https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/debian/install#L6-L11

Also it's good to mention that rpm specs inherit debian code using helpers such as https://github.com/StackStorm/st2-packages/blob/master/rpmspec/helpers.spec#L20-L26.

Kami:
@dennybaa That's for clarifying.

> Can we automate this - aka copy the files over in some make step or something like that? Copying it manually for sure means things will get out of sync at some point.

What do you think about that?

dennybaa:
@Kami right now unfortunately nothing( I'll give a try tomorrow to think on it. It's a useful thing but we should concentrate on bugs fishing and fixing as of now.

Kami:
OK, for now I will copy and paste it but lets get something better figured out soon :)

Kami:
Copied the files over - should be all ready now, I think.

Kami:
Build still failing - is there anything I need to do on my side? What's the process to merge it, do I need to merge st2-dockerfiles PR first?

dennybaa:
Noop. Thank you @Kami, you've already added lots of useful stuff. Now I need to add rake build updates. Is it urgent?

Kami:
Nope, not urgent - just wondering.

dennybaa:
@Kami also verified for el6 https://circleci.com/gh/dennybaa/st2-packages/115, working. We can merge if no objections.

Issue #51, opened 2015-12-19
New Packages and st2ctl


armab:
Core Action `st2 packs.restart_component` and `st2 packs` depends on `st2ctl` script, which doesn't exist in new packages system, because of new shiny init/upstrart startup scripts.

Not sure about current/future plans about `st2ctl`.
Or we should still ship `st2ctl`, or edit `st2` core actions:
* [`packs.load`](https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/load.yaml)
* [`packs.packs.restart_component`](https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/restart_component.yaml)

And other occurencies: https://github.com/StackStorm/st2/search?p=1&q=st2ctl&utf8=%E2%9C%93

If edit `st2` actions, - then new release version should be shipped only with new Packages and startup scripts.

#### Reproduce: 
```sh
$ st2 run packs.restart_component servicename=st2auth
.
id: 567583ff73b55b33b955b0a4
status: failed
parameters:
  servicename: st2auth
result:
  failed: true
  return_code: 127
  stderr: 'bash: st2ctl: command not found'
  stdout: ''
  succeeded: false
```

dennybaa:
I see, imo. If packages show themselfs well. I guess this is not viable duplication... Of course much cleaner approach is to let a user to invoke `systemctl restart st2auth` or `restart st2auth`.

armab:
Double checked the [`st2ctl`](https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L125).

Yeah, we should ship `st2ctl` and adopt it with new packages environment (+ backward compatibility).

dennybaa:
@armab You think we can close it already?

armab:
Hey @dennybaa I mentioned you here: https://github.com/StackStorm/st2-packages/pull/64#issuecomment-170687675
Take a look at code.

armab:
Fixed by https://github.com/StackStorm/st2-packages/pull/64

New refactored `st2ctl` is now part of `st2common` package: https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl

PR #58, opened 2015-12-22
Include st2ctl in st2common package

armab:
The first step of https://github.com/StackStorm/st2-packages/issues/51 is to include `st2ctl` as part of `st2common` package.

Checked/Tested manually especially on Centos7 (because we don't run CI tests on it) - the executable is there.
When it's merged - more TBD in `st2` repo.

---

BTW I really liked this part: https://github.com/StackStorm/st2-packages/blob/master/rpmspec/helpers.spec#L28-L34

### DIFF
--- a/packages/st2bundle/debian/install
+++ b/packages/st2bundle/debian/install
@@ -9,3 +9,4 @@
 ../st2exporter/conf/syslog.*.conf etc/st2/
 ../st2reactor/conf/logging.*.conf etc/st2/
 ../st2reactor/conf/syslog.*.conf etc/st2/
+../tools/st2ctl usr/bin/
--- a/packages/st2common/debian/install
+++ b/packages/st2common/debian/install
@@ -4,3 +4,4 @@
 ../contrib/examples usr/share/doc/st2/
 #../contrib/tests usr/share/stackstorm/
 ../docs usr/share/doc/st2/
+../tools/st2ctl usr/bin/
--- a/rake/spec/spec_helper.rb
+++ b/rake/spec/spec_helper.rb
@@ -62,12 +62,12 @@ class ST2Spec
     },
 
     package_has_binaries: {
-      st2common: %w(st2-bootstrap-rmq st2-register-content),
+      st2common: %w(st2-bootstrap-rmq st2ctl st2-register-content),
       st2reactor: %w(st2-rule-tester st2-trigger-refire),
       st2client: %w(st2),
       st2debug: %w(st2-submit-debug-info),
-      st2bundle: %w(st2-bootstrap-rmq st2-register-content st2-rule-tester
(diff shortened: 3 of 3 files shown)


dennybaa:
This is nice. But I'm afraid it should be done like that, packages are completely different in the structure. I think we should not use `tools/st2ctl` there should be a lightweight back-compatibility helper (not even all functionality). As you mentioned in issues there's starting/stoping/restarting problem because actions  rely on st2ctl.

Then IMO should be a new wrapper just for this...

armab:
There is much more (apart of restart services) that rely on existing `st2ctl`.
Small compatibility fix already here: https://github.com/StackStorm/st2/pull/2326

dennybaa:
However it means that we'll extend life of this ugly st2ctl much longer:) Anyway I suggest a slim one so that we are finally able to migrate away from st2ctl... 
Anyway if you add a slim one and problems will arise and arise, you can always switch to default `tools/st2ctl`. But introducing a slim one you do a good thing - trying to finish that old crap!)

But you already merged then it seems you just resurrected a monster:)

dennybaa:
At least we can give a try imo

dennybaa:
Also, maybe if we follow slim not to base on `st2ctl` at all. There's services handling code here https://github.com/StackStorm/st2-packages/blob/master/packages/st2actions/bin/runners.sh also

It's not ideal, but it might useful

armab:
Not merged yet.

`st2ctl` should be there in any way to help control many services related to `st2`, register components and so on. And `st2ctl` already works with init scripts.

Another story, what we can do: is to remove obsolete code, see: https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L198-L317
and ship new `st2ctl` version inside [`st2common/bin`](https://github.com/StackStorm/st2/tree/master/st2common/bin)

What do you think?

dennybaa:
> st2ctl should be there in any way to help control all the services related to st2, register components and > so on. And st2ctl already works with init scripts. It's helper.

1) Yep. I know. But it seemed to be a lot of bash there just for handling services, so this part might be new, just as I suggest above https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L198-L317. But of course not to be exactly the same. Just so far it seemed to work for our packages, only about 20 lines...

2) register components I think can be left if required. However there's `/usr/bin/st2-register-content` already, so again if not required we are not bringing useless wrapping.

3) yeap. in `common/bin` I suppose also.

dennybaa:
I suggest new `st2ctl` with
1) service handling
2) register-componets wrapping if required

If problems keep arise throw it away and fallback to `tools/st2ctl`

armab:
Closed in favor of https://github.com/StackStorm/st2-packages/pull/64

Issue #50, opened 2015-12-19
Package Dependencies: git and virtualenv


armab:
Core st2 Action `st2 run packs.install` assumes that `git` and `virtualenv` binaries are installed globally in the system.

@dennybaa Sounds like we should include them in Package dependencies.
For which packages then, `st2common`, `st2client`?

dennybaa:
St2common is full fledged python virtual env. @armab,  I think it's right to use it for `virtualenv` binary, since st2common is always installed unless it's a bundled version of packages then the path will be `/usr/share/python/st2/bin/virtualenv` (instead `../st2common/bin/virtualenv`)

armab:
Can't find any `virtualenv` binary shipped with new packages:

```sh
$ cd / && find . -name 'virtualenv'
# nothing
```

Install `python-virtualenv`
```sh
$ sudo apt-get install python-virtualenv
...

$ cd / && find . -name 'virtualenv'
./usr/bin/virtualenv
```

armab:
To remember for later fix (hardcoded here): https://github.com/StackStorm/st2/blob/941cb331fa034b5b75eeb6a809239743a9e14ccc/contrib/packs/actions/pack_mgmt/setup_virtualenv.py#L135

dennybaa:
`apt-get install python-virtualenv` is not an option. I think we should add virtualenv into st2client requirements.txt file. Then it'll be bundled into the package. @armab what do you think?

armab:
Sure,
do whatever you think is better for packaging.

armab:
Actually, let me experiment with this.
And I think including it in `st2common`, as you first said is more expected.

There could be no `st2client` installed, but action still can be installed via API call.

dennybaa:
yeah. that's true) what about git, we should think about adding it into package dependency. Huf, need to actually find out where it's `git` and where `git-core`)

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450795033017452

PR #62, opened 2015-12-23
added git into requires line for st2actions package

dennybaa:


dennybaa:
resolves #50

Kami:
:+1: 

(package or at least one of the packages used to be called git-core, is that not the case anymore? or we need git which also includes other stuff?)

dennybaa:
yeah. I've tried to google about that, http://askubuntu.com/questions/5930/what-are-the-differences-between-the-git-and-git-core-packages.

So git-core is dummy package, and at least for all oses which we are aiming to use, `git` is a proper name. Maybe it could be an issue for something really like oses older than 8 years.

PR #68, opened 2016-01-13
Relax save_payload globs for st2web compatibility

enykeev:


enykeev:
Not particularly certain we need this routine for st2web to begin with, but without that change, the script would always return failed payload.

armab:
:+1:

Overall we'll need more adjustments here for `st2web` to work.

PR #69, opened 2016-01-13
Make circle scripts reusable for st2web

enykeev:
Since st2web would be on a different revision cycle, we need a way to pick revision per package.

### DIFF
--- a/.circle/bintray.sh
+++ b/.circle/bintray.sh
@@ -15,8 +15,8 @@ CREATED=201
 # Usage:
 # bintray.sh deploy wheezy_staging /tmp/st2-packages
 # bintray.sh deploy trusty /tmp/st2-packages
-# bintray.sh next-revision trusty 0.12dev
-# bintray.sh next-revision wheezy 1.1.2
+# bintray.sh next-revision trusty 0.12dev st2api
+# bintray.sh next-revision wheezy 1.1.2 st2web
 function main() {
   : ${BINTRAY_ORGANIZATION:=stackstorm}
 
@@ -25,7 +25,7 @@ function main() {
         deploy "$2" "$3"
         ;;
       next-revision)
-        LATEST_REVISION=$(latest_revision "$2" "$3")
+        LATEST_REVISION=$(latest_revision "$2" "$3" "$4")
         if [ -n "${LATEST_REVISION}" ]; then
           echo $((LATEST_REVISION+1))
         else
@@ -34,7 +34,7 @@ function main() {
         ;;
       *)
         echo $"Usage: deploy {wheezy_staging|jessie_staging|trusty_staging} /tmp/st2-packages"
-        echo $"Usage: next-revision {wheezy_staging|jessie_staging|trusty_staging} 0.14dev"
+        echo $"Usage: next-revision {wheezy_staging|jessie_staging|trusty_staging} 0.14dev st2api"
         exit 1
     esac
 }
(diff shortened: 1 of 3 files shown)


dennybaa:
looks reasonable +1, but need @armab to ack

.circle/bintray.sh
```diff
@@ -143,24 +139,30 @@ function check_package_exists() {
...
+  if (! check_package_exists); then
+    debug "The package ${PKG_NAME} does not exist"
+    debug "Creating package ${PKG_NAME}..."
+
+    ${CURL} -X POST -d "${data}" ${API}/packages/${BINTRAY_ORGANIZATION}/${BINTRAY_REPO}/
+    echo ""
+  else
+    debug "Updating package ${PKG_NAME}..."
+
+    ${CURL} -X PATCH -d "${data}" ${API}/packages/${BINTRAY_ORGANIZATION}/${BINTRAY_REPO}/${PKG_NAME}/
```
armab: Can we not update package metadata on every 1 package upload?
Looks like overhead for me, especially in context that for single build in `st2-packages` we upload `10` packages x `5` distros = `50` packages.

Also sometimes package in repo is edited manually via Web UI in Bintray (I did it before several times), this additional request will overwrite it.

enykeev: That's the point, from now on we should avoid editing them manually instead do it only through file. As for making it less frequently, I'll see where we can move it tomorrow. It might well be that we actually can't since every repo contains its own set of packages and changes should be applied to each and every single one of them individually.

armab:
:+1 for revision discovery changes in `bintray.sh`,
but really don't like buildenv changes.

`st2web` logic is really simple, there are several lines to discover `version number` and `revision number`. Can we store this logic (simplified) in `st2web` `circle.yml`?

Thus `st2web` will be more self-sufficient, when you'll have more control over `st2web` vars in that repo, instead of relying on `st2-packages` repo.
At the same time less overcomplication of `st2-packages` bash, which is already becomes more unreadable.

Is it an option for you?

enykeev:
Since you can't share env variables between circleci calls, it will imply storing a copy of script inside of st2web and st2flow. We have no shell scripts there besides that and I would rather prefer to keep that status quo. This all seem like a complicated enough logic to try to fit it in oneliner inside of circle.yml itself.

armab:
No, it's really simple for `circle.yml` in `st2web`:

```yaml
...
checkout:
  post:
    - |
      PKG_VERSION=$(node -e "console.log(require('./package.json').st2_version);")
      PKG_RELEASE=$(.circle/bintray.sh next-revision trusty_staging ${PKG_VERSION} st2web)
      echo "export PKG_VERSION=${PKG_VERSION}" >> ~/.circlerc
      echo "export PKG_RELEASE=${PKG_RELEASE}" >> ~/.circlerc
...
```

enykeev:
K, I'll give it a shot tomorrow.

.circle/bintray.sh
armab: Why should we avoid editing packages in Bintray manually via Web ui where it's really needed?
Also every package could have its own custom description/special notes/metadata.

The intent behind this "create package if not exists" with really generic meta is just to avoid errors when we still didn't created package in repo.

Just remember package creation is something that needed only once and not frequently used. Most likely once package in repo created/description is adjusted - we won't return to it later.

enykeev: > Why should we avoid editing packages in Bintray.com manually via Web ui where it's really needed?

For one, because we have a packet per distro and no way of automate the change. That's A LOT of manual changing. Last time I've checked, a single change in current `.bintray_package` inside st2-packages will require you to change it for every single one of 60+ packages out there.

But mostly, it's just a benefits of `* as a code` - known state, history, audit and all that jazz.

enykeev: I've added that initially to be able to define config for st2web since it's quite different from st2. Then I figured it won't cost us anything to update them on every deploy. Am I missing some hidden cost?

enykeev:
Ok, I've inlined buildenv changes to st2web circle.yml and excluded the commits related to it from the branch. As for package updates, I don't see how we can reduce the number of requests, though I don't see any problem in the current situation. Binray documentation claims they don't limit requests for org owned repos and an overhead of 40 (not 60, I've actually went and counted them all) requests would hardly be an issue. We'll review it again if I'll be proven wrong, we have a number of options there.

armab:
Won't be blocker here.
It's just unexpected behavior to overwrite it every time over and over, including messing any manual web ui changes (which should be always an option).

I suspect we won't ever need to change any default metadata in `.bintray_package` once it was setup for the first time. We'll see.

armab:
:+1:

PR #75, opened 2016-01-14
Fix the binary paths for garbage collector

lakshmi-kannan:
See https://stackstorm.atlassian.net/browse/STORM-1966 for details.


dennybaa:
thank you! definitely caused by copy paste) merging.

Kami:
Good catch.

To clarify - everything lives under st2/ and nothing under stereactor/?

dennybaa:
yes for bundle it's exactly it.

PR #80, opened 2016-01-18
bundle build only

dennybaa:


armab:
Is `package_name` ENV variable I can pass from Circle to build only these 2 packages?

armab:
Please also change `st2api` here to `st2bundle`: 
https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv.sh#L47

And if `st2bundle` and `mistral` just hardcoded - we're good to try the build.

armab:
Let's try :+1:

armab:
https://stackstorm.slack.com/archives/standup/p1453133448000706

lakshmi-kannan:
:+1:

PR #76, opened 2016-01-15
Rhel fixes

dennybaa:

PR #83, opened 2016-01-18
WIP: Initial check-in for install instructions [ci skip]

lakshmi-kannan:
Let's get these into version control first and iterate on them

### DIFF
--- a/packages/st2bundle/install_rhel6.md
+++ b/packages/st2bundle/install_rhel6.md
@@ -1,48 +1,41 @@
-## Install IUS repo
-sudo yum install -y https://dl.iuscommunity.org/pub/ius/stable/Redhat/6/x86_64/ius-release-1.0-14.ius.el6.noarch.rpm
+# Add EPEL repo
+* sudo yum install -y epel-release
+
+# Install tools
+* sudo yum install -y httpd-tools
 
 # Install mongodb
-sudo yum install -y mongodb mongodb-server
+* sudo yum install -y mongodb mongodb-server
 
 # Install rabbitmq
-sudo rpm --import https://www.rabbitmq.com/rabbitmq-signing-key-public.asc
-curl -sS -k -o /tmp/rabbitmq-server.rpm https://www.rabbitmq.com/releases/rabbitmq-server/v3.3.5/rabbitmq-server-3.3.5-1.noarch.rpm
-sudo yum localinstall -y /tmp/rabbitmq-server.rpm
+* sudo yum install -y rabbitmq-server
 
 ## Install st2
-sudo touch /etc/yum.repos.d/StackStorm-el6_staging-unstable.repo
-
-Add following lines to /etc/yum.repos.d/StackStorm-el6_staging-unstable.repo
-[StackStorm-el6_staging-unstable]
-name=StackStorm-el6_staging-unstable
+* cat << EHD | sudo tee /etc/yum.repos.d/bintray-stackstorm-el6_staging_unstable.repo
+[bintraybintray-stackstorm-el6_staging]
+name=bintray-stackstorm-el6_staging
 baseurl=https://dl.bintray.com/stackstorm/el6_staging/unstable
-enabled=1
 gpgcheck=0
+enabled=1
+EHD
 
-sudo yum install st2bundle
-
+* sudo yum install st2bundle
 
 ## Start mongodb
-sudo chkconfig mongod on
-sudo service mongod start
+* sudo /etc/init.d/mongod start
 
 ## Start rabbitmq
-sudo chkconfig rabbitmq-server on
-sudo service rabbitmq-server start
+* sudo /etc/init.d/rabbitmq-server start
 
 ## Start st2 services
-sudo st2ctl start
+* sudo st2ctl start
 
 ## Register content
-st2-register-content --config-file=/etc/st2/st2.conf --register-all
+* st2-register-content --config-file=/etc/st2/st2.conf --register-all
 
 ## Setup auth user
-sudo yum install httpd-tools
-sudo htpasswd -cs /etc/st2/htpasswd admin
-
-## Install git (Fix st2bundle to require git.)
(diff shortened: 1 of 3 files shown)


manasdk:
If everything needs sudo would it be simpler to `sudo su` as the first step? Likely will make it easier to copy paste.

lakshmi-kannan:
>If everything needs sudo would it be simpler to sudo su as the first step? Likely will make it easier to copy paste.

st2-register-content and st2 CLI commands don't require sudo. Explicitly calling out the steps that really need sudo seems better to me. Anyways, this can be changed later. I was trying to validate https://github.com/StackStorm/st2-packages/pull/76 and without RHEL instructions, I wasted 30 mins to get to a bare st2 install. Also found couple of issues. So intent is to get things into version control first and iterate on packages and instructions.

packages/st2bundle/install_ubuntu.md
```diff
@@ -0,0 +1,36 @@
+## Install repo tools
+sudo apt-get update
+sudo apt-get install -y software-properties-common
+
+## Install virtualenv
+sudo apt-get install -y python-pip
+sudo -H pip install --upgrade pip
+sudo -H pip install virtualenv
+virtualenv --version (> 13.1.2)
```
dennybaa: this copied from manas, and is not required. souldn't be here...

packages/st2bundle/install_rhel7.md
```diff
@@ -0,0 +1,37 @@
...
+sudo yum localinstall -y /tmp/rabbitmq-server.rpm
+
+## Start mongodb
+sudo systemctl start mongod
+
+## Start rabbitmq
+sudo systemctl status rabbitmq-server
+
+## Install st2 (Fix this to use repo.)
+wget https://bintray.com/artifact/download/stackstorm/el7_staging/unstable/st2bundle-1.3dev-67.x86_64.rpm
```
armab: `sudo touch /etc/yum.repos.d/StackStorm-el7_staging-unstable.repo`
```ini
[StackStorm-el7_staging-unstable]
name=StackStorm-el7_staging-unstable
baseurl=https://dl.bintray.com/stackstorm/el7_staging/unstable
enabled=1
gpgcheck=0
```

```sh
yum install st2bundle
```

packages/st2bundle/install_ubuntu.md
lakshmi-kannan: yep. removed.

packages/st2bundle/install_rhel6.md
```diff
@@ -0,0 +1,48 @@
+## Install IUS repo
+sudo yum install -y https://dl.iuscommunity.org/pub/ius/stable/Redhat/6/x86_64/ius-release-1.0-14.ius.el6.noarch.rpm
+
```
dennybaa: Should be then same for el6/el7 `wget -qO - "https://setup.ius.io/" | sudo bash -s`.
To make it more generic.

packages/st2bundle/install_rhel6.md
```diff
@@ -0,0 +1,48 @@
...
+# Install mongodb
+sudo yum install -y mongodb mongodb-server
+
+# Install rabbitmq
+sudo rpm --import https://www.rabbitmq.com/rabbitmq-signing-key-public.asc
+curl -sS -k -o /tmp/rabbitmq-server.rpm https://www.rabbitmq.com/releases/rabbitmq-server/v3.3.5/rabbitmq-server-3.3.5-1.noarch.rpm
```
dennybaa: I suggest to preinstall `wget` and use it instead of curl, usually wget is always available. Since there's no curl installed previously - this is not safe.

packages/st2bundle/install_rhel6.md
```diff
@@ -0,0 +1,48 @@
...
+sudo yum localinstall -y /tmp/rabbitmq-server.rpm
```
dennybaa: Why required specific version of rabbit? It would be cool if in the docs on the step like this the was a remark: *"Grabbing from repo, since ius and os repo provide outdated one (required >= 3.3)"*

packages/st2bundle/install_rhel7.md
```diff
@@ -0,0 +1,45 @@
+# Add EPEL repo
+sudo yum install -y epel-release
```
dennybaa: then we should use ius as above then! https://ius.io/GettingStarted/#scripted-install

packages/st2bundle/install_rhel6.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Start st2 services
+sudo st2ctl start
+
+## Register content
+st2-register-content --config-file=/etc/st2/st2.conf --register-all
+
+## Setup auth user
+sudo yum install httpd-tools
+sudo htpasswd -cs /etc/st2/htpasswd admin
```
dennybaa: Don't use `-s` it is marked as **insecure**, use default (i.e. no crypt option given) https://httpd.apache.org/docs/2.4/programs/htpasswd.html#options

packages/st2bundle/install_rhel6.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Install git (Fix st2bundle to require git.)
+sudo yum install git-all
```
dennybaa: This step is not required anymore, should be deleted. I've fixed it https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/rpm/st2bundle.spec#L7-L11

packages/st2bundle/install_rhel6.md
lakshmi-kannan: It seems like there is a difference in htpasswd shipped in RHEL6 vs Ubuntu. In Ubuntu, better to use -B (very secure) I think.

lakshmi-kannan:
Closing in favor of https://github.com/StackStorm/st2-packages/pull/109

PR #84, opened 2016-01-19
Missing rhel fixes bundle

dennybaa:
1. Added missing copy-paste pieces for rhel rpm specs into st2bundle.
2. Wiped out st2exporter service from packages

cc: @lakshmi-kannan , @manasdk, @emedvedev, @Kami


dennybaa:
feel free to come back to reopen, since i'm merging, because these are important minor fixes we need to have.

lakshmi-kannan:
lgtm

Issue #90, opened 2016-01-20
Remove specs for debian/ rpm/ except bundle and mistral


dennybaa:
cc: @lakshmi-kannan @kirilstrax @armab 
I suggest to think about **premature optimization**.

Honestly, we are not able to ship separate packages in the release. Now we are doing excessive work fixing files in st2* as well st2bundle packages, **not** all of that is only copy paste work, because it requires manual validation.
My personal feeling that what we are doing is  **premature optimization**, I also have a feeling that after this release we won't look at separate packages.

**Pros**
* We get component separation

**Cons**
* It's not real life scenario, since even you do some job on fixing package A. It won't work, since you need to propagate code from package A to st2bundle. Unless you've done work is in vain. (also #66)
* It's not fun validating init files for each service and st2bundle after copy-paste.
* It's simply duplication of efforts.

PS. component separation should be done in st2bundle package (logical enable/disable).
Please welcome, to discuss! But I'm asking you to be honest in terms of double effort and premature optimization.

dennybaa:
I removed 5k of lines and it still works https://circleci.com/gh/StackStorm/st2-packages/243, heh. Why we should ship that fatty!?

dennybaa:
as agreed with @dzimine it's a right move. issue is resolved in #163

armab:
> I removed 5k of lines and it still works

Sounds like Windows

armab:
Closing, since https://github.com/StackStorm/st2-packages/pull/163 solves it

PR #91, opened 2016-01-20
removing apparently "FAR FUTURE PREOPTIMIZATION"

dennybaa:
concerns #90


enykeev:
I'm personally against removing work that already have been done. Worse than that, we'll be breaking `git blame` so when we're going to bring this files back again, there will be no way to figure out who added specific line and why. And in this particular project it's especially important, there's just too much of "WAT?" moments. I'd say that if we are concerned of build time and amount of effort required to move this project forward, lets only focus on st2bundle\mistral and sync the rest later, but let the stuff stay.

dzimine:
Let's go for it.

dennybaa:
awesome @dzimine I totally agree, I will prepare probably new PR a bit later. And of course all the work we'll save in a feature branch.

dzimine:
Please do sooner: once we share this with the world today, people will look and the less confusion the better.

dennybaa:
@dzimine closed in favor #163

PR #92, opened 2016-01-20
fixed umask to 022, for debian init and upstart services

dennybaa:
fixed default umask for debain/ubuntu systems.

PS: systemd, redhat sysv init scripts will use 022 by default if not explicitly set.


lakshmi-kannan:
http://stackoverflow.com/questions/13268796/umask-difference-between-0022-and-022 Looking at that, it seems this would work fine. Since actionrunner runs as root, with umask022, it does look like user st2 will still be able to read the files. So sensor container should run. Can you please merge and ship packages so I can test this?

jfryman:
lgtm

dennybaa:
@lakshmi-kannan merging. There's no difference between 0022 and 022 it's the same octal number :)

Issue #94, opened 2016-01-20
WSGI for st2api and st2auth


armab:
Current PR in progress: https://github.com/StackStorm/st2-packages/pull/100

#### TODO:
* [x] `st2auth`/`st2api` socket decision - use `http`
  - [x] ~~unix socket creation permission problems~~
  - [x] ~~unix socket ownership and how to deal with external nginx/apache/lighttpd~~
* [x] Make it minimally working (automated tests)
  - [x] deb
  - [x] rpm
* [x] Replace `uwsgi` with `gunicorn` for `st2auth`. So both `st2auth` and `st2api` will use the same consistent solution.
* [x] Add basic port tests for `st2api` and `st2auth`: #103
* [x] [bug] Identify and fix `st2api` not listening issue: https://github.com/StackStorm/st2-packages/issues/106
* [x] Manual Tests
  - [x] [Stress and Performance tests with `ab`/`siege`-like tools](https://stackstorm.slack.com/archives/packages/p1453999866001064)
  - [x] Ubuntu Trusty
  - [ ] Debian Wheezy
  - [ ] Debian Jessie
  - [ ] CentOS7
  - [ ] CentOS6

----

Old/obsolete commits with discussions:
* https://github.com/StackStorm/st2-packages/commit/ed74295b36bc0bf0205c554c3e496f07df3f9f25
* https://github.com/StackStorm/st2-packages/commit/d8c29d82922fee1ebd9cc2fec95a3588d5cbccdf

dennybaa:
@armab Thank you for opening tickets for @enykeev . It's makes life a lot clearer :+1:

enykeev:
updated version at d8c29d82922fee1ebd9cc2fec95a3588d5cbccdf

armab:
As noted in https://stackstorm.slack.com/archives/stackstorm/p1453440097029836:

There are several approaches to handle these sockets & permissions: 
* 1.) Run `st2api` and `st2auth` startup scripts with default `setuid`/`setgid` (from root) to do the thing like ensuring directory exists/create `/var/run/st2` with proper permissions in pre-start hook. And launching the main process like `setuidgid st2 command` from `st2` user.
* 2.) Running `st2api` service to ensure `/var/run/st2` (from `root`) with child service `start on starting st2api` (from `st2` user).
[...]

PR #100, opened 2016-01-22
WSGI for st2auth and st2bundle

armab:
Let's open PR, since almost all blockers resolved with [no need of unix sockets anymore](https://github.com/StackStorm/st2-packages/issues/94#issuecomment-173985267) and we can commit, rather then experiment/try & force-push.

Addresses issue #94

> Most of the work is done by @enykeev


enykeev:
@armab I'd say we're ready to merge. If it looks fine to you, merge it and I'm going to test it across all the platforms from bintray tomorrow.

armab:
Let me debug a little.

If timeout increase worked, then we should better hardcode it in `st2-packages`, without touching `circle.yml`.

Don't want to include `circle.yml` changes in `st2` repo.

armab:
Regarding: https://circleci.com/gh/StackStorm/st2-packages/262

I could run tests with `ST2_WAITFORSTART=1` for 10 times in a row in CircleCI + SSH environment, so couldn't reproduce it.

```sh
docker-compose -f docker-compose.circle.yml run -e ST2_GITURL=https://github.com/StackStorm/st2 -e ST2_GITREV=master -e ST2PKG_VERSION=1.4dev -e ST2PKG_RELEASE=8 -e RABBITMQHOST=172.17.0.1 -e POSTGRESHOST=172.17.0.1 -e MONGODBHOST=172.17.0.1 -e ST2_WAITFORSTART=1 wheezy test
```

Let's merge, but if it appears once again - what we can do in future to debug better - share `/var/log/st2` logs and service logs via Circle artifacts, to read these files once something wrong appears (could be a good idea in any way).

armab:
Good to know, that even with `0` timeout it worked well: 
https://circleci.com/gh/StackStorm/st2-packages/268

See this build: https://circleci.com/gh/StackStorm/st2-packages/268
where I accidentally set timeout to `0`, described here: https://github.com/StackStorm/st2-packages/pull/105

packages/st2api/debian/st2api.init
```diff
@@ -18,8 +18,8 @@
 PATH=/sbin:/usr/sbin:/bin:/usr/bin
 DESC="st2api"
 NAME=st2api
-DAEMON=/usr/share/python/st2api/bin/$NAME
-DAEMON_ARGS="--config-file /etc/st2/st2.conf"
+DAEMON=/usr/share/python/st2api/bin/gunicorn_pecan
+DAEMON_ARGS="/usr/share/python/st2api/lib/python2.7/site-packages/st2api/gunicorn_config.py -k eventlet -b 127.0.0.1:9101 --threads 10 --workers 1"
```
dennybaa: @enykeev , @armab this is inappropriate, there's a config file which is meant to be customized.  You make it static by hardcoding the path to the default same way it was before.
cc: @lakshmi-kannan 

Good way is:
1. make, say it, `/etc/st2/config`.
2. place there file there

So the enduser can get control of it. 
What do you think?

enykeev: gunicorn does not allow you to pass arguments the same way you're passing them to the script. You can although redefine config pach using env variable https://github.com/StackStorm/st2/blob/master/st2api/st2api/gunicorn_config.py#L35-L36. I don't see it as an immediate concern, the amount of users who need to change the config is extremely low, but you believe it is a problem, feel free to pass the variable.

armab: +1 @enykeev

In this special case, like it more, comparing to config file.
And path is already hardcoded in many other startup scripts, don't see any inconsistencies here.

dennybaa: yep. that's right I agree. thanks for clarifications.

PR #105, opened 2016-01-25
Fix ST2_WAITFORSTART when it's empty

armab:
Example: https://circleci.com/gh/StackStorm/st2-packages/268

When `ST2_WAITFORSTART` is empty, eg bash:
```sh
export ST2_WAITFORSTART=
```
which is our case when `ST2_WAITFORSTART` could be optional (https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv.sh#L53), this is transformed in Ruby part as `nil` -> `0`.

That results in:
pnx7rq5 1

We want to use default numeric value when `ST2_WAITFORSTART` is empty.

### DIFF
--- a/rake/spec/spec_helper.rb
+++ b/rake/spec/spec_helper.rb
@@ -36,7 +36,7 @@ class ST2Spec
     rabbitmqhost: pipeopts.rabbitmqhost,
     postgreshost: pipeopts.postgreshost,
     mongodbhost:  pipeopts.mongodbhost,
-    wait_for_start: (ENV['ST2_WAITFORSTART'] || 10).to_i,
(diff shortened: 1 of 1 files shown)


armab:
Actually, wrong code change in Ruby, because bash part https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv.sh#L33 should handle it.

armab:
Got it, it comes from: https://github.com/StackStorm/st2-packages/blob/43725de61320c9879bbf9a12fe35951cc5fa7a30/.circle/docker-compose.sh#L52
which sets `export ST2_WAITFORSTART=$ST2_WAITFORSTART` to empty.

So change is Ruby part is legit.

PR #108, opened 2016-01-26
fix to wait_for_start

dennybaa:


dennybaa:
@armab this is the fix for fix. The thing is the bellow code is incorrect:
```
(10 if ENV['ST2_WAITFORSTART'].empty?)
```


It will always fail if env variable is not provided, since `nil` object doesn't have `empty?` instance method.

PR #109, opened 2016-01-26
installation docs for ubuntu/debian [ci skip]

dennybaa:
@lakshmi-kannan I think these instructions are more clear and organised have a look.


debian_install.md
```diff
@@ -0,0 +1,100 @@
...
+
+Second, add repositories and install st2.
+```shell
+## NB! This step provides instructions for ubuntu trusty
+# Don't forget to use correct distro name!
+export DISTRO=trusty
+
+wget -qO - https://bintray.com/user/downloadSubjectPublicKey?username=bintray | sudo apt-key add -
+
+sudo -E add-apt-repository "deb https://dl.bintray.com/stackstorm/${DISTRO}_staging stable main"
```
armab: For `wheezy` at least from my experiments `add-apt-repository` won't work, also `http` should be used to make it working.
So it's better just `>> /etc/apt/sources.list`

debian_install.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Install dependent services
+
+St2 uses MongoDB database and RabbitMQ messaging queue, so we need to install this services.
+```
+sudo apt-get install -y mongodb-server rabbitmq-server
+
+# Allow localhost connections to rabbitmq
+echo '[{rabbit, [{disk_free_limit, 10}, {loopback_users, []}]}].' | sudo tee /etc/rabbitmq/rabbitmq.config
```
armab: Are you sure it's needed for minimal install?

It works by default without this thing (tested before). Looks like part of advanced config tuning and so on stuff.

debian_install.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Install mistral (optional)
+
+If you are planning to use mistral workflows follow the provided instructions otherwise you can ignore them.
+
+Install postgres database and mistral:
+```shell
+# st2mistral package provides st2 plugin for mistral
+sudo apt-get install -y postgresql mistral st2mistral
```
armab: Can we make `st2mistral` dependency for `mistral`? So it will be installed by default.

debian_install.md
dennybaa: yes. it makes guest:guest@localhost available, otherwise no

debian_install.md
dennybaa: yup might be so. actually didn't check on wheezy

debian_install.md
armab: Discussed already, nice trick by @lakshmi-kannan 
Let's use `st2mistral` only in any installation docs.
`apt-get install st2bundle st2mistral`

debian_install.md
```diff
@@ -0,0 +1,100 @@
...
+sudo service postgresql start
+```
+
+Populate mistral configuration file:
```
armab: Can be omitted since:
https://github.com/StackStorm/st2-packages/pull/110

PR #110, opened 2016-01-26
Default connection settings in mistral.conf

armab:
Currently user should generate mistral config by hand, adjusting `rabbitmq` and `postgresql` configuration.

And we do the same via our smoke tests: 
u17xv8d 1

Let's place mistral config with default connection, assuming everything is local, removing one step for 80% of users to make it minimally working.

Since mistral config is managed by packaging, once user change default config - it won't be replaced automatically during package upgrade, what is perfectly nice.

> Please commit changes directly to `mistral-default-config` shared branch if you have something to add/adjust for default [`mistral.conf`](https://github.com/StackStorm/st2-packages/blob/master/packages/mistral/conf/mistral.conf).

### DIFF
--- a/packages/mistral/conf/mistral.conf
+++ b/packages/mistral/conf/mistral.conf
@@ -171,6 +171,7 @@
 # configuration. If not set, we fall back to the rpc_backend option
 # and driver specific configuration. (string value)
 #transport_url = <None>
+transport_url = rabbit://guest:guest@localhost:5672
 
 # The messaging driver to use, defaults to rabbit. Other drivers
 # include qpid and zmq. (string value)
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
lgtm.

dennybaa:
+1. awesome @armab

armab:
Guys, keep in mind: I have no idea what I'm doing, since it's mistral.

For example, how bad is this value for default installation?
```sh
[pecan]
# Enables user authentication in pecan. (boolean value)
auth_enable = false
```
Like is this OK for production defaults?

packages/mistral/conf/mistral.conf
```diff
@@ -840,3 +843,4 @@
 
 # Enables user authentication in pecan. (boolean value)
 #auth_enable = true
+auth_enable = false
```
armab: Guys, keep in mind: I have no idea what I'm doing, since it's mistral.
For example, how bad is this value for default installation?

armab: Ok, found it's used for OpenStack only:
https://pypi.python.org/pypi/mistral#configuring-mistral

Merging.

PR #114, opened 2016-01-27
Mistral gunicorn

dennybaa:


dennybaa:
added systemd services at the point

enykeev:
Feel like we need to start generating them instead of writing them manually. Too easy to make a typo, too hard to catch it afterwards.

dennybaa:
well, they are similar (if you mean init files), but not identical. Try to generate and everything will break.
Another option is to make universal and template from it, but it also sucks not so easy :)

PR #107, opened 2016-01-25
[Tests] Check if st2 services are listening on respective ports

armab:
Add real check if `st2api`, `st2auth`, `mistral` are reachable on `9100`, `9101`, `8989`

Guide:
http://serverspec.org/resource_types.html

-----
First we should include `netstat` binary, required for these checks, which is not available under minimal Docker.

@dennybaa can you add `net-tools` for all test Docker images & publish them, see: https://github.com/StackStorm/st2-dockerfiles/pull/7 ?

PR #116, opened 2016-01-27
Remove `prompt_on_failure` for new Rspec tests, change format

armab:
As requested by @dennybaa https://github.com/StackStorm/st2-packages/pull/107/files#r50952487


armab:
@dennybaa: no service logs:  in case of failure: https://circleci.com/gh/armab/st2-packages/439

after removing `prompt_on_failure: true`.

I need those logs.

Issue #119, opened 2016-01-27
Mistral API is not running while service status reports OK


armab:
`jessie` failure: https://circleci.com/gh/StackStorm/st2/496

`mistral` service is considered as running by Rspec tests, but in fact it's not fully running, `mistral-api` is not listening on respective port.

```sh
root@7732908eafdc:/# service mistral status
[FAIL] mistral-api is not running ... failed!
[ ok ] mistral-server is running.
root@7732908eafdc:/# echo $?
0
```

`/var/log/mistral-api.log`:
```log
root@7732908eafdc:/var/log/mistral# cat mistral-api.log
[2016-01-27 19:18:29 +0000] [593] [INFO] Starting gunicorn 19.4.5
[2016-01-27 19:18:29 +0000] [593] [INFO] Listening at: http://127.0.0.1:8989 (593)
[2016-01-27 19:18:29 +0000] [593] [INFO] Using worker: sync
[2016-01-27 19:18:29 +0000] [638] [INFO] Booting worker with pid: 638
[2016-01-27 19:18:29 +0000] [639] [INFO] Booting worker with pid: 639
[2016-01-27 19:18:32 +0000] [639] [ERROR] Exception in worker process:
Traceback (most recent call last):
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/arbiter.py", line 515, in spawn_worker
    worker.init_process()
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 122, in init_process
    self.load_wsgi()
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/workers/base.py", line 130, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 65, in load
    return self.load_wsgiapp()
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 52, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/usr/share/python/mistral/local/lib/python2.7/site-packages/gunicorn/util.py", line 357, in import_app
    __import__(module)
[...]

Issue #113, opened 2016-01-26
Mistral sometimes doesn't listen on `8989` after moving to `gunicorn`


armab:
![0prdztz 1](https://cloud.githubusercontent.com/assets/1533818/12594166/5d84d3da-c47d-11e5-8fca-bb361792557b.png)

I retried builds several times and got always different results. See the builds: 
* https://circleci.com/gh/StackStorm/st2-packages/283
  - `mistral` failed on `wheezy` and `jessie`
* https://circleci.com/gh/StackStorm/st2-packages/284
  - `mistral` + `st2auth` failed on `wheezy`
  - `mistral` failed on `jessie`
* https://circleci.com/gh/StackStorm/st2-packages/285
  - `mistral` failed on `wheezy` and `jessie`

It looks like very similar to ghost bug we have after adding `gunicorn` for both `st2api` and `st2auth`: https://github.com/StackStorm/st2-packages/issues/106

----
Actually, same thing. Here `mistral` just fails almost every time.

Good to debug tomorrow in Circle + SSH, to understand what's happening inside, since we have it more reproducible. For now stats are the following: 
* `st2auth` or `st2api` fails on `jessie` only
* `mistral` fails on both `wheezy` and `jessie`
* `trusty`, `el6`, `el7` are always OK.

dennybaa:
gunicorn starts slowly, you should increase timeout, imo.

dennybaa:
actually wrong conf substitution `/etc/default/mistral` for debian. el6 el7 are always OK, since that PR doesn't bring init files for rhel)

dennybaa:
https://github.com/StackStorm/st2-packages/commit/6c5683902c03cbbfddc093ecffc088e7fce98671

armab:
Sadly, it's still there: https://circleci.com/gh/StackStorm/st2/491
Latest `st2` & `st2packages` codebase, `ST2_WAITFORSTART` is set to `15 sec`, `mistral` failed under `jessie`.

Same as `st2auth` and `st2api` after moving to `gunicorn`. Still investigating.

armab:
Related (but could be the same issue actually): https://github.com/StackStorm/st2-packages/issues/119

armab:
@dennybaa confirmed https://github.com/StackStorm/st2-packages/issues/122 fixed the issue, closing.

PR #122, opened 2016-01-28
Correct order for mistral commands

armab:
Try to fix mistral

https://stackstorm.slack.com/archives/packages/p1453939452000836
https://stackstorm.slack.com/archives/packages/p1453939787000852

Related to https://github.com/StackStorm/st2-packages/issues/113 and https://github.com/StackStorm/st2-packages/issues/119

### DIFF
--- a/install_debian.md
+++ b/install_debian.md
@@ -57,11 +57,11 @@ CREATE DATABASE mistral OWNER mistral;
 EHD
 ```
 
-Start mistral and update schema:
+Update schema and start mistral:
 ```
-sudo service mistral start
 /usr/share/python/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
 /usr/share/python/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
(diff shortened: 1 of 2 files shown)


lakshmi-kannan:
Holy jolly!

dennybaa:
@lakshmi-kannan @armab 
yup. the order was correct when used clean db. So it might be so that we don't have a clean db anymore.

I wonder if circle ci nodes (1,2,3..) use the same shared service containers such as postgres, rabbit and mongo?

dennybaa:
if db is not clean and it is so. So also it's a mystery why it worked before with mistral started with one worker spawned with `mistral-server` (handled by eventlet.wsgi)... Hmm it's strange...

dennybaa:
i've merged seems, order fixed the issue completely.

armab:
> yup. the order was correct when used clean db.

No, as @m4dcoder the order should be *exactly*:
1) run all migrations
2) start mistral

armab:
We couldn't merge this PR only because GitHub went down :/

PR #124, opened 2016-01-28
mistral workers = 2 and start up order test

dennybaa:


rake/spec/default/60-st2_all-services-ok_spec.rb
```diff
@@ -28,10 +25,9 @@
 
 describe 'start st2 components and services' do
   before(:all) do
-    # Populate mistral db tables with initial data.
-    # We don't test this invocation, due to docker-compose up actually DOES NOT
-    # start clean postgres (so command fails on the second up invocation %-).
-    #
+    # run upgrade head
+    # run populate
+    # only after start mistral
```
armab: Only **before** start mistral

rake/spec/default/60-st2_all-services-ok_spec.rb
```diff
@@ -7,20 +7,17 @@
 
   describe 'rabbitmq' do
     subject { host(spec[:rabbitmqhost]) }
-    it { is_expected.to be_reachable }
```
armab: I think these tests are legit.

I intentionally leaved them here, because we can have 2 issues:
1.) When host doesn't exists at all
2.) When host reachable, but mongo or any other service is not reachable by port

They're completely different and show different problems.
Leaving both tests will help to identify issues faster, for example when we change Docker version and will have changed IPs (as it was when I switched to Docker 1.9, if you remember).

armab:
This should fix: https://github.com/StackStorm/st2-packages/issues/113

armab:
Probably this PR should fix: https://github.com/StackStorm/st2-packages/issues/113 

Are you sure with `workers > 1` there is clearly some race condition/bug? Because after changing order of migration commands we thought all problems (at least for mistral) were gone.

rake/spec/default/60-st2_all-services-ok_spec.rb
dennybaa: yeah)

dennybaa:
i'm merging. yes fixing order, helped mistral specs never failed.

rake/spec/default/60-st2_all-services-ok_spec.rb
armab: Man, you merged as is.

Issue #106, opened 2016-01-25
"ERROR: HTTPConnectionPool(host='127.0.0.1', port=9101)" with gunicorn


armab:
![it7dra4 1](https://cloud.githubusercontent.com/assets/1533818/12561025/8ffb5d70-c3a6-11e5-9d6e-c6d01e3606d2.jpg)

For some still unknown reason there is a ghost bug, appearing rarely when `st2api` with recent [`gunicorn` change](https://github.com/StackStorm/st2-packages/pull/100) is not reachable by port (service is starting OK).

* First occurence: https://circleci.com/gh/StackStorm/st2-packages/262
* Second occurence (even after lifting `ST2_WAITFORSTART` to 10sec as possible workaround): https://circleci.com/gh/StackStorm/st2/477

~~Actually, we're not even sure if `st2auth` is reachable at all, because there are [no tests](https://github.com/StackStorm/st2-packages/issues/103).~~
We have the tests now.

armab:
Stats:

* `st2auth` failed on `wheezy`: https://circleci.com/gh/StackStorm/st2/485
* `st2api` failed on `wheezy`: https://circleci.com/gh/StackStorm/st2-packages/287
* `st2auth` failed on `wheezy`: https://circleci.com/gh/StackStorm/st2-packages/289
* `st2api` failed on `el6`: https://circleci.com/gh/StackStorm/st2-packages/291 (even after lifting `ST2_WAITFORSTART` timeouts to `15sec`)

armab:
On `el6` we have some logs:
```py
last 20 lines from log file of service st2api
>>>
2016-01-27 15:35:19,160 140432054114048 ERROR log [-] Conflict while trying to save in DB.
Traceback (most recent call last):
  File "/usr/share/python/st2/lib/python2.7/site-packages/st2common/persistence/base.py", line 155, in add_or_update
    model_object = cls._get_impl().add_or_update(model_object)
  File "/usr/share/python/st2/lib/python2.7/site-packages/st2common/models/db/__init__.py", line 174, in add_or_update
    instance.save()
  File "/usr/share/python/st2/lib/python2.7/site-packages/mongoengine/document.py", line 286, in save
    raise NotUniqueError(message % unicode(err))
NotUniqueError: Tried to save duplicate unique keys (E11000 duplicate key error index: st2.trigger_d_b.$pack_1_name_1  dup key: { : "core", : "st2.sensor.process_exit" })
[...]

PR #126, opened 2016-01-28
fixing status_of_proc invocation debain sysv

dennybaa:


armab:
:+1: Fixes part of issue https://github.com/StackStorm/st2-packages/issues/106#issuecomment-175828932 when died services were reported `OK` status under `wheezy` and `jessie`.

PR #127, opened 2016-01-29
Revert "fixing status_of_proc invocation debain sysv"

enykeev:
Reverts StackStorm/st2-packages#126

Just checking if it would affect CircleCI "wait for ssh" issue.


enykeev:
Ah nvm, miscommunication.

dennybaa:
@enykeev you can play with such stuff in your circle ci repository, there's no need to make a PR. Generally or that you playing with it's better to do in your circleci account, this helps not block other builds.

Issue #112, opened 2016-01-26
st2 services logs are in DEBUG level. Set them to INFO by default. Same for mistral.


lakshmi-kannan:

lakshmi-kannan:
So I wanted to make this change and it looks like the change needs to happen in actual st2 repo. https://github.com/StackStorm/st2/blob/master/st2reactor/conf/logging.sensorcontainer.conf#L11

Can we copy all those logging.conf files to conf/ directory in packages? I am split on this personally. Having two places of truth is bad. I'd like to hear y'all's opinions.

armab:
Closed via https://github.com/StackStorm/st2-packages/pull/131

PR #131, opened 2016-01-30
St2bundle to st2

dennybaa:
- [x] renaming package st2bundle to st2.
- [x] moving st2 and mistral virtual environments under `/opt/stackstorm`.
- [x] providing examples in `/opt/stackstorm/packs/examples`.
- [x] setting default log level for components to **INFO**. (#112)


packages/st2/rpm/st2.spec
```diff
@@ -65,15 +65,15 @@ Conflicts: st2common
...
-  %{_datadir}/python/%{venv_name}
-  %{_datadir}/doc/st2/examples
-  %attr(755, %{svc_user}, %{svc_user}) %{_localstatedir}/log/st2
+  %config(noreplace) %attr(640, %{svc_user}, %{svc_user}) %{_sysconfdir}/st2/htpasswd
+  /opt/stackstorm/%{venv_name}
   /opt/stackstorm/packs/core
   /opt/stackstorm/packs/linux
   /opt/stackstorm/packs/packs
-  %attr(640, %{svc_user}, %{svc_user}) %{_sysconfdir}/st2/htpasswd
+  /opt/stackstorm/packs/examples
```
armab: Shouldn't be directory for packs examples `/opt/stackstorm/st2/packs/examples`?
Since packs is part of `st2`.

packages/st2/debian/install
```diff
@@ -1,7 +1,7 @@
 ../contrib/core opt/stackstorm/packs/
 ../contrib/packs opt/stackstorm/packs/
 ../contrib/linux opt/stackstorm/packs/
-../contrib/examples usr/share/doc/st2/
+../contrib/examples opt/stackstorm/packs/
```
armab: Same point here with renaming, `opt/stackstorm/st2/packs/examples/` IMO.

armab:
Since we rename `st2bundle` to `st2` and have new dirs `/opt/stackstorm/mistral` and `/opt/stackstorm/st2`, I really assumed that we place all `mistral` and `st2` bits into respective directories.

So I've installed generated packs from this build and directory treee looks this way:

```
root@ubuntu14:/vagrant# tree -L 2 /opt/stackstorm
/opt/stackstorm
├── exports
├── mistral
│   ├── bin
│   ├── etc
│   ├── include
│   ├── lib
│   ├── local
│   ├── pip-selfcheck.json
│   └── share
├── packs
│   ├── core
│   ├── examples
│   ├── linux
│   └── packs
└── st2
    ├── bin
    ├── include
    ├── lib
    ├── local
    ├── pip-selfcheck.json
    └── share

19 directories, 2 files
```


According to: http://www.pathname.com/fhs/pub/fhs-2.3.html#OPTADDONAPPLICATIONSOFTWAREPACKAGES 
> The structure of the directories below /opt/<provider> is left up to the packager of the software, though it is recommended that packages are installed in /opt/`provider`/`package` and follow a similar structure to the guidelines for /opt/package. A valid reason for diverging from this structure is for support packages which may have files installed in /opt/provider/lib or /opt/provider/bin.

suggests `/opt/vendor/package` directory structure.
As I understand `/opt/stackstorm/packs` is left there for backward-compatibility reasons. I believe `/packs` directory is really part of `st2`.

We had a small discussion in Slack regarding this ^^.

dzimine:
Looks ok by me; keep in mind there may be chatops or st2-chatops there, too however you decide to name it.

Kami:
Looks good to me as well.

Where do we plan to put virtual envs, `/opt/stackstorm/virtualenvs`?

dennybaa:
@Kami , I'd like discuss this, your suggestions. For me would be nice smth like this `/opt/stackstorm/run/cache/venvs`. I don't know really, but `/opt/stackstorm/virtualenvs` definitely not the best choice.
Or maybe:
```
/opt/stackstorm/var/cache/venvs
/opt/stackstorm/var/lib/packs
``` 
Where as if we migrate later to default paths, we can use:
```
/var/cache/st2/venvs
/var/lib/st2/packs
```
That's all my vision at the moment...

lakshmi-kannan:
Though not ideal, let's leave the dir structure as such. In a future release, we can target moving packs and virtualenvs to appropriate locations and when we do that, we should post upgrade notes and blog posts with that explicit change.

lakshmi-kannan:
LGTM.

Kami:
Agreed with lakshmi.

Flat structure also makes it a bit easier to move around and inspect thing.

dennybaa:
@armab as it seems everyone choses opt structure as he finds rational for himself...
So our choice is not so bad, at least it backward compatible. Though I agree we might want  to change it in the future, nothing blocks us from that...
BTW. chef server layout also doesn't repeat `/opt/<vendor>/<package>`  scheme completely...

PR #132, opened 2016-02-01
installation docs update centos/rhel added [ci skip]

dennybaa:


install_centos.md
```diff
@@ -0,0 +1,123 @@
...
+First let's update the system and fetch software which will be required during installation and setup process:
+```shell
+sudo yum makecache fast
+sudo yum install -y wget httpd-tools
+```
+
+Second, add repositories and install st2.
+```shell
+## NB! This step provides instructions for centos7 distro
+# Don't forget to use correct distro (el6 - for rhel6/centos6, and el7 - for rhel7/centos7)!
```
armab: Don't rely on Bintray file.

@enykeev has instructions about placing correct repository configuration file, see:
https://github.com/enykeev/st2testground/blob/master/bootstrap-el6.sh#L19-L32
https://github.com/enykeev/st2testground/blob/master/bootstrap-el7.sh#L19-L32

install_centos.md
```diff
@@ -0,0 +1,123 @@
...
+
+If you are planning to use mistral workflows follow the provided instructions otherwise you can ignore them.
+
+First you need to choose what PostgreSQL version you want to use. Since we are installing st2 on Centos 7 (in the current example). We will use the following [link](http://yum.postgresql.org/9.5/redhat/rhel-7-x86_64/pgdg-centos95-9.5-2.noarch.rpm). (For Centos 6,  you may use this [link](http://yum.postgresql.org/9.5/redhat/rhel-6-x86_64/pgdg-centos95-9.5-2.noarch.rpm)).
+*Make sure you have chosen the correct distro and arch, follow this [__instructions__](http://yum.postgresql.org/) to locate the desired repo*. As you find the required URL substitute it into the command bellow:
+
+```shell
+# NB! Repo creation link for Cento 7.
+sudo yum install -y http://yum.postgresql.org/9.5/redhat/rhel-7-x86_64/pgdg-centos95-9.5-2.noarch.rpm
+```
```
armab: Just default `postgresql`: `yum install -y postgresql-server` for default installation. See: https://github.com/enykeev/st2testground/blob/master/bootstrap-el7.sh#L16

Everything else like version is up to user.

install_centos.md
```diff
@@ -0,0 +1,123 @@
...
+sed -i "s|/${DISTRO}_staging|/${DISTRO}_staging/stable|" /etc/yum.repos.d/bintray-stackstorm-${DISTRO}_staging.repo
+unset DISTRO
+
+# Update repo and install st2
+sudo yum install -y st2
+```
+
+## Install dependent services
+
+St2 uses MongoDB database and RabbitMQ messaging queue, so we need to install these services. We suggest to use [IUS repo](https://ius.io/), it contains MongoDB and RabbitMQ. It's worth mentioning that IUS repo is *EPEL*-compatible, because EPEL comes as its part. So let's install the repo and proceed with the installation:
```
armab: Why custom repo?

This is good one for mongo:
https://github.com/enykeev/st2testground/blob/master/bootstrap-el6.sh#L9-L15
https://github.com/enykeev/st2testground/blob/master/bootstrap-el7.sh#L9-L15
and default for rabbitmq.

install_centos.md
```diff
@@ -0,0 +1,123 @@
...
+
+```
+# Install IUS repo
+wget -O - https://setup.ius.io/ | sudo bash -s
+
+sudo yum install -y mongodb-server rabbitmq-server
```
manasdk: What version of mongodb-server does this install?

install_centos.md
dennybaa: I don't to bother with docs and installation, neither anyone I guess. IUS(EPEL) seems good at it was used in our production! So I prefer to use it.

install_centos.md
dennybaa: https://dl.fedoraproject.org/pub/epel/7/x86_64/repoview/mongodb-server.html == 2.6.11-1
https://dl.fedoraproject.org/pub/epel/6/x86_64/repoview/mongodb-server.html == 2.4.14-1

dennybaa: My point is that it's a decent idea to use standard and well tested OS repositories such as EPEL. As long as it's possible due to version. Also don't forget it's intro documentation and should stay clean and small as possible.
If later on we want to provide best practices and tuning for production use cases (such as postgresql tuning and configuration), that definitely should be a separate doc. Well, let's not put all the eggs into one  basket now :)

install_centos.md
dennybaa: I'm totally not sure that you can go with this on rhel, though for centos it apparently works.

install_centos.md
dennybaa: This should be a separate small doc on how to add staging unstable/stable...
This docs for stable for an and user! Of course they don't work at the moment since bintray is empty :)

dennybaa:
anyone minds from merging this?

Issue #89, opened 2016-01-20
Wrong st2client is used in st2bundle


armab:
Seems `st2bundle` somehow uses https://pypi.python.org/pypi/st2client for installation.

#### st2bundle way:
```sh
wget -qO - https://bintray.com/user/downloadSubjectPublicKey?username=bintray | apt-key add -
add-apt-repository 'deb https://dl.bintray.com/stackstorm/trusty_staging unstable main'
apt-get update
apt-get install st2bundle
```

```sh
# st2 --version
st2 ('1.2.0.8',)
```

#### Standalone way:
```sh
wget https://circle-artifacts.com/gh/StackStorm/st2-packages/227/artifacts/0/home/ubuntu/packages/trusty/st2common_1.3dev-97_amd64.deb
wget https://circle-artifacts.com/gh/StackStorm/st2-packages/227/artifacts/0/home/ubuntu/packages/trusty/st2client_1.3dev-97_amd64.deb

dpkg -i *.deb
```

```sh
# st2 --version
st2 1.3dev
```

----

I double checked the source code for both and seems client in `st2bundle` is really outdated, because it doesn't include recent commit affecting `st2client` source code: https://github.com/StackStorm/st2/commit/6928e7bbf0c3174bc0baa854f87c2250a2fb02ce

My best guess that the possible cause could be related to: https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/in-requirements.txt#L5 ?

armab:
Yeah, for sure.
`st2client` updated to `1.3.0.9` and now it's here as well: 

![xp2orij 3](https://cloud.githubusercontent.com/assets/1533818/12554377/e1799e68-c384-11e5-8f60-6a9dd89dafcb.png)

armab:
@dennybaa Any details where it was fixed?

armab:
:+1:

PR #133, opened 2016-02-02
Pam backend added into bundle

dennybaa:


dennybaa:
i believe it also resolves #89

dennybaa:
`/opt/stackstorm/st2/bin/pip list`

dennybaa:
@lakshmi-kannan we can merge this in, have a look please.

PR #138, opened 2016-02-02
a dep for st2mistral

dennybaa:


packages/st2mistral/debian/control
```diff
@@ -11,6 +11,6 @@ Homepage: http://stackstorm.com/
 Package: st2mistral
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.16), python2.7, mistral | python-mistral, ${misc:Pre-Depends}
-Depends: ${shlibs:Depends}, ${misc:Depends}
+Depends: ${shlibs:Depends}, ${misc:Depends}, mistral | python-mistral
```
dzimine: why python-mistral is required and where do we get it from? we don't build python-mistral, it is likely upstream from https://github.com/openstack/python-mistralclient cc @m4dcoder

packages/st2mistral/debian/control
```diff
@@ -11,6 +11,6 @@ Homepage: http://stackstorm.com/
...
 Description: StackStorm plugins for OpenStack Mistral
  <insert long description, indented with spaces>
```
dzimine: Make sure this is fixed when you ship.
"StackStorm's Mistral workflow service."

packages/st2mistral/debian/control
dennybaa: don't ask so many questions) we are ditching plugin package

PR #77, opened 2016-01-15
Let Bintray repos to keep only latest 10 revisions for each pkg version

armab:
Introduce `MAX_REVISIONS` - number of latest revisions to keep for package version.

Ex: with `MAX_REVISIONS=10`, after deploying `st2api_1.2dev-20_amd64.deb`, obsolete package revision `st2api_1.2dev-10_amd64.deb` will be deleted during the same run.

Do this for all packages which are deployed.

PR #140, opened 2016-02-03
Add packagecloud support

enykeev:


armab:
:+1: thanks for keeping backward-compatible logic with old Bintray script.

From what I see we deploy only to `st2_staging` repo. `stable`/`unstable` is still would be good for diversification.

armab:
Also we'll need a way to delete obsolete packages like https://github.com/StackStorm/st2-packages/pull/77

But that's later, for sure.

PR #141, opened 2016-02-03
St2mistral full

dennybaa:
This our full mistral bundle package, which contains st2mistral pip module. @dzimine @lakshmi-kannan


armab:
@dennybaa do I correctly understand, it fetches contesnts from both
https://github.com/stackstorm/mistral
https://github.com/stackstorm/st2mistral
repos and creates `st2mistral` package?

packages/st2/Makefile
```diff
@@ -53,7 +53,7 @@ ifeq ($(DEBIAN),1)
 endif
 
 wheelhouse: .stamp-wheelhouse
-.stamp-wheelhouse: populate_version requirements
+.stamp-wheelhouse: | populate_version requirements
```
armab: Some trick here with `|`?
Curious.

packages/st2mistral/debian/control
```diff
@@ -4,13 +4,13 @@ Priority: optional
 Maintainer: StackStorm Engineering <opsadmin@stackstorm.com>
 Build-Depends: debhelper (>= 9), python, dh-virtualenv (>= 0.8)
 Standards-Version: 3.9.5
-Homepage: http://stackstorm.com/
-#Vcs-Git: git://git@github.com:StackStorm/st2mistral.git
-#Vcs-Browser: https://github.com/StackStorm/st2mistral.git
+Homepage: <insert the upstream URL, if relevant>
```
armab: Why wtackstorm.com was purged as a homepage?

packages/st2mistral/debian/control
```diff
@@ -4,13 +4,13 @@ Priority: optional
...
+#Vcs-Git: git://anonscm.debian.org/collab-maint/mistral.git
```
armab: I guess this is correct git url: https://github.com/StackStorm/mistral.git, no?

lakshmi-kannan: Good catch, @armab

packages/st2mistral/Makefile
```diff
@@ -1,8 +1,14 @@
...
+COMPONENT := st2mistral
 MISTRAL_RELEASE ?= 1
-# hardcode it for now since st2mistral doesn't have it's version
-MISTRAL_VERSION ?= 1.2.0
+MISTRAL_VERSION ?= $(shell python -c "from pbr import version; print version.VersionInfo('mistral').version_string(),")
+
+# We use special additional requirements in our mistral bundle!
+define INJECT_DEPS
+gunicorn
+-e git+https://github.com/StackStorm/st2mistral.git@$(GITREV)#egg=st2mistral
```
lakshmi-kannan: It appears we should also include python-mistralclient which gets the mistral CLI as well. python-mistralclient also follows the same branching mechanism as st2mistral. See https://stackstorm.slack.com/archives/packages/p1454525032003033 for why.

packages/st2/Makefile
dennybaa: this a order only dependency, here's some info https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

packages/st2mistral/debian/control
dennybaa: that's copy paste bro

packages/st2mistral/debian/control
dennybaa: all these are insignificant issues. to be honest all descriptions, licensing etc should be fixed.

armab: I don't argue with that, for sure insignificant issues. Not even issues.

But we deploy that package with that description once this PR is merged.

PR #149, opened 2016-02-04
install nginx example to doc location

dennybaa:


armab:
:+1:

armab:
Also, after merging this, here is they way how you can trigger the build and produce new `1.3` packages manually via CircleCI:
- Navigate to `v1.3` branch builds https://circleci.com/gh/StackStorm/st2/tree/v1.3
- Find latest build
- Click `Rebuild`

Since under `st2` repo we always download latest https://github.com/StackStorm/st2-packages - new deployed packages will be based on this updated codebase.

packages/st2/debian/install
```diff
@@ -6,3 +6,4 @@
 ../st2actions/conf/syslog.*.conf etc/st2/
 ../st2reactor/conf/logging.*.conf etc/st2/
 ../st2reactor/conf/syslog.*.conf etc/st2/
+../conf/nginx usr/share/doc/st2/conf/
```
dzimine: suggest `/st2/conf/nginx`: it's likely we'll get more config examples (e.g. one day we decide support deploying on apache, or integrate with other 3rd party where a config sample needed).
With that, +1.

dennybaa: thie target directory is `/usr/share/doc/st2/conf/nginx`. we copy directory nginx to directory `doc/st2/conf/`...

PR #152, opened 2016-02-05
Choose 127.0.0.1 or localhost for consistency

armab:
@lakshmi-kannan The problem you're experiencing with mistral client couldn't connect to `localhost` is deeper than just adding `OS_MISTRAL_URL=127.0.0.1` env variable, when default `localhost` is used by mistral client.

Currently we mix both `127.0.0.1` and `localhost` in entire `st2` codebase: 
* [`27` code occurencies for `127.0.0.1`](https://github.com/StackStorm/st2/search?p=1&q=127.0.0.1&type=Code&utf8=%E2%9C%93)
* [`65` code occurencies for `localhost`](https://github.com/StackStorm/st2/search?p=1&q=localhost&type=Code&utf8=%E2%9C%93)

With this mish-mash, when you have in `/etc/hosts`:
```ini
127.0.1.1 localhost
```
as you showed before https://gist.github.com/lakshmi-kannan/3ac4d39ed91721da7cfb, - it will fail in any way somewhere in `st2` where `localhost` is used.

So or we change everything everywhere to `localhost` or everything everywhere to `127.0.0.1`.
In both [`mistral`](https://github.com/openstack/mistral/search?utf8=%E2%9C%93&q=localhost) and `mistralclient` [`localhost`](https://github.com/openstack/python-mistralclient/search?utf8=%E2%9C%93&q=localhost) is used consistently.

Originally suggested by @dennybaa is to use `localhost` as default, see discussion starting from:
https://stackstorm.slack.com/archives/packages/p1454677177003670

-----

So this PR should fix your issue with packages & mistral client. I'll test it further.

**But** we also should change everything (what possible) in `st2` repo for consistency, at least:
https://github.com/StackStorm/st2/blob/master/st2api/st2api/gunicorn_config.py#L29
https://github.com/StackStorm/st2/blob/master/st2auth/st2auth/gunicorn_config.py#L29

because in config used by packages, everything is already `localhost`:
https://github.com/StackStorm/st2/blob/master/conf/st2.package.conf#L70


lakshmi-kannan:
@DoriftoShoes What do you think? I am skeptical.

DoriftoShoes:
This PR is good but doesn't solve the client issue.  Unless I am missing where we set the environment variable for the mistral endpoint.

armab:
Discussion: https://stackstorm.slack.com/archives/packages/p1454691471003800
Majority to use `127.0.0.1` over `localhost`.

:+1: on choosing something consistent and follow that, but that still means upstream changes in `st2`.

dennybaa:
I also plus +1 to `127.0.0.1`, since at the begging I didn't think carefully enough. What if localhost will be `::1` only and will be resolved as IPv6 address, so it was a bad idea initially. Sorry for that.

armab:
Closed in favor of https://github.com/StackStorm/st2-packages/issues/157

PR #153, opened 2016-02-05
Don't forget chatops pack

dzimine:
For now, without st2chatops package adding chatops by user is really ugly. 
I am ok to revert this when we do st2chatops, ideally, the chatops pack would be in that package.


armab:
:+1:

PR #159, opened 2016-02-08
Update and rename st2_bootstrap_deb.sh to st2bootstrap-deb.sh

dzimine:
rename by the name convention
remove redundant line from ST2 OK


armab:
You can use `[ci skip]` to avoid building/deploying new packages on doc change.
It will work for both commit message and Pull Request title.

https://circleci.com/docs/skip-a-build

Issue #157, opened 2016-02-08
Use 127.0.0.1 instead of localhost for consistency


armab:
> Original story here https://github.com/StackStorm/st2-packages/pull/152 and https://gist.github.com/lakshmi-kannan/3ac4d39ed91721da7cfb

Mixing `localhost` vs `127.0.0.1` in configs leads to bugs/connection problems with some special configurations in `/etc/hosts`, like:
```ini
127.0.1.1 localhost
```

* [`27` code occurencies for `127.0.0.1`](https://github.com/StackStorm/st2/search?p=1&q=127.0.0.1&type=Code&utf8=%E2%9C%93)
* [`65` code occurencies for `localhost`](https://github.com/StackStorm/st2/search?p=1&q=localhost&type=Code&utf8=%E2%9C%93)

Apart of chaging `localhost` -> `127.0.0.1` and use one solution from now, we also should change what possible in `st2` repo for consistency to avoid confusion and mish-mash in future.

PR #160, opened 2016-02-08
Mistral client wrapper

armab:
Overrides `OS_MISTRAL_URL` to use `127.0.0.1` instead of `localhost` for mistral client via wrapper script, approach descried: https://www.debian.org/doc/debian-policy/ch-opersys.html#s9.9

Related to issue #157

cc @lakshmi-kannan

### DIFF
--- a/packages/st2mistral/bin/mistral
+++ b/packages/st2mistral/bin/mistral
@@ -1,0 +1,8 @@
+#!/bin/sh
+# Wrapper script, used to override default mistral client environment variables
+# https://github.com/StackStorm/python-mistralclient#running-mistral-client
+
+OS_MISTRAL_URL=${OS_MISTRAL_URL:-http://127.0.0.1:8989/v2}
+export OS_MISTRAL_URL
+
+exec /opt/stackstorm/mistral/bin/mistral "$@"
--- a/packages/st2mistral/debian/install
+++ b/packages/st2mistral/debian/install
@@ -1,1 +1,2 @@
 conf/mistral.conf /etc/mistral
+bin/mistral /usr/bin
--- a/packages/st2mistral/debian/st2mistral.links
+++ b/packages/st2mistral/debian/st2mistral.links
@@ -1,1 +1,0 @@
-opt/stackstorm/mistral/bin/mistral usr/bin/mistral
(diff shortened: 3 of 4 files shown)


armab:
Reproducing your issue: 

```sh
echo '127.0.1.1 localhost' > /etc/hosts
```

This works OK now, since `127.0.0.1` is default for mistral client: 
```sh
root@ubuntu14:/home/vagrant# mistral execution-list
+--------+----------+-------------+-------------------+--------+------------+------------+------------+
| ID     | Workflow | Description | Task Execution ID | State  | State info | Created at | Updated at |
+--------+----------+-------------+-------------------+--------+------------+------------+------------+
| <none> | <none>   | <none>      | <none>            | <none> | <none>     | <none>     | <none>     |
+--------+----------+-------------+-------------------+--------+------------+------------+------------+
```

Overriding `127.0.0.1` -> `localhost` example: 
```sh
export OS_MISTRAL_URL=http://localhost:8989/v2

root@ubuntu14:/home/vagrant# mistral execution-list
ERROR (app) HTTPConnectionPool(host='localhost', port=8989): Max retries exceeded with url: /v2/executions?sort_keys=created_at&sort_dirs=asc (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x7f7214dd29d0>: Failed to establish a new connection: [Errno 111] Connection refused',))
```

dennybaa:
So does it work? I think no... Wrapper script should the follow logic, imo...

1. if `OS_MISTRAL_URL` is not given, then should be set as `http://127.0.0.1:8989/v2`
2. if `OS_MISTRAL_URL` given, it should be used.
3. if `OS_MISTRAL_URL` contains localhost, then localhost should be overwritten to `127.0.0.1` (since our service listens there).

This logic handles wrapper scenario in more correct way, fix me if I'm wrong.

armab:
It implements `1` and `2` from your list.

`3` I think is unexpected and agressive behavior. If you pass `OS_MISTRAL_URL` env variable explicitly (no matter what value) - it will be used. No string modification/replace is needed here, - you have full control over env variable.

dennybaa:
yup. maybe it's an overkill indeed.

lakshmi-kannan:
Me likey likey.

armab:
That thing is in stable packages (after rebuild)

PR #162, opened 2016-02-10
Runners move (fixup)

dennybaa:


dennybaa:
ps. copy in makefile resolves permissions issue which had place when copying using `debian/install`

PR #165, opened 2016-02-12
Fix gunicorn graceful timeout

armab:
There are cases with `st2api`, `st2auth`, `mistral-api` failed to stop in "everything went wrong" cases.
"Everything went wrong" case for them was when DB is down.

Found it in `wheezy` for `.init` scripts.

#### Reproduce
1.) `st2ctl stop`
2.) `service mongodb stop && service postgresql stop`
3.) `st2ctl start`
4.) `st2ctl stop`
5.) `st2ctl status`

#### Problem 1
`st2api`, `st2auth`, `mistral` are still running after `service stop`:
```
st2api PID: 5350
st2auth PID: 5604
mistral PID: 11612
mistral PID: 11613
mistral-api PID: 11612
mistral-api PID: 11613
```

#### Problem 2
`st2ctl stop` command will last for `90` seconds waiting for `gunicorn` with `mistral`, `st2api`, `st2auth` to finish (30 seconds each).

The fix is to rely on http://docs.gunicorn.org/en/stable/settings.html#graceful-timeout setting. It should be less then `start-stop-daemon` https://github.com/StackStorm/st2-packages/blob/d35721d9c6b70c0108a72bba4885ca4a476004ae/packages/st2/debian/st2.st2api.init#L81 which probably doesn't handle things well for `gunicorn`.

----

So `TERM`/`KILL` and graceful timeout decision is handled by `gunicorn` now.


dennybaa:
well it's not a bug, though)

armab:
`service st2auth stop`
`service st2api stop`
`service mistral stop `
and running services after that: 
```
st2api PID: 5350
st2auth PID: 5604
mistral PID: 11612
mistral PID: 11613
mistral-api PID: 11612
mistral-api PID: 11613
```

is bug.

dennybaa:
I can't understand anything, from your explanation( But that services shutdown slow it is not a bug, i.e. honestly that is not a fix!
If services shutdown slow the reason should be found and fixed, however reducing wait to kill time fixes only symptoms, but not the issue.
In other words normally operating service which has received **SIGTERM** should shutdown fast if not, and you just reduced wait before **KILL**, service will be shutdown brutally. Of course it fixes the symptom since services stop faster, but it is not solving the originating reason in any way.
Maybe of course I  misunderstanding smth...

armab:
Services shut down slow - is not the root bug.

The bug is when you `service st2api stop` and it's not stopped at all. I mean at all, even after some time (1 minute, 2 minutes), meaning `start-stop-daemon` doesn't really kill/clean all processes for some reason.

Here I switched the responsibility to gracefully shut down its master and nodes to `gunicorn`, which is smart enough to handle that better and probably has custom mechanisms.

dennybaa:
This I do understand) Then this delegation should help well enough :+1:.

Kami:
This graceful-timeout doesn't seem to work for me. It still seems to wait 30+ seconds or so for st2actionrunner and st2api.

armab:
@Kami In what distro you got this timeout issue again? Since you know that startup scripts are different and may behave in different ways.

I would try to play with added `timeout=30` parameter [in your PR](https://github.com/StackStorm/st2-packages/pull/192),  like setting it to 5 secs (to isolate it).
http://docs.gunicorn.org/en/stable/settings.html#timeout
To isolate it and make sure it doesn't affects anything.

Also it might be related to different worker type `eventlet` -> `sync`.

^^ Things to play with.

Kami:
Yeah, sorry for not including more details.

I have this issues on Ubuntu 14.04 (upstart). I did try changing (removing and lowering) a timeout argument and it didn't seem to make a difference. And yeah, I do think it could be related to the sync worker :)

PR #169, opened 2016-02-15
Enabling auth [ci skip]

DoriftoShoes:
This PR requires the following one on the st2 repo:

https://github.com/StackStorm/st2/pull/2501

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -61,6 +61,7 @@ use_st2ctl() {
 }
 
 configure_authentication() {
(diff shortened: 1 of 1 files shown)


armab:
:+1: 

Also added `[ci skip]` to avoid the build: https://circleci.com/docs/skip-a-build

lakshmi-kannan:
lgtm.

dennybaa:
Aren't you thinking to make it default, if so better here https://github.com/StackStorm/st2/blob/master/conf/st2.package.conf#L39

PR #175, opened 2016-02-18
Script to help migrate from bintray to pkgcloud [ci skip]

lakshmi-kannan:
Tested on a VM with old packages installed.


dzimine:
cool.

PR #176, opened 2016-02-18
Fix bootstrap script to use pkgcloud [ci skip]

lakshmi-kannan:
TODO:

- [x] Test the script -- Tested on a brand new ubuntu box. Seems to work fine!


armab:
https://circleci.com/docs/skip-a-build

> Currently [ci skip] does not support fork PRs.

I think you can add `[ci skip]` into your commit message in that case to control for what commits you need build and for what you don't need build.

PR #177, opened 2016-02-18
Update Packagecloud to follow `st2` versioning strategy for private repos

armab:
Since private packages now following `st2` versioning strategy, it means we need +2 `unstable` private repos from now (we had only `stable` before), meaning 2 different tokens generation.
FYI @lakshmi-kannan


dennybaa:
:+1:

PR #181, opened 2016-02-25
Add bootstrap script for el7 [ci skip]

armab:
Bootstrap script, which codifies manual installation steps for `el7`: https://gist.github.com/armab/2ab4d2ccc592db422439

Tested under: 
* RHEL7 AWS
* CentOS7 AWS

Had a chance to dig into Selinux policies.
Included some rules which I faced during tests to avoid disabling Selinux completely and rebooting machine  (RHEL7 case).
So far it works on "default installations", but for custom distros with custom settings it may fail in any place because of more enforcing policies enabled.
The plan B is: disable Selinux.

Thanks @lakshmi-kannan for example with [Selinux Mongo](https://docs.mongodb.org/getting-started/shell/tutorial/install-mongodb-on-red-hat/#configure-selinux).

### DIFF
--- a/scripts/st2bootstrap-el7.sh
+++ b/scripts/st2bootstrap-el7.sh
@@ -86,7 +86,7 @@ verify_st2() {
   st2 execution list
   
   # Fire a remote comand via SSH (Requires passwordless SSH)
-  st2 run core.remote hosts='localhost' -- uname -a
+  st2 run core.remote hosts='127.0.0.1' -- uname -a
   
   # Install a pack
   st2 run packs.install packs=st2
--- a/scripts/st2bootstrap-el7.sh
+++ b/scripts/st2bootstrap-el7.sh
@@ -146,6 +146,31 @@ install_st2web() {
   sudo systemctl restart nginx
 }
 
+ok_message() {
+    echo ""
+    echo ""
+    echo "███████╗████████╗██████╗      ██████╗ ██╗  ██╗";
+    echo "██╔════╝╚══██╔══╝╚════██╗    ██╔═══██╗██║ ██╔╝";
+    echo "███████╗   ██║    █████╔╝    ██║   ██║█████╔╝ ";
+    echo "╚════██║   ██║   ██╔═══╝     ██║   ██║██╔═██╗ ";
+    echo "███████║   ██║   ███████╗    ╚██████╔╝██║  ██╗";
+    echo "╚══════╝   ╚═╝   ╚══════╝     ╚═════╝ ╚═╝  ╚═╝";
+    echo ""
+    echo "  st2 is installed and ready to use."
+    echo ""
+    echo "Head to https://YOUR_HOST_IP/ to access the WebUI"
+    echo ""
+    echo "Don't forget to dive into our documentation! Here are some resources"
+    echo "for you:"
+    echo ""
+    echo "* Documentation  - https://docs.stackstorm.com"
+    echo "* Knowledge Base - https://stackstorm.reamaze.com"
+    echo ""
+    echo "Thanks for installing StackStorm! Come visit us in our Slack Channel"
+    echo "and tell us how it's going. We'd love to hear from you!"
+    echo "http://stackstorm.com/community-signup"
+}
+
 
 adjust_selinux_policies
 
@@ -160,4 +185,4 @@ install_st2mistral
 
 install_st2web
 
-echo "OK"
+ok_message
--- a/scripts/st2bootstrap-el7.sh
+++ b/scripts/st2bootstrap-el7.sh
@@ -1,0 +1,183 @@
+#!/bin/bash
+
+set -eux
+
+# Note that default SELINUX policies for RHEL7 differ with CentOS7. CentOS7 is more permissive by default
+# Note that depending on distro assembly/settings you may need more rules to change
+# Apply these changes OR disable selinux in /etc/selinux/config (manually)
+adjust_selinux_policies() {
+  if getenforce | grep -q 'Enforcing'; then
+    # Allow rabbitmq to use '25672' port, otherwise it will fail to start
+    semanage port --list | grep -q 25672 || semanage port -a -t amqp_port_t -p tcp 25672
+
+    # Allow network access for nginx
+    setsebool -P httpd_can_network_connect 1
+  fi
+}
+
+install_st2_dependencies() {
+  sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  sudo yum -y install mongodb-server rabbitmq-server
+  sudo systemctl start mongod rabbitmq-server
+}
+
+install_st2() {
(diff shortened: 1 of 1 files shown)


scripts/st2bootstrap-el7.sh
```diff
@@ -0,0 +1,183 @@
...
+configure_st2_user
+configure_st2_authentication
+verify_st2
+
+install_st2mistral_depdendencies
+install_st2mistral
+verify_st2mistral
+
+install_st2web
+verify_st2web
```
lakshmi-kannan: I am not sure why you went with a different approach than ubuntu? https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L119

It would be better to be consistent with the steps in ubuntu. Otherwise, we'd have maintenance nightmares. 

cc: @DoriftoShoes

lakshmi-kannan: I see you do st2 verification here. https://github.com/StackStorm/st2-packages/pull/181/files#diff-8e27538ece7dcd42e91428ec3390d2f9R71.  Can you make sure this script is consistent with ubuntu though? Basically, we have to have same verification steps and same OK message printed at the end.

scripts/st2bootstrap-el7.sh
```diff
@@ -0,0 +1,183 @@
...
+  fi
+  
+  sudo mkdir -p /home/stanley/.ssh
+  sudo chmod 0700 /home/stanley/.ssh
+  
+  # On StackStorm host, generate ssh keys
+  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  
+  # Authorize key-base acces
+  sudo cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys
```
lakshmi-kannan: I ran ./st2bootstrap-el7.sh (note without sudo)
+ sudo cat /home/stanley/.ssh/stanley_rsa.pub
./st2bootstrap-el7.sh: line 44: /home/stanley/.ssh/authorized_keys: Permission denied

Fails. 

http://askubuntu.com/questions/351403/permission-denied-while-making-a-file-without-asking-for-password
This works:
sudo cat /home/stanley/.ssh/stanley_rsa.pub | sudo tee -a /home/stanley/.ssh/stanley_rsa.pub

scripts/st2bootstrap-el7.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
+
+set -eux
```
lakshmi-kannan: Please don't set -x. In fact don't set anything unless needed. If installation fails at a step and I re-run the script, I get 

```
[vagrant@st2test ~]$ ./st2bootstrap-el7.sh
+ adjust_selinux_policies
+ grep -q Enforcing
+ getenforce
+ install_st2_dependencies
+ sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
Loaded plugins: fastestmirror
epel-release-latest-7.noarch.rpm                                                                                                                                                                                                      |  14 kB  00:00:00
Examining /var/tmp/yum-root-NM4sll/epel-release-latest-7.noarch.rpm: epel-release-7-5.noarch
/var/tmp/yum-root-NM4sll/epel-release-latest-7.noarch.rpm: does not update installed package.
Error: Nothing to do
[vagrant@st2test ~]
```

lakshmi-kannan: If you need to explicitly exit on some failure, catch exit codes and quit. BTW ubuntu script doesn't do this. Not a bash expert here and I am ears for opinions but status quo is bad in terms of UX.

lakshmi-kannan:
Other than the two issues, there was a couple of other issues as well. 
I test this stuff on VMs from here
https://github.com/StackStorm/st2-test-ground/blob/master/Vagrantfile

1) SSH: I was testing on centos 7 and I always hit this issue! Happens 100% of the time for me.
https://gist.github.com/lakshmi-kannan/b4fa4c7134a5272d7795

2) UI doesn't come up: If I point my browser to https://{{VM_IP}/, I always get connection refused.
This never happens for the ubuntu image in that Vagrantfile. 

So would you findout what's different between this image and the one you use? I want to fix this Vagrantfile so I can point it to users as well. We do this today.

scripts/st2bootstrap-el7.sh
armab: This is expected to exit on error, from my point of view.

I tried Ubuntu script and it doesn't catch any errors at all.
Even if any command fails - script marked as succeeded. Not sure if that's expected.

Kami: We should definitely fix this.

This was one of the more annoying parts of existing bootstrap script - even if some step or command failed we still tried to continue which caused a lot of confusion and unnecessary support burden.

We should fail loud and early.

scripts/st2bootstrap-el7.sh
armab: Changed to `Ch@ngeMe`

scripts/st2bootstrap-el7.sh
armab: Important note, fixed.

armab:
@lakshmi-kannan I'm testing on `AWS` for both `RHEL7` and `CentOS7`, as noted here: https://gist.github.com/armab/2ab4d2ccc592db422439

Your issues are user-configuration/infra related. I checked both: 

1) Tried to dig really quick into this, seems something with box/sftp. Needs more deep debugging with understanding what `st2 run core.remote hosts='127.0.0.1' -- uname -a` really does.

2) You need to disable firewall
```sh
sudo systemctl stop firewalld
sudo systemctl disable firewalld
```

^^ These issues are custom distro related which come with different configurations. For example here is Vagrant box which doesn't have these issues (line for `Vagrantfile`):
```
config.vm.box = 'hfm4/centos7'
```

scripts/st2bootstrap-el7.sh
armab: I removed verification for `st2mistral` and `st2web`.
You're right, what we really should do here - is codify changes from the installation docs.

armab:
Synced script with installation doc.

What important to remember here:
The script just codifies installation instructions from the docs.
It's not Idempotent script (same as I see applies for deb script) and not going to be, otherwise we'll end up again with `st2_deploy.sh` approach with lots of code and many if/then/else instructions.

scripts/st2bootstrap-el7.sh
```diff
@@ -0,0 +1,163 @@
...
+configure_st2_user
+configure_st2_authentication
+verify_st2
+
+install_st2mistral_depdendencies
+install_st2mistral
+
+install_st2web
+
+echo "OK"
```
lakshmi-kannan: Please use same message used in ubuntu installation for less confusions.

armab:
FYI also this script worked for me under official Vagrant CentOS7:
```sh
config.vm.box = 'centos/7'
```

PR #192, opened 2016-02-29
Add new st2stream service, update gunicorn config to fix performance related issue

Kami:
This pull request adds config files and other stuff for the new ``st2stream`` service.

In addition to that, it updates gunicorn config for st2api and st2auth so they use sync worker. This change should fix the performance issues we saw in the st2api (async worker + code depending on eventlet monkey patching inside the worker doesn't work well together).

Note: It's important to use only one thread with sync worker otherwise things dead-lock because of the eventlet monkey patching we do (running more workers is fine though since they are separate processes and it's similar as running multiple api / auth processes).

If I missed something or did something wrong, please let me know.

This PR depends on the following code changes - https://github.com/StackStorm/st2/pull/2528

## TODO

- [x] Update nginx config


Kami:
What file should I update so it uses a non-master branch of st2 when building packages? I want to use `gunicorn_config_changes` branch.

dennybaa:
@Kami you should cherry-pick changes to `v1.3` branch.

PR #193, opened 2016-02-29
Follow official st2 docs [ci skip]

armab:
Get rid of inconsistencies between different installation doc versions.
The one and only one single point of truth - is official doc https://docs.stackstorm.com/install/index.html


dennybaa:
+1, but wait for official docs to appear first.

armab:
They're already appeared, see:
https://docs.stackstorm.com/install/deb.html
https://docs.stackstorm.com/install/rhel7.html
https://docs.stackstorm.com/install/rhel6.html

lakshmi-kannan:
+1

PR #194, opened 2016-02-29
Remove Bintray deployments [ci skip]

armab:
Remove Bintray deployments in favor of PackageCloud

Part of: 
https://github.com/StackStorm/st2/pull/2540
https://github.com/StackStorm/st2/pull/2539


lakshmi-kannan:
+1

PR #195, opened 2016-02-29
[WIP] actionrunner as root:st2packs, packs reownership for upgrades compatibility

dennybaa:


armab:
:+1: for including sudoers logic into packaging, so in docs https://docs.stackstorm.com/install/index.html user will be required to only generate SSH keys by hand.

lakshmi-kannan:
1. I don’t know if running actionrunner as stanley gets us anything when he has all permissions without password. https://github.com/StackStorm/st2-packages/pull/195/files#diff-2633c15d3c1376a81e049e026dac083bR5

2. If this decides to get merged, upgrade testing of packages should happen and we test it doesn't break customer's installation.

dennybaa:
@lakshmi-kannan @DoriftoShoes 
1. https://stackstorm.slack.com/archives/packages/p1456769461002857
2. anyways `chown -r stanley:st2 /opt/stackstorm/{packs,virtualenvs}` as an upgrade step, fixes this specific issue if it might happen. In fact the whole upgrade story is not clear for 1.3 => 1.4, you are talking about this). Also what was the previous configuration setup. Is that right?
3. will do, decide with 1, 2 first

cc: @enykeev he wants flow to save normally
PS. I'm suggesting to bring order and define how it should be, not how it was...)

manasdk:
-1 to this PR.

I do not think we should be creating users in a package.

manasdk:
Can we please have in the description why this is necessary?

dennybaa:
> Creating a user is sorta ok although not sure what happens if that user already exists. However, why can this not be an install step outside of packages?

All packages create users it's defacto behavior. Well the process won't fail... st2 is a good name and we should own it for creation. The stanley itself - it's ok if it will be created if it doesn't exist. NP.

> -1 to this PR. I am specifically not happy about adding a user that has passwordless sudo privileges as part of packaging.

I think you are right, agreed. Thus I will revert actionrunner to run as `root` and corresponding changes, so this PR will be slightly changed.

@manasdk , @lakshmi-kannan @armab  unless you have any other suggestions how to use actionrunner as unprivileged user. Since I see no options, actionrunner **HAS TO BE ROOT** at the moment.

dzimine:
I may be missing context here, please help: 
1) what problem is this trying to solve? specific current blocker?

On upgrade: packages have not graduated from Beta. We can get a pass for now. I am more concerned on the growing complexity of the installer. We can discuss.

dennybaa:
@dzimine as told I agree with actionrunner as root. Packs resetting permissions during packages upgrade however is required.

armab:
> Packs resetting permissions during packages upgrade however is required.

yoda_35626 1

dennybaa:
E_TOOMANY_REBASE please follow #205

PR #196, opened 2016-02-29
Remove RabbitMQ 25672 selinux port rule for el6 [ci skip]

armab:
RabbitMQ package used in instructions doesn't use `25672` port, only `5672`, which is by default allowed by policies. So the rabbitmq selinux rule is not needed for EL6.

Docs change: https://github.com/StackStorm/st2docs/pull/81


lakshmi-kannan:
Bleh. +1.

PR #172, opened 2016-02-15
Add support for different versions

DoriftoShoes:
This PR adds support for both stable and unstable repos in staging and production.  It also adds a flag to specify the version of the packages to install.

Install with defaults (staging & stable)
```
./st2bootstrap-deb.sh
```

Specify unstable
```
./st2bootstrap-deb.sh --unstable
```

Pass in a specific version
```
./st2bootstrap-deb.sh --version=1.3.2-1
```

PR #198, opened 2016-02-29
WIP - Add ability to install specific version of st2 [ci skip]

lakshmi-kannan:
## What? 

We should be able to kickoff a workflow to test packages installation on a box and perform validation. See https://github.com/StackStorm/st2cd/pull/130.

Patrick also opened a PR https://github.com/StackStorm/st2-packages/pull/172 to address this. I just picked up that work. I fixed some bugs and we are at a point where there are some problems. 

## TODO

* [X] ./st2bootstrap-deb.sh WORKS
* [X] ./st2bootstrap-deb.sh --stable WORKS
* [X] ./st2bootstrap-deb.sh --unstable DOES NOT WORK because mistral doesn't have unstable packages. See https://gist.github.com/lakshmi-kannan/a15cdc60ba00adc478c2. 
   1. [X] Generate mistral packages in unstable repo cc: @dennybaa  --> ** Script now aborts. Fix it when we have unstable packags **
* [x] ./st2bootstrap-deb.sh --version=1.3.2 ** IN PROGRESS **
   1. [x] Testing 
   2. [x] Regex validation for version string
   3. [x] When asking for dev version, automatically switch to unstable
* [X] ./st2bootstrap-deb.sh --version=1.3.2-1 NOT POSSIBLE because of mismatch in revision across packages and not a use case want to solve here.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -2,11 +2,65 @@
 
 set -eu
 
+REPO_TYPE='production'
+RELEASE='stable'
+
+APTGET_SUFFIX=''
+VERSION=''
+# XXX: Set this to '' when production becomes the shipping repo.
+STAGING=1
+
 fail() {
   echo "############### ERROR ###############"
   echo "# Failed on step - $STEP #"
   echo "#####################################"
   exit 2
+}
+
+setup_args() {
+  for i in "$@"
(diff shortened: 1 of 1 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457042787005874

PR #199, opened 2016-03-01
Connect to DB on localhost by default

enykeev:
By default, PostgreSQL binds on localhost and on some distros it might not always be 127.0.0.1.

In discussion with @armab we decided to make an exception from 'always use 127.0.0.1' rule for this particular case.


enykeev:
As @manasdk correctly pointed out, you can change the host postgres binds to in `/etc/postgresql/9.3/main/postgresql.conf`. The problem with that approach that any way you script it, it can fail to make the change and will require human intervention.

This updated config on the other hand will always work unless user previously changed said option himself.

armab:
:+1:

PR #203, opened 2016-03-02
Start Mistral services once they're installed... [ci skip]

dzimine:
it says it all.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -79,6 +79,10 @@ install_st2mistral() {
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
   # Register mistral actions
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+  
+  # Start Mistral
+  sudo service mistral-server start
+  sudo service mistral-api start
 }
 
 install_st2web() {
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -81,7 +81,7 @@ install_st2mistral() {
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
   
   # Start Mistral
-  sudo service mistral-server mistral-api start
+  sudo service mistral start
 }
 
 install_st2web() {
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -81,8 +81,7 @@ install_st2mistral() {
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
   
   # Start Mistral
-  sudo service mistral-server start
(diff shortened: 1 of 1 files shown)


scripts/st2bootstrap-deb.sh
```diff
@@ -79,6 +79,10 @@ install_st2mistral() {
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
   # Register mistral actions
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+  
+  # Start Mistral
+  sudo service mistral-server start
+  sudo service mistral-api start
 }
```
dennybaa: that won't work if `sysvinit-utils` package is not install

dzimine: i think it's ok: 
1) this won't work either https://github.com/StackStorm/st2-packages/blob/DZ/start_mistral_deb/scripts/st2bootstrap-deb.sh#L104
2) not sure if st2ctl will work. 
3) it's not typical, and if it happens folks will figure this out.

dennybaa: 1) yup right
2) st2ctl worked, but it didn't support upstart correctly, well now it does https://github.com/StackStorm/st2/pull/2551
3) maybe, but if use in your script, make sure to
`sudo apt-get install -y sysvinit-utils` along with other stuff installations, it saves from awkward failures.

armab:
+1 for starting mistral, just use one command which starts both `mistral-server` and `mistral-api`
```sh
sudo service mistral start
```
For consistency with other bootstrap scripts (`el6`/`el7`), see: 
https://github.com/StackStorm/st2-packages/blob/72c1950eab92a84af01f229eefd04dfb97578466/scripts/st2bootstrap-el6.sh#L162-L163
https://github.com/StackStorm/st2-packages/blob/72c1950eab92a84af01f229eefd04dfb97578466/scripts/st2bootstrap-el7.sh#L138-L139

armab:
@dennybaa -1 for any additional `sudo apt-get install -y sysvinit-utils`.

I think we shouldn't care too much about custom/modified distros with minimal set of tools, or it will be nightmare, since in that case it's probably not only about `sysvinit-utils`.

Also this bootstrap script codifies docs:  https://docs.stackstorm.com/install/deb.html
If possible, to avoid user overflowing with additional instructions. At least, let's try to do this at least now, at start.

dennybaa:
@armab That's why all those bootstrap scripts are evil :)

armab:
@dennybaa I agree in general :)

scripts/st2bootstrap-deb.sh
lakshmi-kannan: 2) I think your concern is st2ctl won't work here because it already started st2 services in previous step. For already started services, service foo start is a noop. So it should be fine to do st2ctl start although i prefer this.

dzimine: > sudo apt-get install -y sysvinit-utils
may be check if it's not installed and complain? IMO not a priority, and not exactly relevant to this, so I go merge, and if you decide to make this change, go for it.

Issue #202, opened 2016-03-01
`st2api not listening` regression


armab:
See the builds where `st2api` failed (service is dead): 
1.)  https://circleci.com/gh/StackStorm/st2/801
* [st2api.log](https://circle-artifacts.com/gh/StackStorm/st2/801/artifacts/0/home/ubuntu/packages/wheezy/log/st2/st2api.log) (wheezy)
* [st2api.log](https://circle-artifacts.com/gh/StackStorm/st2/801/artifacts/0/home/ubuntu/packages/jessie/log/st2/st2api.log ) (jessie)

2.) https://circleci.com/gh/StackStorm/st2-packages/501
* [st2api.log](https://circle-artifacts.com/gh/StackStorm/st2-packages/501/artifacts/0/home/ubuntu/packages/wheezy/log/st2/st2api.log) (wheezy)

In all cases our famous error (from https://github.com/StackStorm/st2-packages/issues/106):
```py
2016-03-01 17:57:26,831 27696976 ERROR base [-] Conflict while trying to save in DB.
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/persistence/base.py", line 155, in add_or_update
    model_object = cls._get_impl().add_or_update(model_object)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/models/db/__init__.py", line 174, in add_or_update
    instance.save()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/mongoengine/document.py", line 286, in save
    raise NotUniqueError(message % unicode(err))
NotUniqueError: Tried to save duplicate unique keys (E11000 duplicate key error index: st2.trigger_type_d_b.$uid_1  dup key: { : "trigger_type:core:st2.generic.actiontrigger" })
2016-03-01 17:57:26,834 27696976 ERROR triggers [-] Failed registering internal trigger: {'payload_schema': {'type': 'object', 'properties': {'status': {}, 'start_timestamp': {}, 'parameters': {}, 'action_name': {}, 'result': {}, 'execution_id': {}}}, 'description': 'Trigger encapsulating the completion of an action execution.', 'name': 'st2.generic.actiontrigger', 'pack': 'core'}.
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/triggers.py", line 75, in register_internal_trigger_types
    trigger_definition=trigger_definition)
[...]

PR #209, opened 2016-03-04
set number of workers for st2api == 1

dennybaa:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457079851006013

armab:
cc @Kami 

Let's see for a couple of days if this closes https://github.com/StackStorm/st2-packages/issues/202
If not, then [worker class](http://docs.gunicorn.org/en/stable/settings.html#worker-class) `-k sync` requires revert to old version too.

Kami:
I am simply unable to reproduce this issue locally so I'm still trying to figure out what is going on.

PR #211, opened 2016-03-05
Create ISSUE_TEMPLATE

dzimine:
So that we ask about the right info.

What are the minimal yet good set of questions to collect?


lakshmi-kannan:
This is good to start with. Please merge this to master and v1.3 branch.

lakshmi-kannan:
It would be good to get pkg versions etc for both deb and rpm. Let me work on this PR.

PR #216, opened 2016-03-07
WIP: Revert "Merge pull request #215 from dennybaa/gunicorn-native-pidfile"

lakshmi-kannan:
## What? 

We want to freeze code to generate 1.3 packages. 

We want to revert all changes in https://github.com/StackStorm/st2-packages/pull/215/files and https://github.com/StackStorm/st2-packages/pull/192/files pertaining to st2stream service. st2 1.3 branch does not contain stream code so using these changes to build 1.3 packages of st2 would break. 

I went ahead and tried using git foo to revert Denis' and Kami's PRs. It was a disaster and wasn't easy. I had to manually eye ball and fix some stuff. I'd really appreciate if @dennybaa and @Kami can double check I reverted their changes. 

TODO

* [ ] Make sure we can still build 1.3 packages of st2 with this branch
* [ ] Wire st2 release branches to use appropriate release branch of st2-packages to build native packages E.g.: v1.3 branch of st2-packages should build v1.3 branch of st2; master of st2-packages should use master branch of st2 to build packages (1.4dev in our current case). cc: @armab will do that.
* [ ] Write a document or a script to manually kickoff package generation for specific st2 version from specific branch of st2-packages. @dennybaa: please help!


lakshmi-kannan:
Bad strategy. Let me try something else.

dennybaa:
@lakshmi-kannan  well there's no urge in reverting gunicorn sysv init fixes. Just removing of st2stream was sufficient.

The changes to packages repository might be not freezed (not used) and the correcting branches as **v1.3** are not required in the best case scenario. However **features splits like addition of stream service or maybe something** else in the future still bring us to maintain the correcting branches.

In this particular case I would not unmerge gunicorn sysv pid writing fixes and other minor fixes, since first of all they are fixes and on another hand they do not introduce anything that can break packages.
Gunicorn processing itself neither ever worked (both in sync, eventlet mode) nor working now, so changes to `DAEMON_ARGS` for services don't change anything in fact (except only my fixup - the correct pid is written by sysv init)... Better if there will be correct usage scenario figured by @Kami or somebody then just proceed with just a merge of solely that parameters.

PS: Ideally if we are not adding new services to the future releases same master branch should be used for building server releases which might be `1.4 1.5 etc`, but we'll see how it'll be in real life..

PR #217, opened 2016-03-07
Remove all instances of st2stream

lakshmi-kannan:
## What? 

We want to freeze code to generate 1.3 packages. 

We want to revert all changes in https://github.com/StackStorm/st2-packages/pull/192/files pertaining to st2stream service. st2 1.3 branch does not contain stream code so using these changes to build 1.3 packages of st2 would break. 

I went ahead and tried using git foo to revert Kami's PR. It was a disaster and wasn't easy. I had to manually eye ball and fix some stuff. I'd really appreciate if @dennybaa and @Kami can double check I reverted the stream changes. 

```
vagrant@st2dev /mnt/src/storm/st2-packages (v1.3_revert_stream●)$ ag "st2stream"                                                                                                                                                          [ruby-1.9.3p484]
vagrant@st2dev /mnt/src/storm/st2-packages (v1.3_revert_stream●)$
```

TODO

* [x] Make sure we can still build 1.3 packages of st2 with this branch
* [x] Wire st2 release branches to use appropriate release branch of st2-packages to build native packages E.g.: v1.3 branch of st2-packages should build v1.3 branch of st2; master of st2-packages should use master branch of st2 to build packages (1.4dev in our current case). cc: @enykeev did this with  e38210a
* [x] Write a document or a script to manually kickoff package generation for specific st2 version from specific branch of st2-packages. @armab  is on this!


Kami:
LGTM, but I would also grep for "st2stream" and "stream" and make sure there isn't anything left :P

dennybaa:
awesome. looks good.

> Write a document or a script to manually kickoff package generation for specific st2 version from specific branch of st2-packages. @dennybaa: please help!

It's not that hard:

- we need to add v1.3 to publish https://github.com/StackStorm/st2-packages/blob/v1.3/circle.yml#L70-L73
- Then on each commit to v1.3 or manual triggering builds from https://circleci.com/gh/StackStorm/st2-packages/tree/v1.3 packages will be built in published.

enykeev:
@lakshmi-kannan I've updated circle.yml so it would checkout st2 branch with the same name as st2-package branch. Instead of trying to change branch name manually, I've created `v1.3_revert_stream` branch on st2 so please remove it when you're done here.

I couls not help but ask myself whether splitting st2 and st2-packages was good idea to begin with. It seems like we wanted to keep all package specs in one place, but later decided not to do that on a case by case basis. May be we should merge them back then?

dennybaa:
@enykeev :+1: When you'll settle with the stream, publish should be enabled ^^^.

> I could not help but ask myself whether splitting st2 and st2-packages was good idea to begin with. It seems like we wanted to keep all package specs in one place, but later decided not to do that on a case by case basis. May be we should merge them back then?

More easy to support the correcting branch as `v1.3` for st2-package only when it required. In other cases when feature split doesn't happen, I keen to use only master to build packages for several releases, this will help to bring fixes continuously.

Kami:
@enykeev 

> @lakshmi-kannan I've updated circle.yml so it would checkout st2 branch with the same name as st2-package branch. Instead of trying to change branch name manually, I've created v1.3_revert_stream branch on st2 so please remove it when you're done here.

Do we need all of that? We only plan to remove stream stuff from v1.3 packages branch and v1.3 st2 branch doesn't contain stream changes so we should be fine there.

lakshmi-kannan:
@enykeev Thanks for the circle.yml changes. Looks good to me. 

>I could not help but ask myself whether splitting st2 and st2-packages was good idea to begin with. It seems like we wanted to keep all package specs in one place, but later decided not to do that on a case by case basis. May be we should merge them back then?

I was thinking the same yesterday. It doesn't look like we have much advantage in the two repos approach now that we have more knowledge. I am still not completely +1 to combining the repos. I want to do a 1) major release 2) minor release and see if there is still any meaning for two repos approach.

lakshmi-kannan:
ok, merging now.

armab:
That's true, things going to be more and more complicated with 2 separated repos: `st2-packages` and `st2`.
As remember, tried to convince @dennybaa and @lakshmi-kannan at start about combining those.

PR #218, opened 2016-03-08
Install latest stable nginx during bootstrap

enykeev:
The change is required by https://github.com/StackStorm/st2/pull/2560 since the version from official ubuntu repos is too old and doesn't allow you to change headers on 5xx responses.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
-USERNAME='test'
+USERNAME='st2admin'
 PASSWORD='Ch@ngeMe'
 
 HUBOT_ADAPTER='slack'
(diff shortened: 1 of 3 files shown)


lakshmi-kannan:
One could argue that we will mess with customers' existing nginx installation but we anyways require a completely new box to do this. LGTM.

scripts/st2bootstrap-deb.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eu
 
-USERNAME='test'
+USERNAME='st2admin'
```
lakshmi-kannan: weird.

Kami:
Yeah, for "appliance" single box style deployments (which is what those scripts do), we should control all the software versions (nginx, mongodb, rabbitmq).

For everything else (if people want to install packages manually), we should document minmum required / supported versions of software we need.

lakshmi-kannan:
+1 to merge this to v1.3

scripts/st2bootstrap-deb.sh
enykeev: indeed.

PR #226, opened 2016-03-09
mistral services cleanups (all platforms)

### LINKED ISSUE #139: Untie `mistral-api` from `mistral` service
`service mistral start/stop` currently controls both `mistral-server` and `mistral-api` as child services.
https://github.com/StackStorm/st2-packages/blob/master/packages/mistral/debian/mistral.upstart#L14
https://github.com/StackStorm/st2-packages/blob/master/packages/mistral/debian/mistral.upstart#L17

It was a good intent behind this and more simplicity, but it breaks scaling story, when API could live on different machine.

See for more info: https://github.com/StackStorm/st2/commit/6910a2106577af18810590a1ffdb85f5bb2dd227#commitcomment-15837593 @dzimine explanation in `st2ctl` PR.

---

Implementing this will also fix: https://github.com/StackStorm/st2-packages/issues/121

dennybaa:
recent cleanups.
resolves #139


armab:
If you separate `mistral-api` and `mistral-server` in this PR (previously `service mistral`), make sure to reflect changes in :
* `st2docs` installation instructions
* `st2-packages` bootstrap scripts
* services in `st2ctl`

PR #227, opened 2016-03-09
Follow st2 branching (master)

armab:
Autodetect specific `st2` branch depending on `st2-packages` branch.

#### What this change brings
* pushing to `master` of `st2-packages` repo will generate packages based on `master` of `st2`
* pushing to `v1.3` of `st2-packages` repo will generate packages based on `v1.3` of `st2`
* opening PR against `master` branch of `st2-packages` repo will use `st2` `master`
* opening PR against `v1.3` branch of `st2-packages` repo will use `st2` `v1.3`
* opening PR against any other branch of `st2-packages` repo will use `st2` `master`

This autodetect works for both Pull Requests and inside `vX.Y` and `master` branches of `st2-packages` repo.

Branch autodetection for PR is based on target branch: 
> **armab**  wants to merge 1 commit into (--->)  `master` from `feature/branching`

PR #229, opened 2016-03-09
Follow st2 branching (v1.3)

armab:
Read: https://github.com/StackStorm/st2-packages/pull/227
(cherry picked from commit 1b2fa6b)


armab:
See: https://github.com/StackStorm/st2-packages/pull/227#issuecomment-194387100

PR #235, opened 2016-03-12
Declare unbound variables

DoriftoShoes:


dzimine:
doesn't explain why this didn't get called cc @lakshmi-kannan and find time to test more, cc myself. Unblocked now at least.
https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L102

Kami:
While we are at it, it might be a good idea to see if we can run something like this (https://github.com/koalaman/shellcheck) on this script and it provides and value and not too many false negatives.

PR #236, opened 2016-03-14
Update bootstrap v1.3 [ci skip]

DoriftoShoes:
Getting latest bootstrap script changes onto v1.3


scripts/st2bootstrap-deb.sh
```diff
@@ -79,7 +172,7 @@ EHD
 
 install_st2mistral() {
   # install mistral
-  sudo apt-get install -y st2mistral
+  sudo apt-get install -y st2mistral${ST2MISTRAL_PKG_VERSION}
```
Kami: Do we need to quote this whole string (`"st2mistral${ST2MISTRAL_PKG_VERSION}"`) to be on the safe side?

DoriftoShoes: Naw this works fine.

PR #239, opened 2016-03-15
don't use version branch for *dev versions [ci skip]

DoriftoShoes:
Currently we try to pull the bootstrap script from the corresponding version branch even on dev.  This makes us use the master branch for dev releases.


Kami:
LGTM.

Maybe not bad a idea to print something in that function if we are on a dev branch.

DoriftoShoes:
We print it at another point when RELEASE is set.

PR #243, opened 2016-03-16
Preinstall psycopg2

DoriftoShoes:


packages/st2mistral/debian/rules
```diff
@@ -33,5 +33,5 @@ override_dh_installdeb:
 	dh_installdeb
 
 override_dh_virtualenv:
-	dh_virtualenv --extra-pip-arg '--find-links=$(WHEELDIR)' \
+	dh_virtualenv --extra-pip-arg '--find-links=$(WHEELDIR)' --preinstall psycopg2 \
```
dennybaa: @DoriftoShoes better you put it here https://github.com/StackStorm/st2-packages/blob/master/packages/st2mistral/Makefile#L9-L16, so that we don't scatter everything over different places.

armab: @dennybaa see already opened PR: https://github.com/StackStorm/st2-packages/pull/250

dennybaa: awesome, I've been just moving from the bottom of my email notifications)

armab: ^^ Same thing for me :)

Example: 
To avoid something like that, I just open them in separate tab without taking immediate action.
If after going through entire history the issue/mention still worth commenting - then it's really actual and needs action :+1:

PR #245, opened 2016-03-16
RFR: CI for mistral packaging

lakshmi-kannan:
## What? 

So taking everything together: 
- push to `st2` repo will build `st2` package
- push to `mistral` repo will build `st2mistral` package
- push to `st2-packages` repo will build both `st2` and `st2mistral` packages

## TODO

* [x] circle.yml file in mistral repo https://github.com/StackStorm/mistral/pull/14
* [x] Split buildenv for st2 and mistral into two files 195f236
* [x] circle.yml change in st2 repo to call buildenv_st2.sh https://github.com/StackStorm/st2/pull/2580
* [x] docker-compose.sh cc: @dennybaa: Have a look at  
   1. [x] should use only relevant environment variables when building st2 or mistral.
See https://github.com/StackStorm/st2-packages/pull/245/commits/600a146bfae6b17d3d3a92ab709ebdec3b23eea4
   2. [x] mistral repo should only call docker-compose.sh build for st2mistral See https://github.com/StackStorm/mistral/pull/14/commits/47c18f43073a8cbb349464b40dafd311df7d4846
   3. [x] st2 repo should only call docker-compose.sh build for st2. See
https://github.com/StackStorm/st2/pull/2580/commits/00cfd5e2daedbf13cec3c13678110d829de43bab


armab:
>  Split buildenv for st2 and mistral into two files 195f236

So it looks like you would like to leave mistral rpm/deb code in this repo.

Talking about automation:
When someone opens PR in `st2-packages` repo with some fix for `st2mistral`,
how CI would know what package should it build and deploy: `st2` or `st2mistral`?

What's the plan/idea behind this case ^^?

lakshmi-kannan:
@armab 

>So it looks like you would like to leave mistral rpm/deb code in this repo.

We discussed this and in long run, it is going to make more sense to split the packages code into respective repos. 

>Talking about automation and thinking about corner cases: 
When someone opens PR in st2-packages repo with some fix for st2mistral,
how CI would know what package should it build and deploy: st2 or st2mistral?

I had the same concern https://stackstorm.slack.com/archives/packages/p1458174868000429. 
We talked about this in office a bit and for now building both st2 and st2mistral packages on a commit to st2-packages seems like a reasonable compromise until we split the packaging code.

armab:
So taking everything together: 
- push to `st2` repo will build `st2` package
- push to `mistral` repo will build `st2mistral` package
- push to `st2-packages` repo will build both `st2` and `st2mistral` packages

Do I understand it correctly?

Kami:
@armab I believe so, yes (at least that's what I would want, heh) :)

lakshmi-kannan:
@armab Correct.

circle.yml
```diff
@@ -51,7 +51,8 @@ dependencies:
   override:
     - .circle/docker-compose.sh pull ${DISTRO}
   post:
-    - .circle/docker-compose.sh build ${DISTRO}
+    - .circle/docker-compose.sh build_st2 ${DISTRO}
```
armab: This approach (2 commands) removes parallelism and brings additional overhead.
FYI `st2mistral` and `st2` packages are built in parallel and tested by Rspec tests.


You can just pass 1 env variable `ST2_PACKGES`: 

- `ST2_PACKAGES=st2mistral st2` for `st2-packages` repo
- `ST2_PACKAGES=st2` for `st2` repo
- `ST2_PACKAGES=st2mistral` for `mistral` repo

and avoid 2 commands: `build_st2` and `build_st2mistral`

armab: Also there is a typo in `ST2_PACKGES`, should be `ST2_PACKAGES`

.circle/docker-compose.sh
```diff
@@ -29,20 +29,29 @@ case "$1" in
...
+    echo Starting Packages Build for $2 ...
+    docker-compose -f docker-compose.circle.yml run \
         -e ST2MISTRAL_GITURL=${ST2MISTRAL_GITURL} \
         -e ST2MISTRAL_GITREV=${ST2MISTRAL_GITREV} \
         -e MISTRAL_VERSION=${MISTRAL_VERSION} \
         -e MISTRAL_RELEASE=${MISTRAL_RELEASE} \
         -e RABBITMQHOST=${HOST_IP} \
         -e POSTGRESHOST=${HOST_IP} \
-        -e MONGODBHOST=${HOST_IP} \
+        -e ST2_PACKGES=st2mistral \
```
armab: `ST2_PACKGES` -> `ST2_PACKAGES`

circle.yml
lakshmi-kannan: This is a good point. I'll make this fix.

manasdk:
@armab I would like that too.

circle.yml
lakshmi-kannan: See ec415de2d785c66857bc7de45d5973984a7e7802, https://github.com/StackStorm/st2/pull/2580/commits/48e8ac84e20014ec718e33ba62c96df056cf50bf and https://github.com/StackStorm/mistral/pull/14/commits/f3ff2ce353240acecfc7ede034ccf8861e557fc6

armab: Seems it will require changes by @dennybaa, probably entire rake & rspec ruby stuff is affected here:
* https://github.com/StackStorm/st2-packages/blob/master/rake/build/environment.rb
* https://github.com/StackStorm/st2-packages/blob/2c3af10fbbdbee56acf531e369f79fc4e90ee39e/rake/spec/spec_helper.rb


What I know for sure:
- in 1 container they're built in parallel
- in 1 container they're both tested (Rspec smoke tests)

circle.yml
```diff
@@ -51,7 +52,8 @@ dependencies:
   override:
     - .circle/docker-compose.sh pull ${DISTRO}
   post:
-    - .circle/docker-compose.sh build ${DISTRO}
```
armab: To be reverted when you start using `ST2_PACKAGES` env.

.circle/docker-compose.sh
```diff
@@ -43,6 +43,7 @@ case "$1" in
         -e RABBITMQHOST=${HOST_IP} \
         -e POSTGRESHOST=${HOST_IP} \
         -e MONGODBHOST=${HOST_IP} \
+        -e ST2_PACAKGES=${ST2_PACAKGES} \
```
armab: `ST2_PACAKGES` -> `ST2_PACKAGES` typo

circle.yml
```diff
@@ -18,6 +18,7 @@ machine:
     ST2_GITURL: https://github.com/StackStorm/st2
     ST2_GITREV: master
     ST2_DOCKERFILES_REPO: https://github.com/StackStorm/st2-dockerfiles
+    ST2_PACKAGES: "st2 st2mistral"
```
armab: `ST2_PACKAGES`

armab:
Current build fails because of rename: 
`.circle/buildenv.sh → .circle/buildenv_st2.sh`



`circle.yml` need touch to reflect new file names and we can see where it fails next.

armab:
Also there is another case/logic to handle: 

- PR/Push to `v1.3` branch of `st2-packages` should fetch `v1.3` branch from `st2` repo and `v1.3.2` branch of `mistral` repo.

at the same time: 

- PR/Push to `v1.3.2` branch of `mistral` repo should fetch `v1.3` branch of `st2-packages` repo.


Some autodetection is needed here.

dennybaa:
@lakshmi-kannan it's over-complicated approach which mimics st2-packages (the way it looks now). Apparently I don't mind since you picked this task. Later on I want to deprecate all bashery and heavy stuff we have now.
Basically  the will be almost no scripts, just smth like `make deb` or `make rpm` like here https://github.com/StackStorm/st2flow/blob/master/circle.yml#L33-L40 with the only difference that `make deb` will happen inside container.

lakshmi-kannan:
Alright I am a point where it seems like mistral packaging is actually broken for some reason. https://www.evernote.com/l/ANF71MG-HPtEW4ZSbQ5JRZFg5-_udQ-mMuMB/image.png

@dennybaa Can you look into fixing this? It's blocking mistral package CI work.

lakshmi-kannan:
Currently blocked on https://github.com/StackStorm/st2-packages/pull/250/files

PR #250, opened 2016-03-17
Adding psycopg2

DoriftoShoes:
psycopg2 is required to use postgres with mistral.  It doesn't get added to the requirements.txt in the mistral master branch since it isn't required upstream.

### DIFF
--- a/rake/build/environment.rb
+++ b/rake/build/environment.rb
@@ -79,8 +79,8 @@ end
 pipeopts 'st2mistral' do
   checkout true
   envpass :giturl,  'https://github.com/StackStorm/mistral', from: 'ST2MISTRAL_GITURL'
-  envpass :gitrev,  'st2-1.3.2',                             from: 'ST2MISTRAL_GITREV'
+  envpass :gitrev,  'master',                                from: 'ST2MISTRAL_GITREV'
   envpass :gitdir,  make_tmpname('mistral-')
-  envpass :mistral_version, '1.3.2'
-  envpass :mistral_release, 1
+  envpass :mistral_version, '1.4dev'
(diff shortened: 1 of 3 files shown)


armab:
https://stackstorm.slack.com/archives/packages/p1458251137000585

Error from the build log: 
```
Double requirement given: psycopg2 (from -r requirements.txt (line 116)) (already in psycopg2==2.6.1 (from -r requirements.txt (line 74)), name='psycopg2')
```

Also see env variables passed: `ST2MISTRAL_GITREV`, `MISTRAL_VERSION`

rake/build/environment.rb
```diff
@@ -79,8 +79,8 @@
 pipeopts 'st2mistral' do
   checkout true
   envpass :giturl,  'https://github.com/StackStorm/mistral', from: 'ST2MISTRAL_GITURL'
-  envpass :gitrev,  'st2-1.3.2',                             from: 'ST2MISTRAL_GITREV'
+  envpass :gitrev,  'master',                                from: 'ST2MISTRAL_GITREV'
```
armab: Here you changed `default` values, which works if you don't pass env variables like `ST2MISTRAL_GITREV`.

You need to change this part: https://github.com/StackStorm/st2-packages/blob/85fffb4e2ad16561bd5d5c50ddb7b0e2fc50d1eb/.circle/buildenv.sh#L59-L69

lakshmi-kannan: Got it as part of this https://github.com/StackStorm/st2-packages/pull/245. 

Please revert changes to this file (Done in 32188e5). I do think fixing environment.rb is needed but let's get that as part of some other PR.

PR #251, opened 2016-03-18
Make sure we copy st2-apply-rbac-definitions binary into bin

Kami:
The title says it all.

### DIFF
--- a/packages/st2/debian/st2.links
+++ b/packages/st2/debian/st2.links
@@ -2,5 +2,6 @@ opt/stackstorm/st2/bin/st2 usr/bin/st2
 opt/stackstorm/st2/bin/st2-trigger-refire usr/bin/st2-trigger-refire
 opt/stackstorm/st2/bin/st2-rule-tester usr/bin/st2-rule-tester
 opt/stackstorm/st2/bin/st2-register-content usr/bin/st2-register-content
+opt/stackstorm/st2/bin/st2-apply-rbac-definitions usr/bin/st2-apply-rbac-definitions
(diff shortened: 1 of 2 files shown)


packages/st2/debian/st2.links
```diff
@@ -2,5 +2,6 @@ opt/stackstorm/st2/bin/st2 usr/bin/st2
 opt/stackstorm/st2/bin/st2-trigger-refire usr/bin/st2-trigger-refire
 opt/stackstorm/st2/bin/st2-rule-tester usr/bin/st2-rule-tester
 opt/stackstorm/st2/bin/st2-register-content usr/bin/st2-register-content
+opt/stackstorm/st2/bin/st2-apply-rbac-definitions usr/bin/st2-apply-rbac-definitions
```
Kami: Where do we do that for RHEL?

I quickly glanced over the RPM packaging metadata and I couldn't find it.

Kami: Answering my own question - it looks like there is some magic happening in `rpmspec/helpers.spec` and `debian/st2.links` it also used for RHEL packages.

armab:
There are also tests: https://github.com/StackStorm/st2-packages/blob/2c3af10fbbdbee56acf531e369f79fc4e90ee39e/rake/spec/spec_helper.rb#L72 which checks if binary exists or not after package installation.

Kami:
@armab Updated the tests, thanks!

armab:
If we'd add failing tests first it would be `TDD` :smile:

Kami:
@armab :D

PR #252, opened 2016-03-21
Trigger a build of st2-dockerfiles on every deployment

enykeev:


armab:
With this change, you can remove now unneeded block: 

```yaml
   post:
     - .circle/docker.sh build st2
     - .circle/docker.sh build st2actionrunner st2api st2stream st2auth st2exporter st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
     - .circle/docker.sh run st2api
     - .circle/docker.sh test st2api 'st2 --version'
```

and vars

```yaml
     BUILD_DOCKER: 0
     DEPLOY_DOCKER: 0
```

from `circle.yml`

enykeev:
@armab yes, you're right. Also, seems like we don't need docker.sh anymore either.

Issue #254, opened 2016-03-24
Following server reboot /run/mistral/ is missing and Mistral does not start


jjm:
Turns out the /run/mistral directory is created during package installation, however it's a tmpfs filesystem on Ubuntu:

```bash
tmpfs on /run type tmpfs (rw,noexec,nosuid,size=10%,mode=0755)
```

This means following a server reboot Mistral will not start, by the looks of things the init.d script should check if it exists and create it....

jjm:
I'll do a PR to correct this after the AHH, the init.d scripts just need to check it exits on start of the two services and mkdir & chmod on start if missing.

armab:
Yeah, not only `mistral`, but also `st2` services affected of this bug.

Regression starting with these PRs:  
* [st2](https://github.com/StackStorm/st2-packages/pull/215/files)
* [st2mistral](https://github.com/StackStorm/st2-packages/pull/226/files)

@dennybaa FYI we avoided approach of creating `/var/run/some_dir` when added `ginicorn` for `st2api` and `st2auth` because of these ^^ reasons (it's wiped after every reboot).

Thanks @jjm!

jjm:
No worries,

Would you prefer this being a function somewhere and a call added to all the init.d scripts? Happy to change it to that later on....

armab:
@jjm yeah, function would be cool to avoid code duplication :+1:

jjm:
Indeed the PID files of ST2 are missing too, but it does not seem to prevent the services starting.

Given the discussion on PR #255 and it being a single line creating a function seems a little excessive....

jjm:
By the looks of things all of the following are also missing, most likely due to lack of permissions. 

```bash
st2.st2actionrunner-worker.init:PIDFILE=/var/run/$NAME-${WORKERID}.pid
st2.st2exporter.init:PIDFILE=/var/run/$NAME.pid
st2.st2garbagecollector.init:PIDFILE=/var/run/$NAME.pid
st2.st2notifier.init:PIDFILE=/var/run/$NAME.pid
st2.st2resultstracker.init:PIDFILE=/var/run/$NAME.pid
st2.st2rulesengine.init:PIDFILE=/var/run/$NAME.pid
st2.st2sensorcontainer.init:PIDFILE=/var/run/$NAME.pid
```

I'll move these too into an st2 directory too. Hopefully only the iniit scripts will need changed and the files are not referenced else where...

dennybaa:
[...]

PR #255, opened 2016-03-24
Check for /var/run/mistral on service start and create

jjm:
As reported in #254.

Currenlty following a server reboot, mistral will not start as the /run directory is missing. This adds a check to the init.d scripts for mistral-api and mistral-server.


packages/st2mistral/debian/mistral-api.init
```diff
@@ -20,7 +20,8 @@ DESC="mistral"
 NAME=mistral-api
 DAEMON=/opt/stackstorm/mistral/bin/gunicorn
 COMPONENTS="api,engine,executor"
-PIDFILE=/var/run/mistral/$NAME.pid
+RUN_DIR="/run/mistral"
```
armab: Use `/var/run` (old compatible way), because a symlink will be in place to `/run` if applicable.

packages/st2mistral/debian/mistral-api.init
```diff
@@ -69,6 +70,18 @@ do_start()
...
   lsb_running && return 1
+
+  # /run is a tmpfs mount, so we need to check the directory exists and create.
+  # Otherwise the service will not start following a reboot.
+  if [ ! -d "$RUN_DIR" ] ; then
+      if ! mkdir -p "$RUN_DIR" ; then
+          log_failure_msg "can't create $RUN_DIR"
+          exit 1
+      fi
+      if ! chown mistral:mistral "$RUN_DIR" ; then
```
armab: Group is `root`, from https://github.com/StackStorm/st2-packages/blob/da3b20de4f454f77c87632b15fd46b9a46e03f7f/packages/st2mistral/debian/postinst#L27

packages/st2mistral/debian/mistral-server.init
```diff
@@ -67,6 +68,19 @@ do_start()
...
+  # Otherwise the service will not start following a reboot.
+  if [ ! -d "$RUN_DIR" ] ; then
+      if ! mkdir -p "$RUN_DIR" ; then
+          log_failure_msg "can't create $RUN_DIR"
+          exit 1
+      fi
+      if ! chown mistral:mistral "$RUN_DIR" ; then
+          log_warning_msg "can't chown $RUN_DIR"
+      fi
+  fi
```
armab: This is good :+1: 
I think, we can make all that logic really simple and avoid so much verbosity.

As example, here is how `mysql` handles it: 
```sh
[ -d /var/run/mysqld ] || install -m 755 -o mysql -g root -d /var/run/mysqld
```

^^ Just a random idea.

Not sure though if `install` binary is available for all our needed platforms, but it's easy to figure out by pushing the change and check the PR status/log (our CI checks under 5 distros).

packages/st2mistral/debian/mistral-api.init
```diff
@@ -69,6 +70,18 @@ do_start()
   #   1 if daemon was already running
   #   2 if daemon could not be started
   lsb_running && return 1
+
+  # /run is a tmpfs mount, so we need to check the directory exists and create.
+  # Otherwise the service will not start following a reboot.
+  if [ ! -d "$RUN_DIR" ] ; then
+      if ! mkdir -p "$RUN_DIR" ; then
```
dennybaa: looks. good. As of `mkdir -p` it's a highly rare circumstance that it can fail. Usually it never fails, i.e. exits with 0 status even when the directory exits. Can only happen if there's not enough permissions, since init scripts are handled with root privileges, so the failure should never happen.

packages/st2mistral/debian/mistral-server.init
dennybaa: @jjm. @armab obviously suggests more elegant way, so it would be nice if it looked as above ^^ .
@armab `install` binary is always available and safe to use, since it's a part of **coreutils**.

dennybaa:
after a small refactoring, as @armab pointed, we must merge it. @jjm thank you, it's an important catch :+1:

armab:
Not to forget changes in `st2` services where applicable, as said in [#254 Issue](https://github.com/StackStorm/st2-packages/issues/254#issuecomment-200906267)
Startup scripts for the following services are using `/var/run/st2` (which is affected by this bug too):  
- `st2api`
- `st2auth`
- `st2stream`

See: https://github.com/StackStorm/st2-packages/search?utf8=%E2%9C%93&q=%22%2Fvar%2Frun%2Fst2%2F%22&type=Code


^^ If not, we'll fix it in next PR.

packages/st2mistral/debian/mistral-api.init
jjm: Thanks, this has been fixed.

packages/st2mistral/debian/mistral-api.init
jjm: Sure, this part of the change has been reverted and all PIDFILES are back in /var/run/.

packages/st2mistral/debian/mistral-server.init
jjm: Good suggestion, gone with `install` instead of the more complex statements from before.

jjm:
I've covered the init scripts in the ST2 pachage too. So hopefully all good to merge if the CI passes.

dennybaa:
@jjm overall really cool. But I'd like to point out that only "gunicorned" services, use `/var/run/st2` directory, so not to mess up up other services please rebase the changes as @armab suggested for only:
* st2api
* st2auth
* st2stream

and  mistral-api) but check your self they should be gunicorned, just picking what i remeber...

dennybaa:
in case of mistral `/var/run/mistral` though.

jjm:
Yes, from a grep of the st2 init scripts only st2.st2api.init, st2.st2auth.init & st2.st2stream.init reference gunicorn. I'll reverse the other changes.,,,

dennybaa:
yup, also plz keep updated, since sounds that @armab has found a fella bug...

jjm:
Those un-needed changes have been reverted.

Thanks for bearing with me, it' been years since I did Debian packages and so much has changed!

jjm:
So CI has passed, it's showing all is up to date from upstream.

Anything else that needs addressed or are we :+1: following the last commit?

jjm:
@armab I reverted the change to `st2.st2actionrunner-worker.init` as it was not needed..

dennybaa:
looks good to me :+1:

armab:
Yep, pretty nice :+1: 

@dennybaa Shouldn't we cover in the same way startup scripts in `rpm` as well?

See: https://github.com/StackStorm/st2-packages/blob/da3b20de4f454f77c87632b15fd46b9a46e03f7f/packages/st2mistral/rpm/mistral-api.init#L32

dennybaa:
@armab well, yeah it definitely wouldn't be an overkill to have `install /var/run/$service` pre-creatinon I think. Let's also add it to rhel inits, who knows may be we can stumble upon `tmpfs` in that camp :))

armab:
@jjm Can you resolve pull request conflict, so we can merge it?

jjm:
@armab That's the branch updated and I've fixed a typo that caused the install to always run along with making the comment more concise.

armab:
Thanks!

jjm:
@armab Your welcome!

PR #256, opened 2016-03-24
RabbitMQ and user addition for testing in circle

manasdk:


.circle/configure-test-rabbitmq.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+
+sudo sh -c 'echo "[{rabbit, [{disk_free_limit, 10}, {loopback_users, []}, {tcp_listeners, [{\"0.0.0.0\", 5672}]}]}]." > /etc/rabbitmq/rabbitmq.config'
```
armab: See: https://github.com/StackStorm/st2-packages/blob/master/.circle/configure-rabbitmq.sh

manasdk: Yup. Will try to rewrite with that approach. Guess I was being lazy :)

If you are suggesting I use that file then it will not work. Need to explicitly specific the interface to listen on else it ends up on ipv6 and then messes with test config.

If you are saying this already happens elsewhere and I need to fix that then I am happy to change existing scripts too.

So, of the three above which is your suggestion? Perhaps its somethign else entirely :)

manasdk: remove new script entirely

.circle/configure-rabbitmq.sh
```diff
@@ -2,7 +2,7 @@
 
 # Enable remote guest access
 CONFIG=$(cat <<EHD
-[{rabbit, [{disk_free_limit, 10}, {loopback_users, []}]}].
+[{rabbit, [{disk_free_limit, 10}, {loopback_users, []}, {tcp_listeners, [{"0.0.0.0", 5672}]}]}].
```
armab: Yeah :+1: 

Since there is once instance of `rabbitmq`.
And it's used by both packages (with docker) and your new unit + integration tests.

armab: Same is about `mongo` and `postgresql`.
To keep in mind when thinking about just added tests and packages testing (to avoid conflicts).

manasdk: Will do!

PR #257, opened 2016-03-25
Fix bug when `mistral` upstart invoked `api` and `server` init scripts, instead of upstart

armab:
Solves the root cause of https://github.com/StackStorm/st2-packages/issues/254 when command like: 
```
service mistral start
```
under `ubuntu trusty` invoked `init` scripts (creates `pid`), instead of using native `upstart` (doesn't create `pid`).

### DIFF
--- a/packages/st2mistral/debian/mistral-api.upstart
+++ b/packages/st2mistral/debian/mistral-api.upstart
@@ -16,6 +16,7 @@ kill timeout 60
 
 script
   NAME=mistral
+  COMPONENTS="api,engine,executor"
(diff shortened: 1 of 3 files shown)


armab:
@dennybaa take a look if I'm missing anything here

armab:
@lakshmi-kannan take a look: https://stackstorm-community.slack.com/archives/community/p1458933923002123

^^ As requested, we also need to backport this change into `v1.3` branch to fix `st2mistral` stable package, so some more work will be needed related to `CI for mistral` which was started here: https://github.com/StackStorm/st2-packages/pull/245

dennybaa:
@armab looks good, thank you!

armab:
Verified `unstable` packages, it worked.

PR #259, opened 2016-03-29
Fix bug when `mistral` upstart invoking `api` and `server` init scripts (v1.3)

armab:
Backporting mistral bugfix to `v1.3`

See: https://github.com/StackStorm/st2-packages/pull/257
(cherry picked from commit 627977d)


armab:
@jjm The mistral service startup fix is backported to `stable` 1.3 packages.
You can `apt-get update` now.

jjm:
Thanks, all looking good post reboot and reverting my local changes to the init.d scripts.

PR #262, opened 2016-04-06
Password prompt in bootstrap scripts

emedvedev:
I hate having a default username/password pair in our bootstrap script because that’s what a lot of people is going to end up using without paying much attention. It’s a security flaw, even if we’re not responsible for it.

So, in this latest RDD (Rage-Driven Development) effort I’m proposing having a username/password prompt along with options for unattended install. Everyone ok with this?

Also, I’m not sure if we’re using these scripts for CI, and if we do, the CI step should be updated with flags.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -10,6 +10,8 @@ RELEASE='stable'
 REPO_TYPE='staging'
 BETA=''
 ST2_PKG_VERSION=''
+USERNAME=''
+PASSWORD=''
 BRANCH='master'
 
 setup_args() {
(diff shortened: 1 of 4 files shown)


manasdk:
Fix https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_pkg_e2e_test.yaml#L45 as well.

emedvedev:
Updated `st2cd` with https://github.com/StackStorm/st2cd/issues/153, updated `st2build002`, ran the test: the workflow passes the correct parameters and runs fine with this branch. Merging.

Issue #264, opened 2016-04-18
Don't require tty for sudo


Kami:
AIO installer commented out `Defaults +requiretty` line in the `/etc/sudoers` file, but the packages installation instructions and current install scripts don't do that. This will cause issues for the users (`error: sudo: sorry, you must have a tty to run sudo`).

Here are the corresponding lines from workroom.

```ruby
  file_line { 'disable_tty':
    path  => '/etc/sudoers',
    match => '^Defaults\s+requiretty',
    line  => '## Defaults    requiretty',
  }
```

We need to do the same for new packages (update scripts and docs).

Kami:
It turned out we already kinda did this, but there were two problems with existing approach:

1. We didn't do it Ubuntu
2. We used a strict regex so it didn't match line with multiple spaces

Kami:
Resolves as part of #264 and https://github.com/StackStorm/st2docs/pull/118.

PR #265, opened 2016-04-18
Don't require tty for sudo

Kami:
This disables requiretty for all the users. I believe it's only enabled by default on RHEL, but it doesn't hurt to do the same on Ubuntu (if it's not enabled it's a no-op).

In theory, we could only do that for `st2` user but this could potentially cause issues if user doesn't run stuff under `st2` user, etc. and `requiretty` doesn't offer many security benefits anyway.

Note: It turned out the script already disabled it, but regex modifier was to strict (it didn't allow multiple spaces) so it wouldn't always get disabled.

This resolves part of #264.


Kami:
P.S. We need to split common functionality from those scripts in some common module which is sourced in all the scripts. All that copy and pasta made my eyes bleed :P

PR #275, opened 2016-04-20
Switch back to async eventlet based gunicorn worker for st2api and st2auth

Kami:
https://github.com/StackStorm/st2/pull/2635 + this PR fixes the gunicorn shutdown issue.

We would actually only need https://github.com/StackStorm/st2/pull/2635 to fix it, but I decided to make that change anyway.


armab:
:church: Amen!

Kami:
To clarify it - the actual issue was fixed by - https://github.com/StackStorm/st2/pull/2635

There still seems to be some bug when workers > 1 on Circle CI (tests for st2api listening fail) which I simply can't track down (it always works fine locally).

armab:
Yeah, I saw the following order: 

1) Test failed `after` adding workers > 1 and `before` merging https://github.com/StackStorm/st2/pull/2635 (same as it was here https://github.com/StackStorm/st2-packages/issues/202)
2) Tests passed `after` merging https://github.com/StackStorm/st2/pull/2635

If https://github.com/StackStorm/st2/pull/2635 really fixes it - that's really cool :+1: 
Let's see how it goes in several days.

Kami:
@armab Hm, are you sure?

Because I tried a couple of different scenarios (using 2 workers, etc.) after st2 change has been merged, but the build still failed so I reverted back to 1 worker (not ideal) for now.

You don't see other commits I used to test stuff because I rebased and squashed them back to the original commit.

armab:
Oh, I see now. I remember >1 worker commit.
You're right, my oversight then :)

PR #285, opened 2016-04-28
Register rules during bootstrap

enykeev:
Failure to do so results in broken chatops installations. Without `chatops.notify` rule present, no results are sent back to the bot.


Kami:
LGTM and good catch.

This should work fine as long as we don't ship and install `examples` pack by default anymore.

manasdk:
Registering rules by default has always been a bit contentious. As Kami said since `examples` is not installed by default this should be ok.

lakshmi-kannan:
@dzimine needs to approve this. We decided not to load rules by default.

dzimine:
+1 and even with examples, rules are "disabled" in code, I know it's overhead on trigger-instance, etc

PR #294, opened 2016-05-03
Fix st2chatops service restart in st2bootstrap-el7.sh script

martezr:
The syntax for the service restart of the st2chaptops service following
the configuration incorrectly utilized the st2chatops service name as
the first argument of the systemctl command.


lakshmi-kannan:
Good catch. Merging!

PR #295, opened 2016-05-17
Detect OS flavor and quit on unsupported version [ci skip]

lakshmi-kannan:


manasdk:
I am meh on the sleep but I see it is mostly harmless.

Issue #148, opened 2016-02-04
PR Status & Rebuild older build for `st2` repo


armab:
> Low priority
#### Problems
- If you `Rebuild` specific CircleCI build for specific PR - it always downloads `latest` code inside the `docker-compose`, see [this screenshot](http://i.imgur.com/Dku0vxq.png). It means impossible to Rebuild older builds (yesterday was the needed case).
- Even if we trigger the build in `st2` repo (meaning entire codebase is already downloaded), we still do [git clone](http://i.imgur.com/Dku0vxq.png) in compose.
#### Solution
- Option to share the codebase with `docker-compose` build instead of doing `git clone` will allow to avoid double download when in `st2` repo. Also it's a workaround for [Circle issue](https://discuss.circleci.com/t/access-real-branch-name-for-pull-request-fork-build/907) for forked PR builds.
- Another way is to specify SHA during `git clone`.
#### What it Brings
- Trigger re-build in CircleCI and be sure that we build against specific `SHA`, rather then `HEAD`.
- Enable package builds/tests for PR statuses in `st2` repo. There are already cases when packages CI/Rspec tests helped to catch bugs. See [[ 1 ]](https://stackstorm.slack.com/archives/stackstorm/p1454534560004864), [[ 2 ]](https://github.com/StackStorm/st2/pull/2458)

PR #299, opened 2016-05-19
[RFR] Getting the SHA to the deb and rpm packages

humblearner:
For details: 'https://github.com/StackStorm/st2/pull/2696'


.circle/buildenv_st2.sh
```diff
@@ -52,6 +52,8 @@ fi
 re="\\b$DISTRO\\b"
 [[ "$NOTESTS" =~ $re ]] && TESTING=0
 
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE ST2_WAITFORSTART DISTRO TESTING
+ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
```
armab: You don't need this. `CIRCLE_BUILD_URL` is env variable which is already available during build time.

See all Circle env variables: https://circleci.com/docs/environment-variables/

rake/build/package_st2.rake
```diff
@@ -33,6 +33,7 @@ namespace :package do
             opts.components.each do |component|
               execute :cp, "packages/st2/component.makefile ${GITDIR}/#{component}/Makefile"
             end
+	    execute :bash, '$GITDIR/scripts/populate-sha.sh'
```
armab: Looks like scripts like that logically belong to current `st2-packages` repo, rather then `st2` repo, since it's part of packaging. IMO.

armab: You're mixing both `tabs` with `spaces` in indentation for this line.
Please use spaces only, so it will be consistent with existing ruby code (less confusion in future).

armab:
The reason why you didn't get that passed env variable inside Docker + rake is probably because it's missed in: 

https://github.com/StackStorm/st2-packages/blob/master/rake/build/environment.rb

armab:
Apart of that, it would be good to introduce new optional env variable responsible for `SHA` fixation and pass it from CircleCI build. This way rebuild would be based on fixated git `SHA`, rather then `HEAD` (inside Docker build latest `st2` code is fetched from the git repo).

See responsible line: https://github.com/StackStorm/st2-packages/blob/2ddc0e99cb37844b79df790b4f0f93e1b5f59bb0/rake/build/upload_checkout.rake#L55

The problem described here: https://github.com/StackStorm/st2-packages/issues/148

rake/build/package_st2.rake
lakshmi-kannan: >Looks like scripts like that could be better to place in current st2-packages repo, rather then st2 repo, since it's part of packaging logic. IMO

The package.meta file will be shipped with the source. I made a call to place it in st2 repo so we could probably run this as part of st2 circle if needed. Once we figure out how to put this file in packages (deb/rpm), we can figure out where to put this script. Do you have any ideas how to achieve that? Basically, I want this fle to be in cat /opt/stackstorm/st2/package.meta after package install.

.circle/docker-compose.sh
```diff
@@ -27,6 +27,7 @@ case "$1" in
         -e RABBITMQHOST=${HOST_IP} \
         -e POSTGRESHOST=${HOST_IP} \
         -e MONGODBHOST=${HOST_IP} \
+	-e ST2_CIRCLE_URL=${CIRCLE_BUILD_URL} \
```
lakshmi-kannan: Nit: Space issue. Are you mixing tabs and spaces?

humblearner: Vim issue !! Not me :)!!

humblearner:
```
$ git rev-parse HEAD
4bac378955fe1d2

root@5e2cdb8b7c64:/tmp/st2-20160524-8-ixgfzr# dpkg -c st2_1.5dev-62_amd64.deb | grep package.meta
-rw-r--r-- root/root       116 2016-05-24 00:49 ./opt/stackstorm/st2/package.meta
lrwxrwxrwx root/root         0 2016-05-24 00:51 ./opt/stackstorm/st2/local/package.meta -> ../package.meta

$ cat st2/opt/stackstorm/st2/package.meta
[server]
version = 1.5dev
git_sha = 4bac378
circle_build_url = https://circleci.com/gh/StackStorm/st2-packages/904

```

===========================


```
$ tar -tvzf st2-1.5dev-64.x86_64.rpm | grep package.meta
-rw-r--r--  1 0      0         115 May 24 15:02 ./opt/stackstorm/st2/package.meta

$ cat st2_1.5dev-64_amd64/data/opt/stackstorm/st2/package.meta
[server]
version = 1.5dev
git_sha = 4bac378
circle_build_url = https://circleci.com/gh/StackStorm/st2-packages/904

```

lakshmi-kannan:
LGTM

PR #301, opened 2016-05-23
Fix st2bootstrap file name in echo

lakshmi-kannan:


DoriftoShoes:
Looks super fantastic to me.

manasdk:
Why the rename?

PR #305, opened 2016-05-31
[RFC] Update bootstrap scripts to write a CLI config with the default credentials

Kami:
This pull request updates bootstrap script to write a CLI config for root user and user who ran the script with the default credentials.

I think this is fine to do in case of the bootstrap script since bootstrap script is there for quick testing and evaluation and not for production deployment. In the testing / evaluation scenario, speed and UX (aka getting up and running ASAP) matters even more so it makes sense to write default credentials in the config and save user one or two command (st2 auth, exporting token).

Note: I'm also fine if we write config just for the root user and not for both.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -217,14 +217,13 @@ password = ${PASSWORD}
 EOT"
 
   # Write config for root user
-  if [ "${CURRENT_USER}" == "${ROOT_USER}"]; then
+  if [ "${CURRENT_USER}" == "${ROOT_USER}" ]; then
       return
   fi
 
   # Write config for current user (in case current user != root)
   if [ ! -d ${CURRENT_USER_CLI_CONFIG_DIRECTORY} ]; then
     sudo mkdir -p ${CURRENT_USER_CLI_CONFIG_DIRECTORY}
-    sudo chown -R ${CURRENT_USER}:${CURRENT_USER}
   fi
 
   sudo sh -c "cat <<EOT > ${CURRENT_USER_CLI_CONFIG_PATH}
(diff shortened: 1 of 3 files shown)


lakshmi-kannan:
You probably already know but you also need to edit the el6 and el7 scripts.

Kami:
@lakshmi-kannan Yeah, that was just RFC / proposal to get everyone agree on that (notably the question if we should do it for current user and root or just root).

And if and once they do, I will also update RHEL scripts :)

manasdk:
> I think this is fine to do in case of the bootstrap script since bootstrap script is there for quick testing and evaluation and not for production deployment.

We need to stop making such assumptions.

The way I understand it `st2_bootstrap` is the codification of docs therefore would you be comfortable putting this in the install section of the docs as well?

Kami:
@manasdk I don't agree.

We need to be very explicit on what the script is for and what are it's limitations, etc., otherwise people will get burned.

And yes, I don't see a problem with putting this in the docs as well.

DoriftoShoes:
The bootstrap scripts were written to be a codified version of the docs to run our documented deployment process through CI.  We seem to be deviating from that original intent.  IMO, if we add something to the script, it needs to go in the docs also.

manasdk:
@Kami You may choose not to agree with me but plenty of folks use it for production deployments. The only way not to have a script like this be used for prod deployments was to not write one.

Anyway, my argument is not that such a change should not be made to this script. In fact I am quite ok with the change as long as you are also willing to document the same steps.

Kami:
I'm fine with adding it to documentation as a convenience / optional step (basically with a notice that this step is only needed if you want CLI to automatically authenticate, cache the token, etc.).

manasdk:
@Kami Whatever makes you happy :)

Kami:
I also updated RHEL scripts and made some minor improvements (tested it on fresh box, etc.).

I will go ahead and merge it into master tomorrow.

PR #306, opened 2016-05-31
Temporary hack to fix amqp version for mistral

m4dcoder:
Inject a specific amqp version into requirements.txt.


manasdk:
I assume this will be removed when upstream is fixed?

m4dcoder:
Yes

PR #308, opened 2016-06-03
Add autocomplete to st2 cli client

enykeev:


armab:
Great idea! :+1: 
Such tiny details make user experience much better.

enykeev:
So I've tested it in our distros:
 - u14: works out the box
 - el6.7: doesn't work. Bash is too old (4.1.2<4.2.0). No `bash_complete` package.
 - el7.1: doesn't work out the box. Requires `bash_complete` to be installed manually.

That's FYI. Going to merge it anyway.

PR #309, opened 2016-06-16
Use API keys instead of username/password for st2chatops

manasdk:
Based on https://github.com/StackStorm/st2chatops/pull/41

Individual command has been validated locally. Hard to run script without corresponding `st2chatops` packages.


DoriftoShoes:
LGTM

PR #311, opened 2016-06-21
Make sure we link st2-submit-debug-info binary into /usr/bin

Kami:
The title says it all.

IIRC, this .links file is also used by rpms so if that's indeed the case we should move it outside debian/ directory to reduce the confusion (e.g. into common/ or similar).


Kami:
@armab This still doesn't seem to work (build is failing since st2-submit-debug-info binary is not present in st2 package).

Do you happen to know if I'm missing some step?

Kami:
It looks like the built package now contains that binary, but the tests still fail.

armab:
Looks like you figured it out :+1:

Kami:
Yeah, I didn't actually change anything but the test seems to be working fine now (same code didn't work before although it looked like it should have).

Anyway, I will go ahead and merge it :)

Thanks.

packages/st2/debian/st2.links
```diff
@@ -1,8 +1,10 @@
+# Note: This file is used for both - debian and rpm packages
```
manasdk: 👍

PR #315, opened 2016-06-26
st2actionrunner upstart script fails to stop all running instances

nzlosh:
I mistakenly created an issue under the st2 repo because I didn't realise the package scripts were managed under another repo.

Details of the issue can bee seen here: https://github.com/StackStorm/st2/issues/2778

The fix is to calculate the number of processors in the same way as the pre-start script.

### DIFF
--- a/packages/st2/debian/st2.st2actionrunner.upstart
+++ b/packages/st2/debian/st2.st2actionrunner.upstart
@@ -25,6 +25,7 @@ end script
 post-stop script
   NAME=st2actionrunner
   # same number as in runner.sh!
+  WORKERS=$(/usr/bin/nproc 2>/dev/null)
   WORKERS=${WORKERS:-10}
 
   # Read configuration variable file if it is present


Kami:
Thanks and again, good catch.

The change looks good to me.

/cc @lakshmi-kannan @DoriftoShoes when you two +1 it, I will go ahead and also cherry-pick it into v1.5 branch.

Kami:
Minor thing and not really related to your change - https://github.com/nzlosh/st2-packages/blob/8ba1a2d886b49646d72bc4007e814614f661a3ad/packages/st2/debian/st2.st2actionrunner.upstart#L16

To me it seems like we should default to the same number of workers if we can't retrieve number of CPUs in both places (pre and post start). In theory defaulting to more in post start could also be fine in case there are stray workers or similar, but this should never happen during a normal scenario.

lakshmi-kannan:
lgtm.

Kami:
Merged (and cherry picked into v1.5), thanks!

PR #316, opened 2016-06-28
Fix -v / --version flag in the bootstrap script

Kami:
All the options use -v foo / --value foo notation, but short option for version flag required "-v=foo" notation which is inconsistent with other options.

I assumed that was a copy and paste / type.

In addition to that, this pull request also fixes the whole version concept. It was broken and didn't work because the script passed invalid argument (--version foo instead of --version=foo) to the underlying OS specific script.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -17,7 +17,7 @@ setup_args() {
   for i in "$@"
     do
       case $i in
-          -v=*|--version=*)
+          -v|--version=*)
           VERSION="${i#*=}"
           shift
           ;;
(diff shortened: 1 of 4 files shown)


Kami:
This just fixes the notation, but it looks like the whole version concept is broken as well.

I tried supplied version v1.4.0 but it still installed v1.5.

Kami:
I pushed another commit which also fixes the whole version thing so it now works as it should (previously the version argument was simply ignored).

I will go ahead and merge this into master as soon as test pass, but we should work on the test coverage so such simple issues won't repeat again.

PR #317, opened 2016-06-30
Pickup bootstrap scripts from master

lakshmi-kannan:
Upgrade testing workflows require v1.4 specific scripts to actually work. Before this PR, EL7 scripts were broken. We just decided to roll forward the scripts to pick up fixes from master.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -8,7 +8,6 @@ RHTEST=`cat /etc/redhat-release 2> /dev/null | sed -e "s~\(.*\)release.*~\1~g"`
 VERSION=''
 RELEASE='stable'
 REPO_TYPE=''
-BETA=''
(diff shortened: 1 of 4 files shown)


Kami:
LGTM, +1

Going forward, we should probably just always use master in the future if issues like that will repeat itself often :P

lakshmi-kannan:
>Going forward, we should probably just always use master in the future if issues like that will repeat itself often :P

That has it's downsides. What if we introduced new services or new packages in newer versions.

Kami:
Yeah, the problem really is that if we want to do versioning the right way, we also need to version the main script which downloads distribution specific script.

Otherwise we really need to be careful with the changes in the main / wrapper script.

lakshmi-kannan:
>Yeah, the problem really is that if we want to do versioning the right way, we also need to version the main script which downloads distribution specific script.

Yep. Agreed.

PR #320, opened 2016-07-12
Make sure st2-self-check is installed into /usr/local/bin

Kami:
The title says it all.

This should be done for user convenience.


Kami:
Will also cherry pick it into v1.5.

PR #322, opened 2016-07-15
Switch to packagingtest container

enykeev:
Adding wheezy-backports repo to resolve `init-system-helpers` package required by `dh-systemd`.

### DIFF
--- a/docker-compose.circle.yml
+++ b/docker-compose.circle.yml
@@ -94,31 +94,31 @@ centos7build:
 ## Package testing nodes
 #
 wheezytest:
-  image: quay.io/dennybaa/droneunit:wheezy-sshd
+  image: stackstorm/packagingtest:wheezy
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 jessietest:
-  image: quay.io/dennybaa/droneunit:jessie-sshd
+  image: stackstorm/packagingtest:jessie
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 trustytest:
-  image: quay.io/dennybaa/droneunit:trusty-upstart
+  image: stackstorm/packagingtest:trusty
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 centos6test:
-  image: quay.io/dennybaa/droneunit:centos6-sshd
+  image: stackstorm/packagingtest:centos6
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 centos7test:
-  image: quay.io/dennybaa/droneunit:centos7-systemd
+  image: stackstorm/packagingtest:centos7
   extends:
     file: docker-compose.override.yml
     service: volumes-circle


Kami:
LGTM, +1

armab:
As I understand, under wheezy `dh-systemd` injects `init-system-helpers` dependency.
Package description: 
>  This package contains helper tools that are necessary for switching between the various init systems that Debian contains (e. g. sysvinit or systemd). An example is deb-systemd-helper, a script that enables systemd unit files without depending on a running systemd.

> It also includes the "service", "invoke-rc.d", and "update-rc.d" scripts which provide an abstraction for enabling, disabling, starting, and stopping services for all supported Debian init systems as specified by the policy. 

I don't think `systemd` is something that should be "emulated" under wheezy, when we ship `sysV` scripts.

As I researched, you may see an error looking this way: 
```sh
The following packages have unmet dependencies: st2 : Depends: init-system-helpers (>= 1.18~) but it is not installable
```

Asking to add every time `deb http://http.debian.net/debian wheezy-backports main` for Wheezy looks like pretty much overkill. `wheezy` shouldn't depend on any systemd stuff.


Do you think some conditional `.deb` magic is possible here? So we enable `systemd` only for `jessie` and `ubuntu16`, see this comment: https://github.com/StackStorm/st2/issues/2799#issuecomment-232496900

^^ To research & try (eg. I don't have direct answers on this).

armab:
Also there could be an issue when shipping *BOTH* `systemd` unit files and `sysV` scripts.
It's considered as a bad practice: If there are init scripts and .service files for the same service, systemd will try to execute both (service duplication).

Some resources I could find related to that problem: 
* https://fedorahosted.org/fpc/ticket/359
* https://www.freedesktop.org/wiki/Software/systemd/Incompatibilities/
* http://unix.stackexchange.com/questions/255449/how-to-create-a-deb-that-deals-with-different-init-systems

docker-compose.circle.yml
```diff
@@ -94,31 +94,31 @@ centos7build:
...
 centos6test:
-  image: quay.io/dennybaa/droneunit:centos6-sshd
+  image: stackstorm/packagingtest:centos6
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 centos7test:
-  image: quay.io/dennybaa/droneunit:centos7-systemd
+  image: stackstorm/packagingtest:centos7
```
armab: It would be much helpful to include daemon naming explicitly, eg `-systemd`, `-upstart` is pretty useful.
Dockerized `centos7-systemd` or `trusty-upstart` is not something that you can see every day.

armab:
This fact could be useful:
* https://docs.docker.com/engine/installation/linux/debian/#/debian-wheezy-stable-7-x-64-bit
* https://packages.debian.org/jessie-backports/amd64/docker.io/filelist

`Docker` `.deb` package supports `debian wheezy` only with backported `init-system-helpers` version (same as you did it). Probably might be good to research about their packaging approach.

docker-compose.circle.yml
enykeev: Yeah, I get what you're talking about, but this are not general purpose containers, we made them for one and only practical purpose - to test st2 packages against distrib's default init system. Factoring that, as long as we know it's centos7, we have no reason to assume it's something other than systemd. Am I missing something?

armab: > Am I missing something?

Yes, every distro could be changed to use different `init` system. For example, for `jessie-sshd` the default init system should be `systemd`, while we ship `sysvinit`, as I remember.

Since we deal with different init systems, It's better to say explicitly what system is used:

- Less assumptions
- Not everyone knows/remember which init system is default for this or that distro

See: https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=systemd&starCount=0

lakshmi-kannan: FWIW, I am perfectly fine with the container names enykeev has picked :). I am not against changing the names though but I feel this is too nit picky of a comment. 

For me ``sshd`` is SSH server. How it relates to init systems, I have no idea.

armab: > For me sshd is SSH server. How it relates to init systems, I have no idea.

You're right, it runs SSH server as main process, that's why it's called `jessie-sshd`, see [Dockerfile](https://github.com/dennybaa/dockerfiles/blob/master/droneunit/jessie/sshd/Dockerfile#L8): 
```sh
CMD [ "/usr/sbin/sshd", "-D" ]
```
By this rule `centos7-systemd` container runs [systemd as a main process](https://github.com/dennybaa/dockerfiles/blob/master/droneunit/centos7/systemd/Dockerfile#L25) and `trusty-upstart` runs upstart system as main PID: 1 process.

This naming is here, because it's _very unnatural_ to run init system inside the container, allowing to start/stop many services and use container like a real VM (you know that Docker is usually about 1 process per container).

If we name container just `centos7`, - it will look like the official one [centos7](https://hub.docker.com/_/centos/) and *nothing* will indicate that this image has any init system installed.

Will you figure that out if we omit that naming? Will it be harder/easier to understand if we omit that naming?
How about 6 months from now? I won't remember anything for sure.
There are just best practices in naming and ideally, I would like if we follow them for less confusion and time wasting in future.

PR #323, opened 2016-07-15
Update installer scripts to install latest version of MongoDB (3.2) by default

Kami:
This pull request updates our installer scripts to install MongoDb 3.2 by default. Going forward with StackStorm v1.6, MongoDB 3.2 will be our preferred (and only supported) version of MongoDB.

Note: Because of the way our documentation works right now (we don't version installer scripts, we just point to the latest master), this can only be merged after StackStorm v1.6 is released (or when we change / fix versioning).

## TODO

- [x] Manual test on Ubuntu 12.04
- [x] Manual test on RHEL 6
- [x] Manual test on RHEL 7
- [x] Update documentation
- [ ] Update e2e tests to use MongoDB 3.2

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -11,7 +11,9 @@ REPO_TYPE=''
 ST2_PKG_VERSION=''
 USERNAME=''
 PASSWORD=''
-BRANCH='master'
+
+# Note: This variable needs to default to a branch of the latest stable release
+BRANCH='v1.5'
 
 setup_args() {
   for i in "$@"
@@ -73,6 +75,15 @@ setup_args() {
 }
 
 setup_args $@
+
+# Note: If either --unstable or --staging flag is provided we default branch to master
+if [[ "$RELEASE" == 'unstable' ]]; then
+  BRANCH="master"
+fi
+
+if [[ "$REPO_TYPE" == 'staging' ]]; then
+  BRANCH="master"
+fi
 
 get_version_branch() {
   if [[ "$RELEASE" == 'stable' ]]; then
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -94,11 +94,15 @@ setup_args() {
 
 install_st2_dependencies() {
   sudo apt-get update
-  sudo apt-get install -y curl rabbitmq-server
+
+  # Note: gnupg-curl is needed to be able to use https transport when fetching keys
+  sudo apt-get install -y gnupg-curl
+  sudo apt-get install -y curl
+  sudo apt-get install -y rabbitmq-server
 }
 
 install_mongodb() {
-    # Add key and repo for the latest stable MongoDB (3.x)
+    # Add key and repo for the latest stable MongoDB (3.2)
   sudo apt-key adv --fetch-keys https://www.mongodb.org/static/pgp/server-3.2.asc
   sudo sh -c "cat <<EOT > /etc/apt/sources.list.d/mongodb-org-3.2.list
 deb http://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.2 multiverse
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -94,7 +94,18 @@ setup_args() {
 
 install_st2_dependencies() {
   sudo apt-get update
-  sudo apt-get install -y curl mongodb-server rabbitmq-server
+  sudo apt-get install -y curl rabbitmq-server
+}
+
+install_mongodb() {
+    # Add key and repo for the latest stable MongoDB (3.x)
+  sudo apt-key adv --fetch-keys https://www.mongodb.org/static/pgp/server-3.2.asc
+  sudo sh -c "cat <<EOT > /etc/apt/sources.list.d/mongodb-org-3.2.list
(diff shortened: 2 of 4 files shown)


Kami:
@lakshmi-kannan Let's discuss how we should roll out this change / make versioning changes in the bootstrap script.

I think we should just version all the bootstrap scripts (including the wrapper one) and update documentation so the links point directly to the versioned script. This way we can roll forward and merge this (and potentially other) changes before v1.6 final is out.

As discussed before, not (fully) versioning those scripts is a gamble anyway (and if we ever need to update script for an older version, this also shouldn't be a big deal - we simply merge / pick changes into another branch).

Kami:
I would like to roll this out as soon as possible.

One of the problems with updating wrapper script to always download distro-specific bootstrap script using a specific tag (and not from master) is that this won't work for staging and unstable channels because st2-packages branch / tag for the next release (e.g. v1.6) won't exist yet.

One simple "solution" would be to simply hard-code default value for `BRANCH` argument to the latest stable version (currently `v1.5.1`) in the wrapper script.

This is similar to what we did in the past and afaik, it should also work fine right now. We always point to the bootstrap wrapper script in master so we only always need to update version in the master so this should be fine.

Also keep in mind that this `BRANCH` argument only controls which version of distro-specific bootstrap script is downloaded and uses and not which StackStorm version is installed.

/cc @lakshmi-kannan

Kami:
Here is my proposed change - d071e0e5a283a9cf0d04e44c68e6929faf8a4d59.

This should satisfy all the criteria:

* Use latest master when testing with staging / unstable repos (e.g. v1.6dev)
* Use version specific branch when `--version` is explicitly provided (same as before)
* When no flag is provided and stable channel is used, default to the branch for the latest stable release

lakshmi-kannan:
@Kami I think this is the first time we've hit a problem that we can't merge upstream because it would break existing installations. I think the reason is the default installation of 3.2 which I think is not a good default choice. If it were me, I'd make the script take an argument --mongo3 and only then install mongo3. Since the script is used only for first time installation, just having the --mongo3 flag would've worked. 

I am not necessarily against pointing to latest version in the script but then this is another step that needs to be automated in the release workflows. Otherwise, we'd miss it. Another point is the redirect for https://stackstorm.com/packages/install.sh. Regardless of any version, the docs would point to the above fancy URL which means you still won't hit the appropriate script unless URL is also versioned. So something like ```https://stackstorm.com/packages/1.6/install.sh``` is needed. This would mean we update the URLs and docs but don't have to hardcode version in script. The URL fix is something I'd personally do because it's anyways needed. Without it, your change won't help. But note if we are making the URL fix, we've to go back and fix all the previous docs. We can decide to skip this step but just have the URLs pointing to right versions.

DoriftoShoes:
I agree on the mongo3 flag.  I think it should be optional not the default...if for no other reason than backwards compatibility.

As for the URL fix...I really hate to say it but I think Lakstorm is right.  We need to version the URLs.  I don't see a way around it other than continuing to add stuff to the wrapper and it will grow unwieldy quite quickly.

Kami:
@lakshmi-kannan Why is 3.2 as a default not a good choice? I think it's a great choice, much better than 2.x on which we have been stuck for way too long.

It works with local testing and we obviously need more automated testing and that's where setting it as default in master / v1.6dev comes in.

And 3.2 is a stable fully supported release by MongoDB (released more than half a year ago) so I see no reason to not install it by default (unless we find some critical breaking issue or whatever which is unlikely).

lakshmi-kannan:
I am not saying it isn't. In fact we should switch that to default in 1.7. For one release, we say 2.4 is default, you can install with 3.x through a flag. We call out in docs and blog post about this and then switch to mongo 3. Like patrick said, it's more from user expectations. Let's say I am running a st2 node today and I want to spin another one. My default expectation would be to see a mongo 2.4 instance.

DoriftoShoes:
Yup.  I am perfectly happy with 3.2 as the default in 1.7 but let's give the users a little heads up.

Kami:
@lakshmi-kannan @DoriftoShoes It sounds like we are still having a misunderstanding.

1. New users

New users have "no expectations" - aka inside the docs and installer script we tell users what's the best choice and install that for them. That's the whole ideal behind the installater script - simple, fast and opinionated installation.

If user wants to do a custom installation they can do whatever they want - we can't and don't force them to use 3.2 if they don't want to (as with any other things, we provide packages and documentation and it's up to them to do whatever).

2. Existing users

This change doesn't affect existing users in any way. If they have StackStorm v1.5 installed and then upgrade using `sudo apt-get install st2==1.6.0` / `sudo apt-get upgrade st2==1.6.0`.

Even if they did just `apt-get upgrade`, which is a terrible practice (but we can't really stop them, and we don't have repo per version atm, which I think we should have, but that's a topic for another day), MongoDB won't be upgraded since they will still have repos for the old version in the system.

armab:
+1 for what @Kami said. 

I had the same concerns when proposed:
> [Add .deb/.rpm `Conflicts: mongodb-server (<< 3.0.0)` after migrating to Mongo 3.x](https://github.com/StackStorm/st2-packages/issues/324)

But the main @Kami argument is that it's backwards compatible: `1.6` will work on both mongo `2` and `3`, what is pretty cool (hope that's cool enough in reality too :smile: )

If so, why not to switch to mongo3 as a preferred/recommended version in docs.

PR #326, opened 2016-07-21
Start packaging for Ubuntu 16.04

enykeev:


packages/st2/debian/st2actionrunner.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=StackStorm service st2actionrunner
+After=network.target
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/etc/sysconfig/st2actionrunner
```
armab: This is used by `RedHat`-like systems,
for `Debian`-like systems the directory with default environment variables is `/etc/default`

BTW, see: https://github.com/StackStorm/st2chatops/issues/50

armab:
@enykeev Did `systemd` worked for `debian jessie` VM?

To make sure we fix this issue: https://github.com/StackStorm/st2/issues/2799

packages/st2/debian/rules
```diff
@@ -23,6 +24,18 @@ override_dh_installdirs:
 	$(MAKE) install
 
 override_dh_installinit:
+ifeq (true, $(IS_SYSTEMD))
```
armab: :+1:

docker-compose.circle.yml
```diff
@@ -62,63 +74,78 @@ el6:
...
 wheezytest:
-  image: quay.io/dennybaa/droneunit:wheezy-sshd
+  image: stackstorm/packagingtest:wheezy
   extends:
     file: docker-compose.override.yml
     service: volumes-circle
 
 jessietest:
-  image: quay.io/dennybaa/droneunit:jessie-sshd
+  image: stackstorm/packagingtest:jessie
```
armab: Please use old container naming convention for those who runs with the `init` system or `sshd`.
See this comment: https://github.com/StackStorm/st2-packages/pull/322#discussion_r71505925

armab:
In `docker-compose.yml` and `docker-compose.circle.yml` different containers are used.

That will hurt in future, meaning local testing and circle testing have slightly different environments.

enykeev:
> @enykeev Did systemd worked for debian jessie VM?

I was under the impression Jessie was always SysV-based. We'll have to solve StackStorm/st2#2799 some other day, I'm fed up with packaging. On the other hand, you'll have something to do when DZ hires you back =)

> In docker-compose.yml and docker-compose.circle.yml different containers are used.

Yeah, I believe it was lost in rebases. I hate the whole idea you need two compose files, but that's quite another story.

> /etc/sysconfig is used by RedHat-like systems.
> For Debian-like systems the correct directory with default environment variables is /etc/default

ACK. I'll finish with the rest and then we'll get back to that.

armab:
>  I hate the whole idea you need two compose files

Yeah, that's because `CircleCI` ships with their own Mongo, RabbitMQ and PostgreSQL. No reason to waste CI build time downloading Docker images for those.

So without two compose files - more time wasting in waiting when build finishes :hourglass:

packages/st2/debian/st2actionrunner.service
lakshmi-kannan: I agree with @armab here. I'd really like this to be fixed.

circle.yml
```diff
@@ -17,8 +17,8 @@ notify:
 machine:
   # Overwrite these ENV variables in parametrized (manual/API) builds
   environment:
-    DISTROS: "wheezy jessie trusty el6 el7"
-    NOTESTS: "el7"
+    DISTROS: "wheezy jessie trusty xenial el6 el7"
+    NOTESTS: "xenial el7"
```
lakshmi-kannan: Did you figure out why el7 tests were not run? Were they breaking and we just YOLOed?

enykeev: Most likely the same reason as with xenial - you need to have a privileged container to run systemd and circleci doesn't allow you to do that. At least that's my assumption on what's going on, I don't really have deep enough understanding of either systemd or docker.

PR #328, opened 2016-07-25
Fixes for recent build container changes

enykeev:
Based on feedback from https://github.com/StackStorm/st2-packages/pull/326


armab:
Thanks for using suffixes!

PR #330, opened 2016-07-26
Reapplying container fixes

enykeev:


lakshmi-kannan:
FYI: https://github.com/StackStorm/st2-packages/pull/332/files has an overlapping change. Do you still want to merge these changes @enykeev?

PR #331, opened 2016-07-26
Refine NodeJS dependency

armab:
Since `nodejs >= 4.0` package is now a dependency for `st2chatops` (added in https://github.com/StackStorm/st2chatops/pull/46), there is no need to install `nodejs` package by hand, - just add a correct repository.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -301,9 +301,8 @@ EOT"
 }
 
 install_st2chatops() {
-  # Install Node
+  # Add NodeJS 4 repo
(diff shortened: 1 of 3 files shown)


lakshmi-kannan:
LGTM

Kami:
@armab Btw, should we also cherry-pick this into v1.6 branch or not?

armab:
@Kami That seems very minor, almost cosmetic thing.

Not sure if that's important, also docs changes are not included in `v1.6`: https://github.com/StackStorm/st2docs/commits/v1.6 so it's OK to include it in next release automatically.

PR #332, opened 2016-08-03
Use correct environment files for Debian-family systems

armab:
As said before https://github.com/StackStorm/st2-packages/pull/326#discussion_r71686505, init configs should use correct files for default environment variables.

Different distributions have different conventions where env should be placed: 
* `/etc/sysconfig` - is used for redhat family OS
* `/etc/default` - is used for debian family OS

This PR fixes it for deb.

----

I saw there were reverts in master https://github.com/StackStorm/st2-packages/pull/329.
Let's include this important fix and change once thing at a time (ENVs in one PR, container changes in another).


armab:
Also fixed some build failure cases for forked PRs

lakshmi-kannan:
LGTM. Sorry for the delay!

PR #333, opened 2016-08-11
Disable upstart logging on Ubuntu for st2 services

Kami:
This pull request disables automatic upstart logging for st2 services into `/var/log/upstart`. With newer versions of upstart, default behavior is `console log` (http://upstart.ubuntu.com/cookbook/#console-log).

We had exactly the same issue with old packages which I have fixed, but it looks like we missed this when moving to new packages.

The problem with leaving upstart logging enabled is that we have duplicated logs now and if user doesn't set up logrotate for `/var/log/upstart/` disk space could fill up. This basically means server DDoS...

I noticed this when randomly debugging build server - build server got stuck because disk filled up (https://gist.github.com/Kami/143447694bfbc8d77cbc62fc38120151).

I will cherry pick this into v1.6.0 once I verify everything is working.

### DIFF
--- a/packages/st2/debian/st2.st2actionrunner-worker.upstart
+++ b/packages/st2/debian/st2.st2actionrunner-worker.upstart
@@ -3,6 +3,9 @@
 
 description     "StackStorm actionrunner worker task"
 
+# Note: We disable upstart logging since the services manage logs themselves
+console none
+
 setuid root
 setgid st2packs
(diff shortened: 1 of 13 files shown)


Kami:
I pushed another commit to do the same for mistral services since they follow exactly the same story (Mistral manages logs itself).

PR #335, opened 2016-08-16
Added architecture check

LindsayHill:
Modified bootstrap script to check that system is 64-bit before proceeding with the install.

We document the 64-bit requirement, but people still miss that. This will at least make it fail early if they are using a 32-bit OS


armab:
@LindsayHill Please rebase on top of `master`, - that will make build green.

Besides of that, :+1:

Kami:
Yeah, nice :+1:

Kami:
Also cherry-picked this into v1.6 branch, thanks again.

PR #336, opened 2016-08-18
Generate datastore encryption key during install inside the install script

Kami:
This script generates symmetric key used for encrypting / decrypting secret values in the datastore inside the install script.

Install script is there to make it easy and fast to get up and running and playing with StackStorm. People who want more control and for production deployments, people should use packages directly and manage rest of the process themselves.

This of course also includes generating an encryption key. I will document this in the docs and make it very clear that for production deployments people should generate the key themselves. People of course also can generate a new key after the install script has finished if they wish so.


Kami:
Manually tested the changes on all the supported distros and they work OK.

At some point in the near future I will also add some checks to our CI.

PR #342, opened 2016-09-20
Fix outdated package lists during install script

m4dcoder:
The image has outdated package lists and causes install of dependencies to fail. Add update prior to install.

### DIFF
--- a/scripts/install_os_packages.sh
+++ b/scripts/install_os_packages.sh
@@ -10,8 +10,14 @@ version_delemiter() {
   [ "$(platform)" = "deb" ] && echo '_' || echo '-'
 }
 
-install_rpm() { sudo yum -y install $(lookup_fullnames $@); }
+install_rpm() {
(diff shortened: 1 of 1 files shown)


Kami:
Nice, good catch! :+1:

armab:
Despite the fact that this fast fix works, the real solution for the root cause is to update Docker environment, eg where packages are built.

Eg. the right way is to apply `upgrade` on Docker images, rebuild them, deploy to hub and idea is that build environment is always in the same state, aka immutable and doesn't depend on random version "upgrades". Upgrading packages on every build is also slower, comparing to fully pre-build Docker images.

See: https://hub.docker.com/r/stackstorm/packagingenv/tags/

Kami:
Agreed, +1.

armab:
I created an issue not to forget about upgrading original Docker Images: https://github.com/StackStorm/st2-packages/issues/343

PR #344, opened 2016-09-21
Remove the st2actionrunner worker from the post section of rpm spec

m4dcoder:
The worker is started by the main st2actionrunner service and the @ sign is causing systemctl operations to fail.

### DIFF
--- a/packages/st2/rpm/st2.spec
+++ b/packages/st2/rpm/st2.spec
@@ -49,7 +49,7 @@ Conflicts: st2common
   %include rpm/preinst_script.spec
 
 %post
-  %service_post st2actionrunner %{worker_name} st2api st2stream st2auth st2notifier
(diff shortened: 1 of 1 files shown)


DoriftoShoes:
+1

packages/st2/rpm/st2.spec
```diff
@@ -49,7 +49,7 @@ Conflicts: st2common
   %include rpm/preinst_script.spec
 
 %post
-  %service_post st2actionrunner %{worker_name} st2api st2stream st2auth st2notifier
+  %service_post st2actionrunner st2api st2stream st2auth st2notifier
```
armab: Should it be removed from the `%preun` and `%postun` sections as well?

Kami:
:+1:

PR #345, opened 2016-09-27
Dependencies requirements for RabbitMQ and MongoDB

bigmstone:
This PR adds dependencies to upstart and systemd service definitions. The change requires RabbitMQ and MongoDB be up prior to starting the service.

### DIFF
--- a/packages/st2/debian/st2.st2actionrunner.upstart
+++ b/packages/st2/debian/st2.st2actionrunner.upstart
@@ -4,7 +4,7 @@
 description     "StackStorm actionrunner task"
 author          "StackStorm Engineering <opsadmin@stackstorm.com>"
 
-start on filesystem and net-device-up IFACE!=lo
+start on rabbitmq-server-running and started mongodb filesystem and net-device-up IFACE!=lo
 stop on starting rc RUNLEVEL=[016]
 
 # Note: We disable upstart logging since the services manage logs themselves
--- a/packages/st2/debian/st2.st2api.upstart
+++ b/packages/st2/debian/st2.st2api.upstart
@@ -1,7 +1,7 @@
 description     "StackStorm st2api service"
 author          "StackStorm Engineering <opsadmin@stackstorm.com>"
 
-start on filesystem and net-device-up IFACE!=lo
+start on rabbitmq-server-running and started mongodb filesystem and net-device-up IFACE!=lo
 stop on starting rc RUNLEVEL=[016]
 
 # Note: We disable upstart logging since the services manage logs themselves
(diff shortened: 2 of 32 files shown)


bigmstone:
My only concern for this change is if a user wants to run the services independently. For instance if you're running `st2actionrunner` on multiple systems for scaling out. Worth some thought and discussion.

bigmstone:
Potential Solution to [my concern](#issuecomment-249938446) is to create a launch service which is the only enabled service by default. This service would be responsible for starting the all of the various ST2 services if you're running in single-server deployment. If a user wants to create a multi-server instance of ST2 they will be responsible for enabling the service to start on boot. This would be preferable to asking the user to change the service definition.

bigmstone:
@m4dcoder also suggested moving back to the app and having it retry AMQP connections. Exploring now how it's setup.

Kami:
@bigmstone Actually, after looking at the PR now I understand with what you mean about the multi node setup.

That's a valid point. For now, one thing we should do is document what service files changes need to be made when using multi node setup (multi node setup docs + upgrade notes docs).

I'm also wondering, since this is potentially a breaking change for people running multi node setup if we should wait and include it in v2.1 and not v2.0.1.

Kami:
Discussed this on Slack with @m4dcoder.

One thing we could do is only make the code change in v2.0.1 (IIRC, we already try on db_setup, we should make the same change for rabbitmq stuff if we don't already) and merge this change + documentation (ha section and upgrade notes) into v2.1.

armab:
What is original issue/story behind this PR? To think from the root.

----

> @m4dcoder also suggested moving back to the app and having it retry AMQP connections.

+100 :+1: 
I believe it's the smartest way for app to handle such cases.

armab:
To consider the following cases: 

- Will the st2 services start if mongo, rabbit are [external servers](https://github.com/StackStorm/chef-stackstorm/blob/master/templates/default/st2.conf.erb#L74-L76)  eg. [remote](https://github.com/StackStorm/st2/blob/master/conf/st2.package.conf#L76)?
- Will this work in Docker? (where every service/process is separated container)

If that won't work, then I'm pessimistic about the change, because it brings more limitations. Is it the point mentioned in https://github.com/StackStorm/st2-packages/pull/345#issuecomment-249938446 ?

Also with configuration management tool story (Chef/Ansible/Puppet), external mongo/rabbit/etc is pretty much usual case and configurable. Having 2 versions of startup configs for this this case looks like overhead.


BTW, there is no explicit dependencies on rabbit, mongo, postgres in packages for this particular reason  (remote services), see: https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/control#L11-L12.

bigmstone:
>If that won't work, then I'm pessimistic about the change, because it brings more limitations. Is it the point mentioned in #345 (comment) ?

Yes

Also see: https://github.com/StackStorm/st2/pull/2927

bigmstone:
Closing this PR since we will address these issues outside of service definitions.

PR #354, opened 2016-10-21
Remove default ST2_GITREV from circle.yml

enykeev:
It prevents env variable to be overloaded during manual builds and gets substituted later in scripts with $CIRCLE_BRANCH anyway.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -20,7 +20,6 @@ machine:
     DISTROS: "wheezy jessie trusty xenial el6 el7"
     NOTESTS: "xenial el7"
     ST2_GITURL: https://github.com/StackStorm/st2
-    ST2_GITREV: master
     ST2MISTRAL_GITREV: master
     ST2_DOCKERFILES_REPO: https://github.com/StackStorm/st2-dockerfiles
     ST2_PACKAGES: "st2 st2mistral"


enykeev:
Actually, it is still unclear to me what is our intention here.

On the one hand, we have [.circle/buildenv_st2.sh#L40](https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv_st2.sh#L40) which would never default to `$CIRCLE_BRANCH` because of the [circle.yml#L23](https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L23).

On the other hand, we have `st2ci.build_pkgs` [actions/workflows/build_pkgs.yaml#L29](https://github.com/StackStorm/st2ci/blob/master/actions/workflows/build_pkgs.yaml#L29) where we trying to to override `ST2_GITREV` and it's clearly not happening ([circle build 1308](https://circleci.com/gh/StackStorm/st2-packages/1308#build-parameters/containers/0)).

We're also practicing redefining ST2_GITREV for version branches [v2.0/circle.yml#L23](https://github.com/StackStorm/st2-packages/blob/v2.0/circle.yml#L23) which makes falling back `$CIRCLE_BRANCH` even less useful.

armab:
As I remember `$CIRCLE_BRANCH` works unexpected with "Forked pull requests" from contributors, showing nonexistent nonsense branch like `pull/2259`, see:
https://discuss.circleci.com/t/access-real-branch-name-for-pull-request-fork-build/907

Keep in mind that variable is used to `git clone` the `st2` codebase: `https://github.com/stackstorm/st2/tree/${ST2_GITREV}` and nonexistent branch leads to error.

Because of that, `$ST2_GITREV` is hardcoded in `circle.yml`, including version branches.

enykeev:
@armab yeah, I was afraid of that. Then we need to remove the mentions of `$CIRCLE_BRANCH` and rely on gitrev from circle.yml, but also need to fix override for manual builds.

armab:
Yeah, you're right,
in the current codebase `$CIRCLE_BRANCH` doesn't anything useful: https://github.com/StackStorm/st2-packages/search?utf8=%E2%9C%93&q=CIRCLE_BRANCH

Probably there could be any water rocks, but I don't remember anything of that from more than a half year old stuff.

We tried to rely on `$CIRCLE_BRANCH` before, but because it works really strange, - hardcoded everything (`master`, `v2.0`, etc). In the end, `$CIRCLE_BRANCH` is just abandoned mess.

Looks safe to remove :+1:

enykeev:
Closed in favor of #393

Issue #361, opened 2016-11-04
User's group failed to create on some systems


armab:
Sometimes it happens that `st2` group or `mistral` group is not created with packages.
Inexistent group results that all services fail to start (except of `st2actionrunner` which is launched from the `root:root`).

So both: 
https://github.com/StackStorm/st2-packages/blob/0542c5bcdb7dc53e5bdbba565ac59fba1774695d/packages/st2/rpm/preinst_script.spec#L24-L25
and
https://github.com/StackStorm/st2-packages/blob/0542c5bcdb7dc53e5bdbba565ac59fba1774695d/packages/st2mistral/rpm/st2mistral.spec#L45

needs better control.

----
Clean machine (working): 
```
sudo adduser --no-create-home --system st2test1
$ groups st2test1
st2test1 : st2test1
```

Here is a disconnect on affected machine: 
```
sudo adduser --no-create-home --system st2test1
$ groups st2test1
st2test1 : users
```
(probably because of custom `/etc/default/useradd` config)

Also TODO: Add additional tests where possible.

PR #362, opened 2016-11-07
Fix user's group creation for rpm

armab:
Address #361 bug

Specify `--user-group` flag for `adduser`, so the group with the same name is created for sure.

If not specified, the default group defined by `USERS_GID` in the configuration file `/etc/adduser.conf` might be created, instead of `st2` group.

> Apparently, the logic is already included in `deb` packages ([via `--group` flag](https://github.com/StackStorm/st2-packages/blob/49f3b45deb02f68aeb8ff301176602c674e7b890/packages/st2/debian/preinst#L35)), so it might be oversight in `rpm`.

### DIFF
--- a/packages/st2/rpm/preinst_script.spec
+++ b/packages/st2/rpm/preinst_script.spec
@@ -17,7 +17,7 @@ EHD
 create_users() {
   # create st2 user (services user)
   (id $ST2_USER 1>/dev/null 2>&1) ||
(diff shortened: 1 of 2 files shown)


armab:
@bigmstone That's what we debugged in #community on Friday.
Please review :eyes:

Kami:
Nice catch :+1:

PR #369, opened 2016-11-21
Adding postinst register runners

bigmstone:
On upgrade `register-runners` should be run so that ST2 will continue operating nominally.


packages/st2/rpm/st2.spec
```diff
@@ -51,6 +51,7 @@ Conflicts: st2common
 %post
   %service_post st2actionrunner st2api st2stream st2auth st2notifier
   %service_post st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
+  %include rpm/preinst_script.spec
```
Kami: I assume this should be postinst?

bigmstone: Good catch, yes.

packages/st2/debian/postinst
```diff
@@ -44,6 +44,10 @@ set_permissions() {
   echo "$fileperms" | cut -f1,2,4 -d' ' | xargs -L1 chmod
 }
 
+register_runners() {
+    st2ctl reload --register-runners
```
armab: Curious, won't this fail on st2 package upgrade if Mongo connection is not available?

bigmstone: Yes it will, but won't starting the services?

armab:
Also, should it really run `st2ctl reload --register-runners` on every package upgrade or only during upgrade from older to `>=2.1` StackStorm version?

TBH it feels really dirty to execute `st2ctl reload --register-runners` on every package upgrade: st2 could be not configured or not in that state to run `reload` and more important, - it could be unexpected for users in future to have that command on every package upgrade.

I think we should find a way to do it in a right way for future migrations, eg. one-time upgrade: 
* parse old version
* parse future version
* determine if the migration should be done

Place migrations in: 
https://github.com/StackStorm/st2/tree/master/st2common/bin/migrations

bigmstone:
Per multiple slack conversations we're merging this as a temp fix for 2.1 release. Will address a cleaner approach in future versions. This has been tested against centos and ubuntu. Will test again once new packages are built in circle. @lakshmi-kannan @armab

PR #372, opened 2016-12-01
Switch back to quay-hosted packagingenv containers

enykeev:


bigmstone:
Change looks good, but I don't think we should cherry pick it to 2.1 until at least after release.

PR #375, opened 2016-12-02
Fix PackageCloud pagination

armab:
The reason why next revision didn't work for `staging-stable` repo.

Will cherry-pick for `v2.1`

> Eventually script like that is good to rewrite into something more Python-ish.


Kami:
LGTM :+1:

Issue #365, opened 2016-11-09
Deprecate Debian Wheezy and Jessie


armab:
Closes https://github.com/StackStorm/st2/issues/2799

We actually build packages for [`debian wheezy` and `debian jessie`](https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L20). That locks `2` CircleCI containers, increases overall build time and queue.

> FYI: `Debian Wheezy` [was needed](https://github.com/StackStorm/st2-dockerfiles/issues/23
) because of its small size as base for Docker images.

These packages never reached the `stable` repos from `staging`. Maybe just 1 or 2 persons asked if we support it.

Some `deb` code is in place for them and must be cleaned up as well.

Looks like it's finally safe to remove them or there is a chance we'll need pure Debian support?

### TODO
- [x] `st2-packages` https://github.com/StackStorm/st2-packages/pull/471
  - [x] `deb` packaging changes/cleanup
- [x] `mistral` https://github.com/StackStorm/mistral/pull/30
- [x] `st2chatops` https://github.com/StackStorm/st2chatops/pull/79
- [x] `st2web` https://github.com/StackStorm/st2web/pull/382
- [x] `bwc` https://github.com/StackStorm/bwc-enterprise-packages/pull/17
- [x] `bwc-ui` https://github.com/StackStorm/bwc-ui/pull/13
- [x] `st2-auth-ldap` https://github.com/StackStorm/st2-enterprise-auth-backend-ldap/pull/34
- [x] `st2flow` https://github.com/StackStorm/st2flow/pull/121
- [x] `st2ci` https://github.com/StackStorm/st2ci/pull/101
- [x] `st2-dockerfiles` https://github.com/StackStorm/st2-dockerfiles/pull/54
  - [x] Remove Dockerfiles
  - [ ] Delete Docker Hub Images
  - [ ] Delete quay.io Images
---
- [ ] https://github.com/StackStorm/bwc-topology/pull/165
- [ ] https://github.com/StackStorm/bwc-cli/pull/64
- [ ] https://github.com/StackStorm/sandbox-actions/pull/288
- [ ] https://github.com/StackStorm/bwc-ipfabric-packs/pull/64
- [ ] https://github.com/StackStorm/ixp_essentials/pull/22
- [ ] https://github.com/StackStorm/network_essentials/pull/129
- [ ] https://github.com/StackStorm/bwc-ipfabric-suite-packages/pull/6
[...]

Issue #343, opened 2016-09-21
Build Environment, docker images are outdated


armab:
> Related to fast fix: https://github.com/StackStorm/st2-packages/pull/342

Originally packages are built in Docker environment, eg pre-bundled images for different OSes, located in https://hub.docker.com/r/stackstorm/packagingenv/tags/

These images need an update from time to time, since OS become obsolete/outdated/vulnerable.
Eg. rebuild Docker images, upgrade, deploy to Docker hub.

enykeev:
Quay is supposed to build new images for us on every change to git tags. Please confirm and close the issue.

armab:
Nice :+1: 

But more visibility would be nice.
Build happens on every git tag change in which repo? Where are links to Quay? How to know if error happens during the build? Documentation or maybe just badge to the index.

armab:
Implemented in https://github.com/StackStorm/st2-dockerfiles/issues/40

PR #377, opened 2016-12-03
Retry some commands when failed

armab:
> This PR is start of series to improve the builds in general and let them suck less
> In 90% cases what we do to fix the build: retry manually

Brings function to retry failing command in Makefile.

And while `pip` has its own connection-retrying mechanism (with `--retries <retries>` default `5`), most probably something wrong with it or it's not enough (?).
Anyways, experimenting. It's strange how frequently we have upstream connection errors like: 

----

More technical debt that is good to address in future: 
- https://github.com/StackStorm/st2-packages/issues/365 less distros/builds will decrease number of failures
- https://github.com/StackStorm/st2-packages/issues/343 to upgrade all Docker build environment, which wasn't touched for months (half of year+).

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -30,6 +30,10 @@ else
 	DEB_DISTRO := $(shell lsb_release -cs)
 endif
 
+# Makefile function to retry the failed command 'N' times
+# Example: $(call retry,3,some_script.sh)
+retry = $(2) $(foreach t,$(shell seq 1 ${1}),|| (echo -e "\033[33m Failed ($$?): '$(2)'\n Retrying $t ... \033[0m"; $(2)))
+
 .PHONY: all install wheelhouse
 all: install
 
@@ -77,7 +81,7 @@ endif
 wheelhouse: .stamp-wheelhouse
 .stamp-wheelhouse: | populate_version requirements inject-deps
 	cat requirements.txt
-	pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt
+	$(call retry,2,pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt)
 	touch $@
 
 # Note: We want to dynamically inject "st2client" dependency. This way we can
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -31,6 +31,7 @@ else
 endif
 
 # Makefile function to retry the failed command 'N' times
+# Make sure you use it for nothing but networking stuff (think about race conditons)
 # Example: $(call retry,3,some_script.sh)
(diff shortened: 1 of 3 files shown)


armab:
It worked in the battlefield!

Could catch an example of this "super-intelligent auto-remediation" :smiley: :


Here is the build where `pip` failed 5 times in different places (and was auto-retried):
https://circleci.com/gh/StackStorm/st2-packages/1483

Kami:
Nice work :+1: 

Sadly there are still too many intermediate failures not directly related to our stuff and which we can't control so this should hopefully help a lot :)

packages/st2/Makefile
```diff
@@ -30,6 +30,10 @@ else
 	DEB_DISTRO := $(shell lsb_release -cs)
 endif
 
+# Makefile function to retry the failed command 'N' times
+# Example: $(call retry,3,some_script.sh)
+retry = $(2) $(foreach t,$(shell seq 1 ${1}),|| (echo -e "\033[33m Failed ($$?): '$(2)'\n Retrying $t ... \033[0m"; $(2)))
```
Kami: Bash one liner!

I would prefer to make it just a normal function since it's more readable, but whatever :)

Kami:
Also, I don't know how easy it is to do, but it would be nice to have some kind of "test" for the retry function.

This way we can be positive that retry function which is not working correctly won't mask the failures and cause builds to succeed. I know the build would still probably fail somewhere else down the chain, but I would feel more comfortable with such "test" / check :)

packages/st2/Makefile
armab: Yeah, I don't like it too. But that's how it's usually done in make: https://www.gnu.org/software/make/manual/make.html#Call-Function

Multi-line would be even worse because of strange formatting stuff like putting `\` in the end of every line which is more error prone and then calling that function making sure positional arguments weren't messed.

I'll play with retries builds https://circleci.com/gh/StackStorm/st2-packages/tree/feature%2Fretry-on-failure more and if I'll find that `1` single command retry was enough in all cases,
maybe just:
```sh
pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt || \
  pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt
```

to remove function with echo-s and that makefile alien syntax to avoid entire complexity.

emedvedev:
MAKE BUILDS GREAT AGAIN.

armab:
So @kam @lakshmi-kannan, :-1: x `2` is too much, so I replaced it in favor of just `command || command` for simplicity.
Left `retry` function in only 1 Makefile for future reference in case we'll need it in some future, it's not used anywhere for now.

Let's see how this change behaves to get less failures and see what else we can adjust.

Kami:
LGTM :+1: 

Let's try it and see how it goes :)

PR #381, opened 2016-12-06
Fix Bootstrap script Branch

armab:
Trusty `MongoDB` and `nginx` are installed when running bootstrap in Xenial.
As a result, `st2` fails to start


armab:
Trusty `MongoDB` and `nginx` are installed when running bootstrap in Xenial.
As a result, `st2` fails to start

PR #382, opened 2016-12-06
Add a space before ]. bash syntax fix

lakshmi-kannan:
^ Setting up st2 (2.1dev-117) ...
/var/lib/dpkg/info/st2.postinst: 70: [: missing ]
Successfully updated /opt/stackstorm/st2/bin/python

Caught by CI https://st2-build/#/history/5846240727263a39c06a8e89/general


armab:
:+1:  for the fix, but :-1:  for doing things like that in packaging in future and find time to rely on DB Migration framework.

Imagine now multi-node deployment. 10 StackStorm nodes talk to external Mongo, Rabbit and so on. And now `st2` package upgrade happens, which triggers `st2ctl reload` on 10 machines simultaneously. And that happens on every package upgrade.

Instead of that, with DB migration framework running migration script many times will apply the actual migration only once (with locking/version pinning in DB).

PR #383, opened 2016-12-06
Add a space before ]. bash syntax fix

lakshmi-kannan:
Picking up https://github.com/StackStorm/st2-packages/pull/382


armab:
:+1: let's merge and see what happens, if none tested the real upgrade behavior with packages.

PR #384, opened 2016-12-08
 [WIP]: Bootstrap script for U16 status check

humblearner:
Bootstrap script for U16, to catch the CI failures:

- [x]  Verify mongod and rabbit-server status before st2 verify
- [ ]  Collect logs from aws instance


armab:
What's the difference with https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh which is already used for both `U14` and `U16`?

Issue #385, opened 2016-12-08
st2bootstrap-deb assumes variables exist, and fails if they dont


AndyMoore111:
when running st2bootstrap-deb.sh through ansible it fails with the error 'Failed on step - Configure st2 CLI config'

This is because the script assumes $HOME exists, however under ansible it doesn't seem to exist

All variables should be checked before use, and set to a default if they don't exist

fwiw i've worked around it with:

```
  if [ -z "$HOME" ];
  then
    HOME=`eval echo ~$(whoami)`
fi
```

AndyMoore111:
the -z doesnt work with set -u

changed to 

: "${HOME:=`eval echo ~$(whoami)`}"

emedvedev:
Fixed in #386. Will cherry-pick to `v2.1` as well.

armab:
Re-opening as not fully solved.

This fix https://github.com/StackStorm/st2-packages/issues/386 needs to be applied for other files as well: `st2bootstrap-el6.sh` and `st2bootstrap-el7.sh`

PR #386, opened 2016-12-09
create HOME var when it doesnt already exist

AndyMoore111:
https://github.com/StackStorm/st2-packages/issues/385

when running st2bootstrap-deb.sh through ansible it fails with the error 'Failed on step - Configure st2 CLI config'

This is because the script assumes $HOME exists, however under ansible it doesn't seem to exist

All variables should be checked before use, and set to a default if they don't exist

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -221,6 +221,8 @@ configure_st2_cli_config() {
   ROOT_USER="root"
   CURRENT_USER=$(whoami)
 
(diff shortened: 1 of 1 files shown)


emedvedev:
LGTM

armab:
@emedvedev Please don't forget to add the same fix for `st2bootstrap-el6.sh` and `st2bootstrap-el7.sh` scripts.

emedvedev:
@AndyMoore111 ^^

AndyMoore111:
https://github.com/StackStorm/st2-packages/pull/388

PR #390, opened 2016-12-15
st2 package now depends on openssh server and client

### LINKED ISSUE #379: st2 package should depend on ssh
`st2` (and `st2mistral`?) `deb` and `rpm` packages should depend on `ssh`.

> This is actually real issue reported in community when user had no `/bin/ssh` (IDK how it's possible though).

warrenvw:
Fixes #379

### DIFF
--- a/packages/st2/debian/control
+++ b/packages/st2/debian/control
@@ -9,7 +9,7 @@ Homepage: http://stackstorm.com/
 Package: st2
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.16), python2.7, ${misc:Pre-Depends}, adduser
-Depends: ${shlibs:Depends}, ${misc:Depends}, sudo, adduser, libpython-dev | python-dev, libssl-dev, libffi-dev, git, libpam0g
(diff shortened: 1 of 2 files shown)


Kami:
Re openssh-client - I believe that in theory client shouldn't be need, but it's something we should verify.

Paramiko library which is used by our remote runner to execute commands and scripts on remote hosts over SSH implements the SSH protocol itself and doesn't shell out to the ssh binary or similar.

In any case, it's probably actually also not a bad idea to also install client if we already install server :)

LGTM

/cc @armab

armab:
@Kami Good to know :+1: 
I probably've seen `ssh` binary usage somewhere in linux pack or something like that, not sure.

Additionally, as checked previously, `openssh-client` is installed by `git` package as dependency (and `st2` depends on `git`). So we just state it explicitly and clearly now in our dependencies.

ssh-server dependency looks more important here, you're right.

Kami:
@armab Oh, you are right about git used by core pack - the client we use right now shells out to git so it probably also requires OpenSSH client for SSH remotes.

PR #393, opened 2016-12-16
Allow overriding defaults with build parameters

enykeev:
Variables defined in `machine.environment` can't be overridden with build parameters. By removing this env variables, we now have better control over the build process. In particular, we can now say to circleci not to publish packages. The change is essential for new CI process since deploying every PR package to staging will make it useless for pretty much every other purpose.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,9 +24,6 @@ machine:
     ST2MISTRAL_GITREV: master
     ST2_DOCKERFILES_REPO: https://github.com/StackStorm/st2-dockerfiles
     ST2_PACKAGES: "st2 st2mistral"
-    BUILD_DOCKER: 0
-    DEPLOY_DOCKER: 0
-    DEPLOY_PACKAGES: 1
   pre:
     - mkdir -p ~/packages
     # Need latest Docker version for some features to work (CircleCI by default works with outdated version)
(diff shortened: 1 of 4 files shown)


armab:
Additionally, since we don't build/test Docker anymore and all that logic was moved to https://github.com/StackStorm/st2-dockerfiles/blob/master/circle.yml we need to clean up all the outdated Docker stuff from current `circle.yml`,
but that's probably scope of another PR.

circle.yml
```diff
@@ -15,18 +15,12 @@ notify:
     - url: https://webhooks.stackstorm.net:8531/webhooks/build/events
 
 machine:
-  # Overwrite these ENV variables in parametrized (manual/API) builds
   environment:
     DISTROS: "wheezy jessie trusty xenial el6 el7"
     NOTESTS: "xenial el7"
     ST2_GITURL: https://github.com/StackStorm/st2
-    ST2_GITREV: master
```
Kami: This makes it a bit less transparent now (for default values you need to check the scripts) so perhaps we should add a comment here or similar which points user to the right place with the default values,

Kami: Besides that, good catch and LGTM :+1:

enykeev: I do get general tone of your concern, I also do like the readability this section provides. In this particular case, though, we're talking git branch name, there should be very little confusion as to what default is.

PR #397, opened 2016-12-19
Teach st2bootstrap to install circleci artifacts

enykeev:
This change is required to be able to run tests against the packages that are yet to be deployed to staging repo.

New syntax allows user to define circleci build number with `--dev` key that would override `--unstable --staging` and instead will try to install st2 and st2mistral directly from CircleCI artifacts.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -46,6 +46,10 @@ setup_args() {
           REPO_TYPE='staging'
           shift
           ;;
+          --dev=*)
+          DEV_BUILD="${i#*=}"
+          shift
+          ;;
           --user=*)
           USERNAME="${i#*=}"
           shift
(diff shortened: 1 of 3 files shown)


scripts/st2bootstrap-deb.sh
```diff
@@ -165,8 +176,22 @@ get_full_pkg_versions() {
...
-  STEP="Get package versions" && get_full_pkg_versions && STEP="Install st2"
-  sudo apt-get install -y st2${ST2_PKG_VERSION}
+
+  if [ "$DEV_BUILD" = '' ]; then
+    STEP="Get package versions" && get_full_pkg_versions && STEP="Install st2"
+    sudo apt-get install -y st2${ST2_PKG_VERSION}
+  else
+    curl -SsL -k -o ./jq https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64
+    chmod +x ./jq
+    PACKAGE_URL="$(curl -Ss -q https://circleci.com/api/v1.1/project/github/StackStorm/st2-packages/${DEV_BUILD}/artifacts | ./jq -r '.[] | select(.path | test("/home/ubuntu/packages/${SUBTYPE}/st2_.*.deb"; "i")) | .url')"
```
enykeev: Spent good portion of the day trying to make this line to resolve ${SUBTYPE} properly in Xenial. Unsuccessfully. 

Bash black magic is killing me.

Kami:
LGTM.

Do we also need to do the same for bwc? It would also be nice to have some tests for this :)

enykeev:
> Do we also need to do the same for bwc?

I don't think we're running bwc e2e tests per package at the moment. We may decide to do it eventually. Nightlies are done against the deployed packages so the cange is not needed at the moment.

> It would also be nice to have some tests for this :)

Our e2e's supposed to be the tests for the one liner. We could probably run them nightly and check the version installed against the known build number, but that the best one I can came up at the moment.

PR #402, opened 2016-12-27
corrected to load system locale settings on Ubuntu environment

userlocalhost:
This is the correction of following issue. 
https://github.com/StackStorm/st2/issues/3107

Currently, we can't execute actions normally when an unicode-format parameter is passed on the environment of Ubuntu Trusty.

This is because the st2actionrunner doesn't load the system locale settings.

A similar problem is reported in here.
https://bugs.launchpad.net/mythbuntu/+bug/541042

This patch corrects the Upstart configuration file to load system locale settings on the pattern of the above.

After applying this patch, we can execute an action that has unicode arguments on Ubuntu like this.
```
vagrant@st2-node:~$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.5 LTS"
vagrant@st2-node:~$ cat /etc/default/locale 
# Created by cloud-init v. 0.7.5 on Wed, 14 Dec 2016 02:49:04 +0000
LANG=en_US.UTF-8
vagrant@st2-node:~$ st2 run core.local cmd='locale'
.
id: 58621a346f086f548e43be0b
status: succeeded
parameters: 
  cmd: locale
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: 'LANG=en_US.UTF-8

    LANGUAGE=

    LC_CTYPE="en_US.UTF-8"

    LC_NUMERIC="en_US.UTF-8"

    LC_TIME="en_US.UTF-8"

    LC_COLLATE="en_US.UTF-8"

    LC_MONETARY="en_US.UTF-8"

    LC_MESSAGES="en_US.UTF-8"

    LC_PAPER="en_US.UTF-8"

    LC_NAME="en_US.UTF-8"

    LC_ADDRESS="en_US.UTF-8"

    LC_TELEPHONE="en_US.UTF-8"

    LC_MEASUREMENT="en_US.UTF-8"

    LC_IDENTIFICATION="en_US.UTF-8"

    LC_ALL=en_US.UTF-8'
  succeeded: true
vagrant@st2-node:~$ st2 run core.local cmd='echo ｈｏｇｅ'
.
id: 58621a436f086f548e43be0e
status: succeeded
parameters: 
  cmd: "echo ｈｏｇｅ"
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: "ｈｏｇｅ"
  succeeded: true
vagrant@st2-node:~$ 
```

### DIFF
--- a/rake/spec/default/70-st2_actions-integrity_spec.rb
+++ b/rake/spec/default/70-st2_actions-integrity_spec.rb
@@ -7,11 +7,11 @@ describe 'st2 actions integrity checks' do
     its(:exit_status) { is_expected.to eq 0 }
   end
 
-  describe command("st2 run core.local cmd='locale'") do
+  describe command("st2 run core.local cmd=locale") do
     its(:stdout) { should match /UTF-8/ }
   end
 
-  describe command("st2 run core.local cmd='echo ¯\_(ツ)_/¯'") do
+  describe command("st2 run core.local cmd=\"echo '¯\_(ツ)_/¯'\"") do
     its(:exit_status) { is_expected.to eq 0 }
   end
 end
--- a/rake/spec/default/70-st2_actions-integrity_spec.rb
+++ b/rake/spec/default/70-st2_actions-integrity_spec.rb
@@ -6,4 +6,12 @@ describe 'st2 actions integrity checks' do
   describe command("st2 run packs.install packs=github") do
     its(:exit_status) { is_expected.to eq 0 }
   end
+
+  describe command("st2 run core.local cmd='locale'") do
+    its(:stdout) { should match /UTF-8/ }
+  end
+
+  describe command("st2 run core.local cmd='echo ｈｏｇｅ'") do
+    its(:exit_status) { is_expected.to eq 0 }
+  end
 end
--- a/rake/spec/default/70-st2_actions-integrity_spec.rb
+++ b/rake/spec/default/70-st2_actions-integrity_spec.rb
@@ -11,7 +11,7 @@ describe 'st2 actions integrity checks' do
     its(:stdout) { should match /UTF-8/ }
   end
 
-  describe command("st2 run core.local cmd='echo ｈｏｇｅ'") do
+  describe command("st2 run core.local cmd='echo ¯\_(ツ)_/¯'") do
     its(:exit_status) { is_expected.to eq 0 }
   end
 end
--- a/packages/st2/debian/st2.st2actionrunner-worker.upstart
+++ b/packages/st2/debian/st2.st2actionrunner-worker.upstart
@@ -23,5 +23,8 @@ script
   # Read configuration variable file if it is present
   [ -r /etc/default/$NAME ] && . /etc/default/$NAME
 
-  exec /opt/stackstorm/st2/bin/$NAME ${DAEMON_ARGS}
+  # Load global locale settings
+  test -f /etc/default/locale && . /etc/default/locale || true
+
+  LANG=$LANG LC_ALL=$LANG exec /opt/stackstorm/st2/bin/$NAME ${DAEMON_ARGS}
 end script
--- a/rake/spec/spec_helper.rb
+++ b/rake/spec/spec_helper.rb
@@ -17,7 +17,7 @@ SSH_OPTIONS = {
 set :backend, :ssh
 set :host, ENV['TESTNODE']
 set :ssh_options, SSH_OPTIONS
-
+set :env, LANG: 'en_US.UTF-8', LC_ALL: 'en_US.UTF-8'
 
 # ST2Spec
 class ST2Spec


Kami:
Thanks and good catch.

The change looks good to me, but I would also prefer @armab to look into it.

On a related note - it would also be good to add a test, but not sure how easy / hard that would be to do inside st2-packages without digging in.

userlocalhost:
Thank you for prompt your reply.
Sorry for sending a patch without test, I'll add it.

Kami:
No problem.

As far as tests go - it might be a bit hard to add them here so another option is to add them to our end to end tests if it turns out it's indeed hard to add them in this repo (https://github.com/stackstorm/st2tests).

armab:
Looks like you guys did a lot of work digging in into this issue.

 :+1: on change.

We can add a test similar to https://github.com/StackStorm/st2-packages/blob/master/rake/spec/default/70-st2_actions-integrity_spec.rb#L6
Just a request to use more clear unicode symbols like: 
```sh
st2 run core.local cmd='echo Bürger'
```

armab:
@userlocalhost Also to be able to use our build settings & GitHub PR status, you can disable CircleCI for your fork, there should be "stop building" button somewhere here: https://circleci.com/gh/userlocalhost/st2-packages/edit

So we'll make sure that parallelism and other settings are from `StackStorm/st2-packages` CircleCI.

rake/spec/default/70-st2_actions-integrity_spec.rb
```diff
@@ -6,4 +6,12 @@
   describe command("st2 run packs.install packs=github") do
     its(:exit_status) { is_expected.to eq 0 }
   end
+
+  describe command("st2 run core.local cmd='locale'") do
+    its(:stdout) { should match /UTF-8/ }
```
armab: nice one :+1: 

@userlocalhost If Serverspec doesn't work, I'm OK with only this test.

rake/spec/default/70-st2_actions-integrity_spec.rb
```diff
@@ -6,4 +6,12 @@
...
+  end
+
+  describe command("st2 run core.local cmd='echo ｈｏｇｅ'") do
+    its(:exit_status) { is_expected.to eq 0 }
```
armab: According to build https://circleci.com/gh/StackStorm/st2-packages/1617#tests/containers/0 this fails on [`Debian Wheezy` and `Debian Jessie`](https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L19).

The difference could be that we're using `sysV` init scripts for these 2 platforms, See:
- https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.st2actionrunner.init
- https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.st2actionrunner-worker.init

rake/spec/default/70-st2_actions-integrity_spec.rb
armab: See my next comment re Debian `Wheezy` & `Jessie`, seems like build failures are legit, not sure if that's Serverspec-related issue.

userlocalhost:
@armab 
Thank you for your review.

I could solve the failure of test in [this patch](https://github.com/StackStorm/st2-dockerfiles/pull/37) by your favor.

If you update the 'stackstorm/packagingtest' container after applying that patch, this test will be successed.

Kami:
LGTM and thanks again.

armab:
@enykeev can you help with uploading/updating new [`stackstorm/packagingtest:whezy`](https://github.com/StackStorm/st2-packages/blob/master/docker-compose.circle.yml#L115) and [`stackstorm/packagingtest-jessie`](https://github.com/StackStorm/st2-packages/blob/master/docker-compose.circle.yml#L121) Docker images?

See the change in `st2-dockerfiles`: https://github.com/StackStorm/st2-dockerfiles/pull/37

I'm on bad connection now, not sure if can do this in nearest weeks.

rake/spec/default/70-st2_actions-integrity_spec.rb
```diff
@@ -11,7 +11,7 @@
     its(:stdout) { should match /UTF-8/ }
   end
 
-  describe command("st2 run core.local cmd='echo ｈｏｇｅ'") do
+  describe command("st2 run core.local cmd='echo ¯\_(ツ)_/¯'") do
```
armab: haha, you rock :metal:

userlocalhost:
If you update the [`stackstorm/packagingtest`](https://hub.docker.com/r/stackstorm/packagingtest/tags/) containers of, the test of this patch should be passed by StackStorm/st2-dockerfiles#37.
Could you please update them? Then, I'll retry to test.

Thank you.

armab:
@userlocalhost Thank you for a reminder and sorry for waiting time.
I'll take this task early next week as a priority, since it requires a bit more work to tidy up our build and test Dockerfiles.

Besides of that, the good news are: we're going to release `2.2` soon and include this fix, so less time to see it in `st2 stable`.

armab:
I re-uploaded manually `packagingtest:wheezy` and `packagingtest:jessie` images to https://hub.docker.com/r/stackstorm/packagingtest/tags/ and restarted the build.

Let's see.

Kami:
Sweet :+1:

userlocalhost:
Thank you very much!

PR #403, opened 2017-01-04
Use mongo:3.2, sync docker-compose.yml

warrenvw:
docker-compose.circle.yml -> docker.compose.yml

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -46,7 +46,7 @@ rabbitmq:
     - "5672:5672"
 
 mongodb:
-  image: mongo
+  image: mongo:3.2
   ports:
     - "27017:27017"
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,7 +47,7 @@ xenial:
   image: quay.io/stackstorm/packagingrunner
   extends:
     file: docker-compose.override.yml
-    service: suite-circle
+    service: suite-compose
   environment:
     - BUILDNODE=xenialbuild
     - TESTNODE=xenialtest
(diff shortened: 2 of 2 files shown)


docker-compose.override.yml
```diff
@@ -46,7 +46,7 @@ rabbitmq:
     - "5672:5672"
 
 mongodb:
-  image: mongo
+  image: mongo:3.2
```
armab: :+1:

docker-compose.yml
```diff
@@ -152,18 +152,18 @@ xenialtest:
...
 
 centos6test:
-  image: quay.io/dennybaa/droneunit:centos6-sshd
+  image: stackstorm/packagingtest:centos6
   extends:
     file: docker-compose.override.yml
     service: volumes-compose
 
 centos7test:
-  image: quay.io/dennybaa/droneunit:centos7-systemd
```
armab: +1

armab:
The change looks good :+1: 

@warrenvw just don't forget to provide a bit more detailed PR description so others could understand what's going on.

In short, what we do in this PR:
- Fix `docker-compose` for local builds
  - Pin `mongo` Docker to `3.2` version (`latest` mongo Docker breaks StackStorm due to https://github.com/StackStorm/st2/issues/3100)
  - Sync-up `docker-compose.yml` (local build) with `docker-compose.circle.yml` (CircleCI build), use the same Docker images.

armab:
cc @enykeev, so when you plan to migrate to `quay.io` for build/test images, don't forget to update all `docker-compose` files as well.

docker-compose.override.yml
Kami: Good catch, indeed.

Issue #146, opened 2016-02-04
Vagrantfile


armab:
> Low priority
> Part of the task when people need more docs, transparency and `1 click` simplicity

Include in `st2-packages` Vagrantfile which would allow us to build packages for respective distro and install/test packages in isolated environment (normal Vagrant VM, instead of Docker container).

Example is simple command `vagrant up el7` which does all the rest things: 
- setup `centos/7` VM
- installs Docker, Docker-compose
- runs `docker-compose` to produce el7 packages (replace `el7` with your distro)
- optionally installs packages on Vagrant machine (note why we launched `centos/7` machine for `el7` packages)
- optionally runs self-verification and so on stuff
- allow doing that for `el6 el7 trusty xenial`

As a result:
- you have Vagrant VM with fresh packages built from sources
- you still have an option to rely on `docker-compose` which runs inside Vagrant VM if you need testing under the Docker
- you can install on vagrant VM just produced packages
- everything is in one single place

`docker-compose` story is useful, but it's absolutely different. It installs packages in the Docker container, which is not always the case we need.
Sometimes (mostly) manual testing is needed under the normal VM.

---

Based on: 
- https://github.com/StackStorm/st2-packages/issues/144
- https://stackstorm.slack.com/files/phool/F0L890P6J/yabs_sh.sh
- https://github.com/enykeev/st2testground/
- https://github.com/StackStorm/st2docs/blob/master/docs/source/install/deb.rst

dennybaa:
Yeah, that definitely would be neat.

armab:
There is another repo: https://github.com/StackStorm/st2-packages-builder which only _partially_ does what's proposed in this Issue.

Eg. if you're working on `st2-packages` codebase there is no point to clone another `st2-packages-builder` repo. It's just waste of time.

Additionally, the idea here is to spin up `u14` Vagrant which produces same platform `u14` packages or `el6` vagrant should produce `el6` packages. This allows us to install just created packages without `up`-ing another Vagrant machine with needed distro.

PR #404, opened 2017-01-05
Create Vagrant VM to build/install/test packages

warrenvw:
See #146 for more info. This initial commit does not take into account feedback from @armab.


Vagrantfile
```diff
@@ -0,0 +1,53 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+release       = ENV['RELEASE'] ? ENV['RELEASE'] : 'stable'
+hostname      = ENV['HOSTNAME'] ? ENV['HOSTNAME'] : 'st2vagrant'
+private_ip    = ENV['PRIVATE_IP'] ? ENV['PRIVATE_IP'] : '192.168.16.20'
```
armab: We definitely don't need so many confgured vars, leave only those which are 100% needed

Vagrantfile
```diff
@@ -0,0 +1,53 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+release       = ENV['RELEASE'] ? ENV['RELEASE'] : 'stable'
+hostname      = ENV['HOSTNAME'] ? ENV['HOSTNAME'] : 'st2vagrant'
```
armab: `st2-packages`, better to hardcode it, don't need to pass it via `HOSTNAME` env.

Vagrantfile
```diff
@@ -0,0 +1,53 @@
...
+      vb.cpus = 2
+    end
+
+    # NFS-synced directory
+    if !source_folder.nil? and !dest_folder.nil?
+      st2.vm.synced_folder "#{source_folder}", "#{dest_folder}", :nfs => true, :mount_options => ['nfsvers=3']
+    end
+
+    # Configure a private network
+    st2.vm.network "private_network", ip: "#{private_ip}"
```
armab: better to hardcode

Vagrantfile
```diff
@@ -0,0 +1,53 @@
...
+
+    # Public (bridged) network may come handy for external access to VM (e.g. sensor development)
+    # See https://www.vagrantup.com/docs/networking/public_network.html
+    # st2.vm.network "public_network", bridge: 'en0: Wi-Fi (AirPort)'
+    # Install docker-engine and docker-compose
+    st2.vm.provision :shell, :path => "scripts/docker.sh", :privileged => false, :args => "#{dest_folder}"
```
armab: Not sure why do we need `dest_folder`.
`scripts/docker.sh` is already available in `/vagrant` dir on the guest Vagrant machine.

That's the actual advantage of using `Vagrantfile` in the repository with code.

Vagrantfile
```diff
@@ -0,0 +1,53 @@
...
+
+    # Optionally run st2_bootstrap.sh
+    if !bootstrap.nil?
+      st2.vm.provision "shell" do |s|
+        s.path = "scripts/st2_bootstrap.sh"
```
armab: `st2_bootstrap` will install st2 from the remote location/repo, while the idea of this Vagrantfile is to install just produced locally packages (from `docker-compose`).

Vagrantfile
warrenvw: Okay.

Vagrantfile
warrenvw: Sure. Understood.

Vagrantfile
warrenvw: Yep.

Vagrantfile
```diff
@@ -0,0 +1,35 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+hostname      = 'st2-packages'
+st2user       = ENV['ST2USER'] ? ENV['ST2USER']: 'st2admin'
+st2passwd     = ENV['ST2PASSWORD'] ? ENV['ST2PASSWORD'] : 'Ch@ngeMe'
+box           = ENV['BOX'] ? ENV['BOX'] : 'centos/7'
```
armab: See this approach: https://github.com/StackStorm/ansible-st2/blob/master/Vagrantfile#L5-L22

warrenvw: Thanks! This was my next task. :)

Vagrantfile
```diff
@@ -0,0 +1,68 @@
...
+    :hostname => 'ubuntu16',
+    :box => 'ubuntu/xenial64',
+    :target => 'xenial',
+  },
+  :centos6 => {
+    :hostname => 'centos6',
+    :box => 'centos/6',
+    :target => 'el6',
+  },
+  :centos7 => {
```
armab: `vagrant up el7`, same as we define them in https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L19
don't need `:target`.

Additionally, it might be good to include in `hostname` something specific with packages, ex: `st2-packages-el7`, otherwise it could be confused with https://github.com/StackStorm/ansible-st2/blob/master/Vagrantfile#L19

warrenvw: Yes, agreed. Thanks for the comments. I have also removed the global `hostname` and `box` variables as they aren't used anymore.

scripts/setup-centos.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo "[Install] dependencies"
+sudo yum update
+
+sudo yum install -y linux-image-extra-$(uname -r)
+sudo yum install -y git curl wget
+
+echo "[Install] docker-compose"
+if [ "$TARGET" = "el6" ]; then
+  # el6 does not work with docker-compose 1.4+
```
armab: @warrenvw  Yeah, this and Kernel issues sounds like big waterblock with `EL6`.

If everything else works OK, I think we can just remove `centos/6` from the Vagrantfile list or throw an error message that EL6 is not supported to run Docker.

Considering that it's End of Production 2 phase for RHEL6: https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Product_life_cycle I think it's mid-safe for us to do something like that.

Eg. you can skip EL6 to avoid spending too much time on it.

armab: Alternative approach would be to build Packages on Ubuntu14 node and spin up respective Trusty/Xenial/EL6/EL7 Vagrant node to install just produced packages by docker-compose (if needed).

LindsayHill: I'm OK with dropping EL6 for Docker here. Am thinking about dropping EL6 altogether once it goes End of Phase 2 in May this year. 

Should be a very small subset of people that want Docker + EL6 together. Most of those folks should be Ubuntu or EL7 users.

Vagrantfile
```diff
@@ -0,0 +1,68 @@
...
+        vb.name = "#{cfg[:hostname]}"
+        vb.customize ['modifyvm', :id, '--memory', '2048']
+        vb.cpus = 2
+      end
+
+      # Sync folder using NFS
+      vm_config.vm.synced_folder '.', '/vagrant', nfs: true
+
+      # Configure a private network
+      vm_config.vm.network "private_network", ip: '192.168.16.20'
```
armab: This will try to set the same IP address for all 4 `VIRTUAL_MACHINES` we have.

Vagrantfile
```diff
@@ -0,0 +1,68 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+hostname      = 'st2-packages'
```
armab: not used, since we set hostname per multi-machine node.

scripts/setup-ubuntu.sh
```diff
@@ -0,0 +1,22 @@
...
+# TODO: Assert TARGET is one of "trusty", and "xenial"
+
+echo "[Install] dependencies"
+sudo apt-get update
+
+sudo apt-get install -y linux-image-extra-$(uname -r)
+sudo apt-get install -y git curl wget
+
+echo "[Install] docker-compose"
+sudo sh -c 'curl -sL https://github.com/docker/compose/releases/download/1.9.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose'
```
armab: Make sure we avoid downloading `docker-compose` for 2nd time if binary already exists.

Vagrantfile
```diff
@@ -0,0 +1,60 @@
...
+      vm_config.vm.network "private_network", ip: "#{cfg[:ip]}"
+
+      # Public (bridged) network may come handy for external access to VM (e.g. sensor development)
+      # See https://www.vagrantup.com/docs/networking/public_network.html
+      # st2.vm.network "public_network", bridge: 'en0: Wi-Fi (AirPort)'
+
+      # Install docker-engine
+      vm_config.vm.provision :docker
+
+      vm_config.vm.provision :shell, :path => "scripts/setup.sh", :privileged => false, :args => ["#{name}", "#{st2user}", "#{st2passwd}", "#{install}", "#{verify}"]
```
armab: `setup.sh` is very generic naming, please include something about "vagrant".

Keep in mind that `scripts/` contains bootstrap scripts used for `curl|bash` installation and just `setup.sh` in that dir could be confusing.

warrenvw: Agreed. Thanks.

lakshmi-kannan:
Please include a README.md with this so we know where to place the packages and how to kick off testing them inside vagrant env. Having the exact commands would be handy.

scripts/setup-vagrant.sh
```diff
@@ -0,0 +1,108 @@
...
+# Install docker-compose
+DC="/usr/local/bin/docker-compose"
+URL="https://github.com/docker/compose/releases/download/1.9.0/docker-compose-`uname -s`-`uname -m`"
+if [[ ! -x $DC ]]; then
+  echo "[Install] docker-compose $TARGET"
+  sudo sh -c "curl -sL $URL > $DC"
+  sudo chmod +x $DC
+fi
+
+# Using docker-compose, 1) build packages, and 2) run quick tests
```
lakshmi-kannan: We've to make sure we supporting building an arbitrary branch of st2. Without that, this would not be so useful. I don't see a way to specify the branch.

warrenvw: Good idea. I'll start that tomorrow morning.

Vagrantfile
```diff
@@ -0,0 +1,65 @@
...
+      # Box Specifications
+      vm_config.vm.provider :virtualbox do |vb|
+        vb.name = "#{cfg[:hostname]}"
+        # NOTE: With 2048MB, system slows due to kswapd. Recommend at least 4096MB.
+        vb.customize ['modifyvm', :id, '--memory', '4096']
+        vb.cpus = 2
+      end
+
+      # Sync folder using NFS
+      vm_config.vm.synced_folder '.', '/vagrant', nfs: true
```
armab: Can we comment out this as "optional"?

I think default Vagrant sync mechanism (https://www.vagrantup.com/docs/synced-folders/basic_usage.html) is more expected for "defaults" and user can fall back to another one if it doesn't work.

armab: BTW, if default Vagrant sync doesn't work for you, that might be a problem with virtualbox guest additions.

I can suggest to try this plugin https://github.com/dotless-de/vagrant-vbguest
```sh
vagrant plugin install vagrant-vbguest
```
Which automatically detects and installs needed Virtualbox Guest Additions (including version sync and requires packages) so the box works as expected.

Vagrantfile
```diff
@@ -0,0 +1,76 @@
...
+}
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  VIRTUAL_MACHINES.each do |name, cfg|
+    config.vm.define name do |vm_config|
+      vm_config.vm.hostname = cfg[:hostname]
+      vm_config.vm.box = cfg[:box]
+
+      # Give VM access to all CPU cores on the host
+      # docker-compose & rake build can benefit from more CPUs
```
armab: I added CPUs autodetection to use max on guest VM.
That's pretty important, since that ruby Rake thing in docker-compose build can benefit from it to speed up the build.

See: 
screenshot from 2017-01-25 19-10-24

scripts/setup-vagrant.sh
```diff
@@ -0,0 +1,114 @@
...
+
+if [ "$ST2_INSTALL" = "yes" ]; then
+  echo 'Install st2 packages'
+
+  # Halt the docker test environment (otherwise, the subsequent self-verification will fail)
+  sudo docker stop "vagrant_${DC_TARGET}test_1"
+
+  # Install the packages we just built
+  if [[ $ST2_TARGET != 'el7' ]]; then
+    sudo sh -c '(cd /tmp/st2-packages && dpkg -i st2*.deb)'
```
armab: On `Ubuntu Xenial` build: 
```
==> xenial: Install st2 packages
==> xenial: vagrant_xenialtest_1
==> xenial: Selecting previously unselected package st2.
==> xenial: dpkg: regarding st2_2.2dev-1_amd64.deb containing st2, pre-dependency problem:
==> xenial:  st2 pre-depends on python2.7
==> xenial:   python2.7 is not installed.
==> xenial: 
==> xenial: dpkg: error processing archive st2_2.2dev-1_amd64.deb (--install):
==> xenial:  pre-dependency problem - not installing st2
==> xenial: Selecting previously unselected package st2mistral.
==> xenial: dpkg: regarding st2mistral_2.1.0-1_amd64.deb containing st2mistral, pre-dependency problem:
==> xenial:  st2mistral pre-depends on python2.7
==> xenial:   python2.7 is not installed.
==> xenial: 
==> xenial: dpkg: error processing archive st2mistral_2.1.0-1_amd64.deb (--install):
==> xenial:  pre-dependency problem - not installing st2mistral
==> xenial: Errors were encountered while processing:
==> xenial:  st2_2.2dev-1_amd64.deb
==> xenial:  st2mistral_2.1.0-1_amd64.deb
```

Maybe try using `gdebi` which should install dependencies.

armab: Additionally, make sure that single command fail will stop entire execution.
In my case even after failed `st2` package installation the entire bash script flow went further.

warrenvw: I'll look into `gdepi`. Thanks for the suggestion. Everything seems to work as is.. but I would expect `gdepi` to allow me to clean things up.

rake/build/environment.rb
```diff
@@ -81,8 +81,8 @@
 pipeopts 'st2mistral' do
   checkout true
   envpass :giturl,  'https://github.com/StackStorm/mistral', from: 'ST2MISTRAL_GITURL'
-  envpass :gitrev,  'st2-2.1.0',                             from: 'ST2MISTRAL_GITREV'
+  envpass :gitrev,  'master',                                from: 'ST2MISTRAL_GITREV'
   envpass :gitdir,  make_tmpname('mistral-')
-  envpass :mistral_version, '2.1.0'
+  envpass :mistral_version, '2.2dev'
```
armab: Defaults generated wrong mistral package from old version: 
```sh
ls /tmp/st2-packages/
st2-2.2dev-1.x86_64.rpm  st2mistral-2.1.0-1.x86_64.rpm
```

Changed to reflect current settings for `master`.

When running build via `CircleCI` we populate them this way, see: 
- https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv_mistral.sh#L32-L34
- https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv_mistral.sh#L17
- https://github.com/StackStorm/mistral/blob/master/version_st2.py

warrenvw:
The examples.mistral_examples test at the end of st2-self-check still fails.

warrenvw:
All tests pass after merging `master` into `feature/docker_compose_in_vagrant`. The `policy.json` file present on `master` was not present in the feature branch.

PR #405, opened 2017-01-09
Use safer default values

Kami:
This pull request updates the installer script to use more sane and safer default values.

It includes the following changes:

1. Update MongoDB to listen on localhost only (this is already the case in some versions but not all of them and it's better to be safe than sorry).
2. Enable authentication for MongoDB and create two users (admin and stackstorm). "stackstorm" user only has access to "st2" database and it's used by StackStorm components to authenticate against MongoDB.
3. Update RabbitMQ config to listen on localhost only by default (previously it listened on all the interfaces)
3. Update PostgreSQL config to listen on localhost only by default  (already a default value on some distros)
4. Generate and use random password for Mistral PostgreSQL authentication.

Keep in mind that since those changes are done in the installer script, the only apply to single box installer only installation (doh). When users performs advanced / custom installation, package responsibility is only to get StackStorm components installed on the system - it's up to the user to handle authentication, configuration, etc.

## TODO

- [x] Documentation changes
  - [x] Documentation on where users can find the password when using the installer script (note) - https://github.com/StackStorm/st2docs/pull/363
  - [x] Update reference deployment and distributed / HA installation page and add a big red warning / notice which warns the user that they should enable authentication for MongoDB (point them to MongoDB docs) - https://github.com/StackStorm/st2docs/pull/363
- [x] Tests which verify those three services are listening on localhost only - https://github.com/StackStorm/st2ci/pull/71

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -430,7 +430,7 @@ install_st2mistral_depdendencies() {
   sudo apt-get install -y postgresql
 
   cat << EHD | sudo -u postgres psql
-CREATE ROLE mistral WITH CREATEDB LOGIN ENCRYPTED PASSWORD 'StackStorm';
+CREATE ROLE mistral WITH CREATEDB LOGIN ENCRYPTED PASSWORD '${ST2_POSTGRESQL_PASSWORD}';
 CREATE DATABASE mistral OWNER mistral;
 EHD
 }
@@ -449,8 +449,12 @@ install_st2mistral() {
     rm ${PACKAGE_FILENAME}
   fi
 
+  # Configure database settings
+  sudo crudini --set /etc/st2/st2.conf database connection "postgresql://mistral:${ST2_POSTGRESQL_PASSWORD}@localhost/mistral"
+
   # Setup Mistral DB tables, etc.
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
+
   # Register mistral actions
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
 
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -202,9 +202,6 @@ install_mongodb() {
   # Configure MongoDB to listen on localhost only
   sudo sed -i -e "s#bindIp:.*#bindIp: 127.0.0.1#g" /etc/mongod.conf
 
-  # Generate random password used for user authentication
-  ST2_MONGODB_PASSWORD=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 24 ; echo '')
-
   # Create admin user and user used by StackStorm
   mongo <<EOF
 use admin;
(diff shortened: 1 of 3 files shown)


Kami:
Confirmed that the changes are working and ran basic sanity tests on:

- [ ] Ubuntu 14.04, Ubuntu 16.04
- [ ] Rhel 6
- [ ] Rhel 7

Kami:
As an alternative, @enykeev suggested to to only configure services to listen / bind to localhost (without setting passwords).

I personally think doing both of them is better and setting a password for dependent services such as MongoDB, RabbitMQ and PostgreSQL is not problematic from the UX perspective, because unlike accessing StackStorm API, user shouldn't access those services directly and they are really just an "implementation detail".

If for some reason the user does need to access those services directly to debug things or similar, they can find passwords in the service config files (st2.conf, mistral.conf).

And having said that, we have already set a password for PostgreSQL database before that change, but it was a static and unsafe password.

LindsayHill:
I'm totally in favour of switching to "only listen on localhost". Only thing we should maybe do is have a KB article or reference or something about it, saying "this is all you need for single-system install. You can change it if you need to, ST2 won't care. But if you do change it, follow these security precautions..."

For passwords on single-server systems, it's slightly different. I believe that we should encourage/enforce passwords, and I believe in using randomised passwords, rather than defaults. Users don't *normally* need to access databases, so it should be transparent. But we have to make sure that our docs include notes on how to find your password, and links on how to change it.

Kami:
Yeah, I also need to point out and make it explicit that those changes (and installer script itself) is only meant for simple single-server "evaluation" / test deployments.

For advanced and production deployments it's up to the user to configure services, manage passwords, certificates, etc. We should of course document those best practices (see corresponding docs PR), but anything outside that is not really (nor should be) our responsibility.

scripts/st2bootstrap-el7.sh
```diff
@@ -442,8 +504,12 @@ install_st2mistral() {
     sudo yum -y install ${PACKAGE_URL}
   fi
 
+  # Configure database settings
+  sudo crudini --set /etc/mistral/mistral.conf database connection "postgresql://mistral:${ST2_POSTGRESQL_PASSWORD}@localhost/mistral"
```
armab: Let's use `127.0.0.1` where possible instead of `localhost` as we agreed before.

scripts/st2bootstrap-deb.sh
```diff
@@ -391,8 +469,12 @@ install_st2mistral() {
     rm ${PACKAGE_FILENAME}
   fi
 
+  # Configure database settings
+  sudo crudini --set /etc/mistral/mistral.conf database connection "postgresql://mistral:${ST2_POSTGRESQL_PASSWORD}@localhost/mistral"
```
armab: Same as in another comment:

> Let's use `127.0.0.1` where possible instead of `localhost` as we agreed before.

scripts/st2bootstrap-el7.sh
Kami: Agreed, I made that change in MongoDB and RabbitMQ, but I just followed what was already there for Mistral.

But yeah, I agree 127.0.0.1 is safer because localhost could cause issues with IPv6, etc.

dzimine:
+1 on @LindsayHill 
limiting to localhost is no-brainer,

for random password, I want to see the recovery procedure as people will need to access it (one idea is add them to script parameters and randomize only if not set by user),

for docs, let's do "Security hardening" section where we describe what needs to be done and 
refer to the script (that's what our isntallation script is doing).

👍

PR #406, opened 2017-01-11
Add policy.json to mistral for default access control

m4dcoder:
The policy.json is required by default whether authentication is enabled or not in mistral.


Mierdin:
LGTM 👍

PR #411, opened 2017-01-26
Fix to circumvent pbr related bug in mistral-db-manage populate

m4dcoder:
Init of git is added here to temporarily fix a bug in python-muranoclient during pbr version checking on OpenStack action generations.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -474,8 +474,9 @@ install_st2mistral() {
   # Setup Mistral DB tables, etc.
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
 
-  # Register mistral actions
-  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+  # Register mistral actions. Init of git is added here to temporarily fix
+  # a bug in one of the OpenStack dependency related to pbr version checking.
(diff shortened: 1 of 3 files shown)


bigmstone:
@m4dcoder What directory are we in at that point? Hard to tell from the script. Should we perhaps `cd` to a folder in `/tmp` first? I realize it's a temporary fix, but don't want all new servers `/root` to be a git repo.

bigmstone:
Never mind. See your RM now. Not sure how I missed that.

m4dcoder:
It's in the directory where the bootstrap script is launched.

scripts/st2bootstrap-deb.sh
```diff
@@ -474,8 +474,9 @@ install_st2mistral() {
   # Setup Mistral DB tables, etc.
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
 
-  # Register mistral actions
-  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+  # Register mistral actions. Init of git is added here to temporarily fix
+  # a bug in one of the OpenStack dependency related to pbr version checking.
+  git init && /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate && rm -rf .git
```
dzimine: @m4dcoder to reduce probability of this in a future and get rid of git dependency, 
would you consider upstream change in `mistral-db-manage` to make registration of openstack actions optional?

m4dcoder: Yes, my plan this week is to submit a patch to make the action registration warn instead of fail for a bad dependency.  I can implement what you suggested as well.

PR #416, opened 2017-02-06
Revert fix to mistral db populate script

m4dcoder:
Upstream mistral database management script is refactored to print the error instead of abend on import of OpenStack client modules during action registration. The fix is at https://github.com/openstack/mistral/commit/f9cf665a482c29c211bb0ade828cf29c6d78b78b.


Kami:
Much better, indeed :+1:

Issue #418, opened 2017-02-11
Upgrade on CentOS/RHEL systems deletes /etc/logrotate.d/st2


LindsayHill:
When a user runs `sudo yum upgrade st2`, it deletes `/etc/logrotate.d/st2`.

This is because the `%postun` step gets run at the end of the upgrade:

https://github.com/StackStorm/st2-packages/blob/master/packages/st2/rpm/st2.spec#L64

Additional logic is needed to only run that action when there are no more instances of st2 left on the system (generally this means it is an uninstall, not an upgrade).

armab:
I can confirm that fix worked on `EL6` during `st2` package upgrade :+1: 

```
sudo yum upgrade --debuglevel=5 --rpmverbosity=debug --errorlevel=+1 st2
```


```sh
...

D: fini      100644  1 (   0,   0)      2066 /etc/logrotate.d/st2 skip
D: fini      100755  1 (   0,   0)        91 /etc/bash_completion.d/st2 skip
D:     erase: %postun(st2-2.2dev-82.x86_64) scriptlet start
D:     erase: %postun(st2-2.2dev-82.x86_64)	execv(/bin/sh) pid 17117
+ '[' 1 -ge 1 ']'
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2actionrunner try-restart
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2actionrunner-worker try-restart
+ :
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2api try-restart
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2stream try-restart
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2auth try-restart
+ for svc in st2actionrunner st2actionrunner-worker st2api st2stream st2auth st2notifier
+ /sbin/service st2notifier try-restart
+ '[' 1 -ge 1 ']'
+ for svc in st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
+ /sbin/service st2resultstracker try-restart
+ for svc in st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
+ /sbin/service st2rulesengine try-restart
[...]

PR #419, opened 2017-02-11
Prevent logrotate removal on RPM upgrade

LindsayHill:
The `%postun` step gets run at the end of an upgrade (confusing, I know). 

So if a user has an existing ST2 system, and upgrades, we run the step that removes `/etc/logrotate.d/st2`. The correct approach is to check how many packages will be left on the system after this operation, and only run the removal step if this is the removal of the final version of st2. 

All users who have installed previous versions of ST2, and then upgraded, will no longer have a log rotate configuration. After this change gets merged, and they upgrade again, they will have a working logrotate config again. 

As a workaround, they can manually copy https://github.com/StackStorm/st2/blob/master/conf/logrotate.conf to `/etc/logrotate.d/conf` if they do not wish to upgrade.

Related issues: https://github.com/StackStorm/st2/issues/3213 and https://github.com/StackStorm/st2-packages/issues/418

### DIFF
--- a/packages/st2/rpm/st2.spec
+++ b/packages/st2/rpm/st2.spec
@@ -60,8 +60,10 @@ Conflicts: st2common
 %postun
   %service_postun st2actionrunner %{worker_name} st2api st2stream st2auth st2notifier
   %service_postun st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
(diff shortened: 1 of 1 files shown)


armab:
@LindsayHill caught writing code :)

Kami:
LGTM, let's merge it for v2.2.0, but let's also make sure we test it.

PR #420, opened 2017-02-14
Do not install linux-image-extra

warrenvw:
Additionally, the gdebi-core package does not exist on centos (el7).
So while attempting to install the package did not cause any failure,
we may as well not even try.


armab:
Yeah, obviously gdebi is only for installing deb.

PR #422, opened 2017-02-16
Install net-tools on el7 (provides netstat)

warrenvw:
In the Vagrant centos/7 image, netstat is not found. This is provided by the net-tools package.

### DIFF
--- a/scripts/st2bootstrap-el7.sh
+++ b/scripts/st2bootstrap-el7.sh
@@ -207,6 +207,9 @@ function port_status() {
 }
 
 check_st2_host_dependencies() {
+  # Install netstat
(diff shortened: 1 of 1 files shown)


armab:
Seems everything is clear here :+1:
@warrenvw I'm merging it ASAP to have that fix in `master`.

warrenvw:
Thanks @armab

LindsayHill:
Didn't want to move with the times and use `ss -ln` eh?

warrenvw:
My original plan was to use `ss`, but for some reason, I stuck with `netstat`. I'll write up an issue so we can address this after the current release. I think it's too late in the process to make the change now.

PR #425, opened 2017-02-17
Fix ChatOps install script for EL7

armab:
```sh
sed: can't read st2chatops.env: No such file or directory
# Failed on Configure st2chatops #
```


armab:
So this bug was introduced [1 year ago](https://github.com/StackStorm/st2-packages/commit/0cc573e6a6a36a231707b61d97abdf51b2639217) :smile: 


Seems nobody really tested StackStorm + ChatOps `curl|bash` install since that on `EL7`  :)

lakshmi-kannan:
> Seems nobody really tested StackStorm + ChatOps curl|bash install since that on EL7

Our CI uses curl| bash. It is possible we don't test chatops in EL7 with CI. I know we test in in U14. @humblearner can confirm.

Issue #431, opened 2017-02-27
VERSION matched as regex, not fixed string


LindsayHill:
If a user specifies a version when running `install.sh`, e.g. with `--version=2.0.1`, the version is matched as a regex pattern, not a fixed string. This can mean matching the wrong version = e.g. if 2.0.1 is specified, it will match 2.2.0-1.

Issue is here: 
https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L268

(and similar for el6/el7 scripts).

Looks like fix is to use `grep -F $VERSION`, rather than `grep $VERSION`

PR #433, opened 2017-02-27
changed version matching to fixed string

LindsayHill:
When users specify a version when running `install.sh`, we attempt to find the latest build for that version. We match the input as a pattern, not a fixed string, leading to inconsistent results - e.g. if a user specifies `2.0.1`, we will match `2.2.0-1`. 

This change switches to `grep -F`.

Addresses #431


Kami:
LGTM from high level, but would also be good to do some more testing to make sure it indeed works as expected.

And good catch on the regex match issue.

LindsayHill:
I think this resolves the immediate issue, but there may still be a longer-term issue here if we had versions like 12.0.1 in the repo, and users still wanted to specify 2.0.1

May need to go to something like: `grep ^${VERSION//./\\.}` ?

Kami:
Yeah, I think strict version match like the one you mentioned above would be better (just need to make sure all the cases still work as expected).

LindsayHill:
Here's an alternative that will be a bit tighter match. Should avoid the problem where 12.0.1 and 2.0.1 could both be matched.

LindsayHill:
Hmm. That's not going to work for RPM-based systems.

On DEB systems, we have this output:
```
lhill@st2:~$ apt-cache show st2web | grep Version | awk '{print $2}'
2.2.0-1
2.1.1-1
2.1.0-1
2.0.1-2
2.0.0-1
1.6.0-1
1.5.1-1
1.5.0-1
1.4.0-1
```

So doing something like `grep ^${VERSION//./\\.}` works. For the above output, if VERSION=2.0.1, then  it returns `2.0.1-2`, which is what we want.

But RPM systems report this from repoquery:
```
[vagrant@st2vagrant ~]$ repoquery --nvr --show-duplicates st2web
st2web-1.4.0-1
st2web-1.5.0-1
st2web-1.5.1-1
st2web-1.6.0-1
st2web-2.0.0-1
st2web-2.0.1-2
st2web-2.1.0-1
st2web-2.1.1-1
st2web-2.2.0-1
```

So I'll need a slightly different approach for those.

LindsayHill:
Looks like it might be safe to do something like this for RPM systems:
```
local ST2_VER=$(repoquery --nvr --show-duplicates st2 | grep -F st2-${VERSION} | sort --version-sort | tail -n 1)
```

LindsayHill:
OK, *this* time I think I've got it working. @lakshmi-kannan would you mind casting your eye over this?

LindsayHill:
Looks like we'll also need to cherry-pick these changes into the v2.0 and v2.1 branches. Otherwise users that specify `--version=2.0.1` will still end up pulling the old bootstrap scripts.

PR #434, opened 2017-03-01
Bump BRANCH in 'st2_bootstrap.sh' to 2.2

armab:
As @LindsayHill said, reported in #community.

We'll need to codify that thing in automation to make sure it won't affect next release.


Mierdin:
FYI, I just opened https://github.com/StackStorm/st2cd/pull/260 - perhaps we can get that merged first and try to use that instead of this manual fix? That way we know it works for next release

Issue #429, opened 2017-02-23
st2 fails to run immediately after service restart


armab:
> Experienced during the https://github.com/StackStorm/ops-infra/pull/107 work

We noticed this issue several times before.

If st2 is restarted and we run command immediately after startup, - it fails probably because init reports status too fast or something like that.

This needs investigation and proper fix.

### Steps to reproduce
```sh
sudo st2ctl restart && st2 action list
```

### Error
```
# st2ctl restart && st2 action list
Failed to stop st2chatops.service: Unit st2chatops.service not loaded.
Failed to stop mistral.service: Unit mistral.service not loaded.
Failed to start st2chatops.service: Unit st2chatops.service not found.
Failed to start mistral.service: Unit mistral.service not found.
##### st2 components status #####
st2actionrunner PID: 25339
st2actionrunner PID: 25341
st2actionrunner PID: 25346
st2actionrunner PID: 25349
st2actionrunner PID: 25352
st2actionrunner PID: 25355
st2actionrunner PID: 25359
st2actionrunner PID: 25363
st2actionrunner PID: 25366
st2actionrunner PID: 25369
st2api PID: 25374
st2stream PID: 25377
st2auth PID: 25383
st2garbagecollector PID: 25387
st2notifier PID: 25391
st2resultstracker PID: 25395
st2rulesengine PID: 25399
st2sensorcontainer PID: 25403
st2chatops is not running.
mistral-server is not running.
mistral-api is not running.

ERROR: HTTPConnectionPool(host='127.0.0.1', port=9101): Max retries exceeded with url: /v1/actions (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x7f1d2dddd250>: Failed to establish a new connection: [Errno 111] Connection refused',))

```

after a couple of seconds, running `st2 action list` works OK.

### TODO
- [x] Check if `CentOS7` with systemd is affected
- [x] Check if `Ubuntu Trusty` with upstart is affected
- [x] Check if `CentOS6` with SysvInit is affected
- [x] Investigate the Fix https://github.com/StackStorm/st2-packages/pull/435#issuecomment-284006738
- [x] Remove hack in [`curl | bash`](https://github.com/StackStorm/st2-packages/blob/788761323d3402c29efa87c2678de684a77eb6f6/scripts/st2bootstrap-deb.sh#L554-L557)
[...]

PR #435, opened 2017-03-01
Fix st2 services availability after restart

armab:
> Work on bug described in [#429 st2 fails to run immediately after service restart](https://github.com/StackStorm/st2-packages/issues/429)

### TODO
See https://github.com/StackStorm/st2-packages/issues/429


armab:
The problem is that when service starts it takes some time before app is initialized, connected to DBs and ports are opened to serve the connection.

After researching there are several solutions to this problem:

## 1) Service sends a notification to systemd when ready
This is 100% right & great solution in terms of long-term reliability. This is how many "well-established" tools work like `ssh` and others.

When service like `st2api`, `st2auth`, `st2stream`, etc is ready and can serve requests it sends a notification with [`sd_notify()`](https://www.freedesktop.org/software/systemd/python-systemd/daemon.html#systemd.daemon.notify) protocol with `READY=1` message to systemd.
Apart of that, it's possible to 

More in [Example 6. Services that notify systemd about their initialization](https://www.freedesktop.org/software/systemd/man/systemd.service.html#id-1.9.7)

- Pros:
  - could be extended to advanced cases like [watchdog](http://0pointer.de/blog/projects/watchdog.html) (service reliability +++)
- Cons:
  - Requires `st2` python code change

## 2) Systemd Socket Activation
Socket activation works for daemons which want to listen on a network port. Systemd would open the socket, listen on it, and then spawn the daemon when a connection comes in. With that, the requests are queued before actual for example `st2api` is still in initialization mode.

More: http://0pointer.de/blog/projects/socket-activation.html

## 3) Upstart Socket Bridge
Same as `systemd` socket activation, but for `upstart`.
http://upstart.ubuntu.com/cookbook/#upstart-socket-bridge

## 4) Hack it with bash
Block `service start` until the port is opened. Example:
```
timeout 15 bash -c 'until echo > /dev/tcp/127.0.0.1/9101; do sleep 0.5; done'
```
For example this is common practice for `upstart` which doesn't have so much flexibility like `systemd`.


#### Resources
- http://stackoverflow.com/questions/13069634/python-daemon-and-systemd-service
- http://stackoverflow.com/questions/36893888/systemd-daemon-in-python-with-watchdog-support/37111518#37111518
- http://stackoverflow.com/questions/32947606/calling-a-sd-notify0-watchdog-1-in-a-service
- https://www.freedesktop.org/software/systemd/man/systemd.service.html
- http://0pointer.de/blog/projects/watchdog.html
- http://upstart.ubuntu.com/cookbook/#upstart-socket-bridge
- http://upstart.ubuntu.com/cookbook/#daemon-behaviour
- https://www.freedesktop.org/software/systemd/man/systemd.service.html#id-1.9.6
- http://unix.stackexchange.com/questions/5277/how-do-i-tell-a-script-to-wait-for-a-process-to-start-accepting-requests-on-a-po
- http://0pointer.de/blog/projects/socket-activation.html
- http://0pointer.de/blog/projects/inetd.html
- https://gist.github.com/drmalex07/28de61c95b8ba7e5017c
- https://gist.github.com/kylemanna/d193aaa6b33a89f649524ad27ce47c4b
- https://www.freedesktop.org/software/systemd/man/systemd-socket-proxyd.html
- https://www.lucas-nussbaum.net/blog/?p=877
- https://www.freedesktop.org/software/systemd/python-systemd/daemon.html#systemd.daemon.notify

armab:
Systemd socket activation is something that in fact was initially recommended by `gunicorn`:
http://docs.gunicorn.org/en/stable/deploy.html#systemd

bigmstone:
Interesting set of changes. I, like you, favor option 1. Will keep an eye on the deltas here.

armab:
Need to finally remove `wheezy` and `jessie` #365, - it only takes more time to bring additional fixes for them.

packages/st2/debian/st2.st2api.init
```diff
@@ -71,6 +71,8 @@ do_start()
   start-stop-daemon --start --quiet --chuid $RUNAS_USER:$RUNAS_GROUP --umask $UMASK --background --pidfile $PIDFILE \
      --exec $DAEMON -- $DAEMON_ARGS \
     || return 2
+  # wait for st2api readiness to serve the requests
+  timeout 5 bash -c 'until echo > /dev/tcp/127.0.0.1/9101; do sleep 0.5; done' || return 2
```
bigmstone: :+1:

armab: I'm just not sure about `5s` timeout, looks a bit high value.
Probably retries during the 3s timeout should be more than enough to understand if there is something wrong with the service and it's broken.

Kami:
Nice work :+1:

packages/st2/debian/st2.st2auth.init
```diff
@@ -71,6 +71,8 @@ do_start()
   start-stop-daemon --start --quiet --chuid $RUNAS_USER:$RUNAS_GROUP --umask $UMASK --background --pidfile $PIDFILE \
      --exec $DAEMON -- $DAEMON_ARGS \
     || return 2
+  # wait for st2auth readiness to serve the requests
+  timeout 5 bash -c 'until echo > /dev/tcp/127.0.0.1/9100; do sleep 0.5; done' || return 2
```
Kami: I'm having some issues with st2 e2e tests failing, and it looks like to be related to this change (st2auth service doesn't start at all causing tests to fail).

Will dig into it and try to increase the timeout, 5 seconds might simply be too little to give service enough time to connect to MongoDB and RabbitMQ.

Kami: Looks like it's actually an issue with systemd (Ubuntu 16.04).

Auth service never starts and looks like systemd listens on 9100 and never releases that port and that's why service fails to start:

```bash
ubuntu@st2-pkg-unstable-u16-ent-dev-9e465edbf5:~$ sudo lsof -i :9100
COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
systemd   1 root   47u  IPv4 181597      0t0  TCP localhost:9100 (LISTEN)
```

armab: From the https://stackstorm.slack.com/archives/C029K4CBU/p1489841065872703 it looks like `st2-auth-backend-etenrprise` what broke the things.

Looks like resolved.

Kami: Yeah it was a versioning issue (build was incorrectly using ldap backend from master instead of v2.2 branch) - chicken and egg problem with st2 and st2-packages.

After I merged st2-packages PR everything turned green again.

PR #438, opened 2017-03-02
Include st2tests Python component in final st2 package distribution

Kami:
This will allow users to run `st2-run-pack-tests` command out of the box without needing to set anything up on servers where StackStorm is installed using packages.

Related to https://github.com/StackStorm/st2/pull/3263


Kami:
Build passed, created artifacts on Circle CI look good.

I will go ahead, merge it, wait for new packages to be build and test it on a fresh server.

PR #439, opened 2017-03-02
changed version matching to fixed string (v2.0)

LindsayHill:
Same changes as #433 back ported to v2.0 branch. Needs to be back-ported because user having problems with install is trying to v2.0.

Addresses #431

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -115,28 +115,28 @@ EOT"
 get_full_pkg_versions() {
   if [ "$VERSION" != '' ];
   then
-    local ST2_VER=$(apt-cache show st2 | grep Version | awk '{print $2}' | grep -F $VERSION | sort --version-sort | tail -n 1)
+    local ST2_VER=$(apt-cache show st2 | grep Version | awk '{print $2}' | grep ^${VERSION//./\\.} | sort --version-sort | tail -n 1)
     if [ -z "$ST2_VER" ]; then
       echo "Could not find requested version of StackStorm!!!"
       sudo apt-cache policy st2
       exit 3
     fi
 
-    local ST2MISTRAL_VER=$(apt-cache show st2mistral | grep Version | awk '{print $2}' | grep -F $VERSION | sort --version-sort | tail -n 1)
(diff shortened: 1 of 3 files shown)


LindsayHill:
Thanks for fixing up the PR title, somehow completely missed that

Kami:
Ugh - looks like this build is breaking because it's trying to use some other st2-packages changes from master (01751dfce725a019dbe365f33e4ab1ca58ca71c5).

Kami:
Also, v2.0?

We don't really (actively) support v2.0 anymore. Or was this supposed to be v2.2?

In any case, it looks like there is some weirdness going on with st2-packages versioning which we need to investigate.

And it also looks like we are missing branches for v2.1 and v2.2 release in this repo...

armab:
The branches are in place in this repo: 
https://github.com/StackStorm/st2-packages/tree/v2.2
https://github.com/StackStorm/st2-packages/tree/v2.1

¯\_(ツ)_/¯

LindsayHill:
The reason it is also needed in the 2.0 branch is because users (andy111) need to install v2.0.1, and are using the install script to do so. If you specify a version when running install.sh, it will use that branch's version of the bootstrap script. 2.0.1 clashes with 2.2.0-1 with the old pattern matching.

LindsayHill:
Note that change has already been made in master. Not bothering with 2.2 at this stage.

Kami:
@armab Sorry, my mistake - I missed those branch.

Kami:
I'm still not sure why the tests are failing through (they shouldn't since st2tests change is only in master, it looks like Circle CI might be using master or testing merge in master or similar), but I haven't had the time to dig in yet.

armab:
Don't think it's related to `st2tests`.

From what I see from the build logs, different python s~e~ix errors like:
```
IOError: [Errno 2] No such file or directory: '/tmp/st2-20170303-8-ht6mtk/st2/debian/st2/opt/stackstorm/st2/lib/python2.7/site-packages/six-1.10.0.dist-info/METADATA'
```

I think we're good to merge, since `curl|bash` bootstrap scripts changes doesn't affect packaging in any way.

LindsayHill:
Yeah, I couldn't figure out the build failures either. As you say, the script change doesn't affect packaging anyway

PR #440, opened 2017-03-04
modified homedir refs

LindsayHill:
Modified references to user's home directories to use ~stanley, rather than /home/stanley, and ~root rather than /root. 

Some environments use a different directory for the base $HOME dir - e.g. /users. By changing to ~user, we can cope with those differences.

Related to that, some environments have changed the default so that new users do not get created with a machine group. Instead they share a common group, e.g. `users`. By changing the `chown -R stanley:stanley` line to `chown -R stanley` we can cope with those differences. Note that the permissions for `~/.ssh/authorized_keys` are still acceptable - OpenSSH only checks the user and mode, it doesn't seem to care about the group.

This will make our one-line installer slightly more robust, and able to cope with different environments. A few more PRs to come :)

Also fixed up a couple of small typos in comments.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -339,17 +339,17 @@ configure_st2_user () {
     sudo useradd stanley
   fi
 
-  sudo mkdir -p /home/stanley/.ssh
+  sudo mkdir -p ~stanley/.ssh
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
-  #sudo cp ${KEY_LOCATION}/stanley_rsa.pub /home/stanley/.ssh/stanley_rsa.pub
+  sudo ssh-keygen -f ~stanley/.ssh/stanley_rsa -P ""
+  #sudo cp ${KEY_LOCATION}/stanley_rsa.pub ~stanley/.ssh/stanley_rsa.pub
 
-  # Authorize key-base acces
-  sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'
-  sudo chmod 0600 /home/stanley/.ssh/authorized_keys
-  sudo chmod 0700 /home/stanley/.ssh
-  sudo chown -R stanley:stanley /home/stanley
+  # Authorize key-based access
+  sudo sh -c 'cat ~stanley/.ssh/stanley_rsa.pub >> ~stanley/.ssh/authorized_keys'
+  sudo chmod 0600 ~stanley/.ssh/authorized_keys
+  sudo chmod 0700 ~stanley/.ssh
+  sudo chown -R stanley ~stanley
 
   # Enable passwordless sudo
   sudo sh -c 'echo "stanley    ALL=(ALL)       NOPASSWD: SETENV: ALL" >> /etc/sudoers.d/st2'
@@ -384,7 +384,7 @@ configure_st2_cli_config() {
 
   : "${HOME:=`eval echo ~$(whoami)`}"
 
-  ROOT_USER_CLI_CONFIG_DIRECTORY="/root/.st2"
+  ROOT_USER_CLI_CONFIG_DIRECTORY="~root/.st2"
   ROOT_USER_CLI_CONFIG_PATH="${ROOT_USER_CLI_CONFIG_DIRECTORY}/config"
 
   CURRENT_USER_CLI_CONFIG_DIRECTORY="${HOME}/.st2"
@@ -516,7 +516,7 @@ install_st2chatops() {
 }
 
 configure_st2chatops() {
-  # set API keys. This should work since CLI is configuered already.
+  # set API keys. This should work since CLI is configured already.
   ST2_API_KEY=`st2 apikey create -k`
   sudo sed -i -r "s/^(export ST2_API_KEY.).*/\1$ST2_API_KEY/" /opt/stackstorm/chatops/st2chatops.env
 
--- a/scripts/st2bootstrap-el6.sh
+++ b/scripts/st2bootstrap-el6.sh
@@ -367,16 +367,16 @@ configure_st2_user() {
     sudo useradd stanley
   fi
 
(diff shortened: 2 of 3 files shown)


scripts/st2bootstrap-deb.sh
```diff
@@ -339,17 +339,17 @@ configure_st2_user () {
     sudo useradd stanley
   fi
 
-  sudo mkdir -p /home/stanley/.ssh
+  sudo mkdir -p ~stanley/.ssh
```
armab: Change like that will require `st2.conf` adjustments as well: https://github.com/StackStorm/st2/blob/master/conf/st2.package.conf#L73
And not sure if that's a good idea. Or we stick with `~` everywhere or with old `/home/stanley/` as it was before.

Kami: Yeah, I agree.

In theory I'm not against this change, but it's a bit more involved and we need to make sure it doesn't break anything.

Also, IIRC, `~` and `$HOME` usually doesn't contain trailing slash by default so if anything it should be `~/stanley` to be on the safe side.

LindsayHill: Ah, good point about st2.package.conf. Not much point fixing it in the bootstrap script if it promptly breaks after install. I'm actually not sure how `~` would be interpreted by st2 when it finds it in the config file. I know how the shell interprets it, but it could be different if it's the st2actionrunner process.

@Kami That is explicitly meant to be `~stanley`, so it translates as "stanley's home directory". The other one there `~/stanley` would mean "a file/folder named 'stanley' in the home directory of whichever users is running this command". What I was trying to do in this change is to account for those environments where `/home` is not the default base path for users. Some people instead use things like `/users`. That make sense?

LindsayHill: OK, did some quick tests, and this is indeed a problem. 

First I changed `/etc/st2/st2.conf` to include this:
```
[system_user]
user = stanley
ssh_key_file = ~stanley/.ssh/stanley_rsa
```
I then ran `st2 run core.remote hosts=127.0.0.1 cmd='id'`, and got this output:
```
id: 58b834b5c4da5f3126df0aaa
status: failed
parameters:
  cmd: id
  hosts: 127.0.0.1
result:
  error: "Unable to connect to any one of the hosts: [u'127.0.0.1'].

 connect_errors={
  "127.0.0.1": {
    "failed": true,
    "traceback": "Traceback (most recent call last):\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/parallel_ssh.py\", line 243, in _connect\n    client.connect()\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/paramiko_ssh.py\", line 138, in connect\n    self.client = self._connect(host=self.hostname, socket=self.bastion_socket)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/paramiko_ssh.py\", line 592, in _connect\n    passphrase_reqd = self._is_key_file_needs_passphrase(self.key_files)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/paramiko_ssh.py\", line 672, in _is_key_file_needs_passphrase\n    cls.from_private_key_file(file, password=None)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/paramiko/pkey.py\", line 196, in from_private_key_file\n    key = cls(filename=filename, password=password)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/paramiko/rsakey.py\", line 46, in __init__\n    self._from_private_key_file(filename, password)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/paramiko/rsakey.py\", line 173, in _from_private_key_file\n    data = self._read_private_key_file('RSA', filename, password)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/paramiko/pkey.py\", line 267, in _read_private_key_file\n    with open(filename, 'r') as f:\nIOError: [Errno 2] No such file or directory: '~stanley/.ssh/stanley_rsa'\n",
    "timeout": false,
    "succeeded": false,
    "stdout": "",
    "stderr": "",
    "error": "Failed connecting to host 127.0.0.1. ",
    "return_code": 255
  }
}"
  traceback: "  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2actions/container/base.py", line 90, in _do_run
    runner.pre_run()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/paramiko_ssh_runner.py", line 145, in pre_run
    self._parallel_ssh_client = ParallelSSHClient(**client_kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/parallel_ssh.py", line 61, in __init__
    connect_results = self.connect(raise_on_any_error=raise_on_any_error)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/parallel_ssh.py", line 91, in connect
    raise NoHostsConnectedToException(msg)
"
```

Running `ls -l ~stanley/.ssh/stanley_rsa` succeeds:
```
[root@st2vagrant ~]# ls -l ~stanley/.ssh/stanley_rsa
-rw-------. 1 stanley stanley 1679 Mar  1 17:44 /home/stanley/.ssh/stanley_rsa
[root@st2vagrant ~]# 
```

But of course this is because the shell is interpreting `~`, but ST2 can't.

I'm not sure what the best way to proceed here is. Couple of options I can think of:

1. Go back to using `/home/stanley` regardless of where stanley's $HOME dir is. Should still work to create that directory & set permissions. *BUT* make the authorized_keys line look like this:
`'cat /home/stanley/.ssh/stanley_rsa.pub >> ~stanley/.ssh/authorized_keys'` - the key there is that the private key can live pretty much anywhere as long as it has the right permissions. But when we put the pub key part in authorized_keys, we *must* put it in `~stanley/.ssh/authorized_keys`, not just `/home/stanley/.ssh/authorized_keys`.

2. Automatically determine the location of `~stanley`, translate it to the actual directory used, and then use `crudini` to set that option in `/etc/st2/st2.conf`.

Right now I'm leaning more towards option 1. Thoughts?

LindsayHill:
Superseded by other changes

PR #442, opened 2017-03-04
Make installer better handle multiple runs

LindsayHill:
If the bootstrap script is re-run, e.g. due to a failed installation, it generally exits at one of two points:
- When generating the datastore encryption key
- When generating stanley's SSH key.

This PR adds two changes:
- Add `--force` to crypto key generate command, so it will over-write existing key. Yes, this could cause problems if you had values encrypted with the old key
- Adds checks to see if stanley's RSA key exists, and to bypass generating/copying a new key if so.

These fixes are not about true idempotence - it's just about about making the script more robust, and better handle situations where it gets re-run due to a failed install.

Note this builds upon https://github.com/StackStorm/st2-packages/pull/440

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -339,17 +339,17 @@ configure_st2_user () {
     sudo useradd stanley
   fi
 
-  sudo mkdir -p /home/stanley/.ssh
+  sudo mkdir -p ~stanley/.ssh
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
-  #sudo cp ${KEY_LOCATION}/stanley_rsa.pub /home/stanley/.ssh/stanley_rsa.pub
+  sudo ssh-keygen -f ~stanley/.ssh/stanley_rsa -P ""
+  #sudo cp ${KEY_LOCATION}/stanley_rsa.pub ~stanley/.ssh/stanley_rsa.pub
 
-  # Authorize key-base acces
-  sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'
-  sudo chmod 0600 /home/stanley/.ssh/authorized_keys
-  sudo chmod 0700 /home/stanley/.ssh
-  sudo chown -R stanley:stanley /home/stanley
+  # Authorize key-based access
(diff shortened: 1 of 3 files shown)


LindsayHill:
@bigmstone yeah I've wondered about that. Roughly, behaviour could be something like:

1/ Default is to ask the user, if either datastore encryption key or stanley key exists
2/ Allow user to over-ride behaviour with `--force` or some similar CLI option.

There's a semi-related issue with BWC installer, where it attempts to run ST2 installer regardless of whether ST2 already installed or not. Ideally it should just add the BWC bits if there's already a working ST2 install.

Kami:
Thanks for working on this, but I'm personally not sure if we should go forward with this change.

The main reason for switching from puppet based installer to a very simple script based installer was simplicity. People could re-run the old puppet based installer to rectify partial / failed installation, but this had many, many issues - the whole approach was very complicated and it was very hard to troubleshoot and debug things.

I personally probably spent more than 1 month total total debugging and fixing all the bugs (and collectively we have spent a lot more time) on that installer. In addition to that the whole experience was also slow and very frustrating for our users.

So yes, there are many ways we could improve the existing simple installer based script (performance, etc.), but I'm strongly against it if changing it means sacrificing simplicity and introducing complexity (it's hard to pull the line and if we will be keep adding "functionality" to it, we will eventually come up with (poorly) implemented puppet / ansible / cfg management system).

Again, I appreciate your work, but we made some deliberate decision with the installer script and I would like to keep those.

It's also worth noting that making the installer script re-entrant (better handle multiple re-runs) sounds good in theory, but in practice it's usually a lot faster (and saves everyone tons of time troubleshooting things, etc.) to simply say to the user to spin up a new fresh box and run installer script there again if it fails compared to wasting time debugging tons of issues which can come up during re-runs.

bigmstone:
I don't disagree with your last statement @kami, but for a majority of enterprise users that's not a reality. It can (literally) take weeks to get a new VM and backing out of a half-installed ST2 is a PITA. Perhaps they should be using ansible provisioner to accomplish provisioning in this case, but a lot don't have config management (or use one we don't support). So while enhancing the bash script might not be the right place to fix all of this we need to attempt to fix this somewhere. Perhaps the right place is in a meta-deb/rpm package. Perhaps not. It's just an actual problem and if it's a user's first window into ST2 then we should make an effort to have it be as seamless as possible. Especially for the enterprise customers.

LindsayHill:
Superseded by other changes

Issue #445, opened 2017-03-14
'st2ctl register' fails to register rules on a fresh DB


armab:
Previous PR https://github.com/StackStorm/st2-packages/issues/429 removed all `sleep` occurencies from the installer and fixed some deep issues with the st2 services startup, handling better `st2api`, `st2auth`, `st2stream` to report correct status only when respective service is listening on the network port.

But not all bugs were caught. Under Ubuntu Xenial after initial `st2` install running
```
sudo st2ctl start && st2ctl reload --register-all
```
fails to register the rules.

### Reproduce
Run xenial installer (with removed `sleep`): 
```sh
curl -sSL https://raw.githubusercontent.com/StackStorm/st2-packages/27ec5a09d87c3bc2bcfd0e01adf3ff4daa294551/scripts/st2bootstrap-deb.sh | bash -s -- --user=st2admin --password=st2admin --unstable --staging
```

Happens only during the first boot.

### Error

```sh
Registering content...[flags = --config-file /etc/st2/st2.conf --register-all]
2017-03-14 17:58:48,626 INFO [-] Connecting to database "st2" @ "0.0.0.0:27017" as user "stackstorm".
2017-03-14 17:58:50,440 INFO [-] =========================================================
2017-03-14 17:58:50,440 INFO [-] ############## Registering triggers #####################
2017-03-14 17:58:50,440 INFO [-] =========================================================
2017-03-14 17:58:50,550 INFO [-] Registered 0 triggers.
2017-03-14 17:58:50,550 INFO [-] =========================================================
2017-03-14 17:58:50,550 INFO [-] ############## Registering sensors ######################
2017-03-14 17:58:50,551 INFO [-] =========================================================
2017-03-14 17:58:50,618 INFO [-] Registered 1 sensors.
2017-03-14 17:58:50,619 INFO [-] =========================================================
2017-03-14 17:58:50,619 INFO [-] ############## Registering runners ######################
2017-03-14 17:58:50,619 INFO [-] =========================================================
2017-03-14 17:58:51,204 INFO [-] Registered 13 runners.
2017-03-14 17:58:51,204 INFO [-] =========================================================
2017-03-14 17:58:51,204 INFO [-] ############## Registering actions ######################
2017-03-14 17:58:51,204 INFO [-] =========================================================
2017-03-14 17:58:53,184 INFO [-] Registered 46 actions.
2017-03-14 17:58:53,185 INFO [-] =========================================================
[...]

PR #446, opened 2017-03-14
Temporary hack for Xenial failing to register rules

armab:
Regression I didn't catch after removing this line: https://github.com/StackStorm/st2-packages/pull/435/files#diff-0cc9e53cba1b060958ebffa16eb85cb5L332

Merging it now to make the current build GREEN, revert when https://github.com/StackStorm/st2-packages/issues/445 is fully investigated and fixed.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -329,6 +329,10 @@ install_st2() {
   sudo crudini --set /etc/st2/st2.conf database password "${ST2_MONGODB_PASSWORD}"
 
   sudo st2ctl start
+  # TODO: Fix https://github.com/StackStorm/st2-packages/issues/445
+  if [[ "$SUBTYPE" == 'xenial' ]]; then
+    sleep 5
+  fi
   sudo st2ctl reload --register-all
 }
 
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -329,7 +329,7 @@ install_st2() {
   sudo crudini --set /etc/st2/st2.conf database password "${ST2_MONGODB_PASSWORD}"
 
   sudo st2ctl start
-  # TODO: Fix https://github.com/StackStorm/st2-packages/issues/445
+  # TODO: Fix https://github.com/StackStorm/st2-packages/issues/445 (under xenial register content fails on first boot)
   if [[ "$SUBTYPE" == 'xenial' ]]; then
     sleep 5
   fi


scripts/st2bootstrap-deb.sh
```diff
@@ -329,6 +329,10 @@ install_st2() {
   sudo crudini --set /etc/st2/st2.conf database password "${ST2_MONGODB_PASSWORD}"
 
   sudo st2ctl start
+  # TODO: Fix https://github.com/StackStorm/st2-packages/issues/445
```
m4dcoder: Can you provide some explanation here why sleep is required for xenial?

m4dcoder:
Otherwise this LGTM. Just want to understand why a sleep is required. The referenced issue didn't exactly spell it out clearly.

scripts/st2bootstrap-deb.sh
armab: Added a few more words of explanation.
But I think that dedicated issue assigned to that bug is enough.

scripts/st2bootstrap-deb.sh
```diff
@@ -329,6 +329,10 @@ install_st2() {
   sudo crudini --set /etc/st2/st2.conf database password "${ST2_MONGODB_PASSWORD}"
 
   sudo st2ctl start
+  # TODO: Fix https://github.com/StackStorm/st2-packages/issues/445 (under xenial register content fails on first boot)
```
m4dcoder: Thanks for updating. But why does the sleep here make this work?

armab:
To be clear, the `sleep 5` was there for 1 year or so.

I've enhanced the services availability & startup as part of previously merged PR: https://github.com/StackStorm/st2-packages/pull/435 and removed all sleeps from the installer, but didn't catch this specific issue.

Probably something inside the st2 core and how services are started (some resources are not available after immediate start). IDK yet about the reasons.
So for now I'm reverting only this sleep how it was before to unblock the build and investigate later the proper fix: https://github.com/StackStorm/st2-packages/issues/445

m4dcoder:
Thanks for the explanation.

scripts/st2bootstrap-deb.sh
Kami: Probably takes a file for services to start up, we had a sleep here (or in a similar place) before for all the distros, iirc.

armab: Could debug that thing deeper here: https://github.com/StackStorm/st2-packages/issues/445#issuecomment-286778253 with bash script to reproduce it every time.

It's something how st2 services interact with mongo. This happens on a clean MongoDB during the first st2 register rules run.

@Kami I hope to get your help there, since after further digging it looks not like a Mongo startup or packaging/services problem, but more like an issue to fix in StackStorm core.

PR #449, opened 2017-03-21
Adding webhook entry for ci-webhooks

### LINKED ISSUE #448: New Webhook Entry for ci-webhooks
Need to add a webhook entry for the new `ci-webhooks` endpoint.

bigmstone:
Fixes #448

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,7 +13,6 @@ general:
 notify:
   webhooks:
     - url: https://webhooks.stackstorm.net:8531/webhooks/build/events
-    - url: https://ci-webhooks.stackstorm.net/api/v1/webhooks/circleci?st2-api-key=${ST2_API_KEY}
 
 machine:
   environment:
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -75,6 +75,8 @@ test:
     - .circle/docker.sh build st2actionrunner st2api st2stream st2auth st2exporter st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2garbagecollector
     - .circle/docker.sh run st2api
     - .circle/docker.sh test st2api 'st2 --version'
+    - "curl -H \"Content-Type: application/json\" -X POST -d '{\"build_num\":\"'\"${CIRCLE_BUILD_NUM}\"'\"}' https://ci-webhooks.stackstorm.net/api/v1/webhooks/circleci?st2-api-key=${ST2_API_KEY}"
+    # Don't even ask about ^ - I wouldn't do it if I didn't have to
```
armab: I get the point. You need it in `deployment` section then, since it will run only `1 time` and only when all parallel builds suceeded.

See similar `circle/save_payload.py ~/packages` task, which I think is outdated but shows the point.

PR #450, opened 2017-03-21
Modified git URL to include proper mistral version [master]

Mierdin:
During e2e testing for 2.2.1, the wrong version of st2mistral was being packaged. This should resolve this.

Note that this is merging to `master`. #452 will merge to v2.2 release branch

### DIFF
--- a/packages/st2mistral/Makefile
+++ b/packages/st2mistral/Makefile
@@ -8,7 +8,7 @@ PBR_GITURL := git+https://github.com/openstack-dev/pbr@d19459daa8616dd18fde016c2
 
 # We use special additional requirements in our mistral bundle!
 define INJECT_DEPS
-git+https://github.com/StackStorm/st2mistral.git@$(GITREV)#egg=st2mistral
+git+https://github.com/StackStorm/st2mistral.git@st2-$(MISTRAL_VERSION)#egg=st2mistral
 $(PBR_GITURL)
 endef
 export INJECT_DEPS


armab:
Yeah, the PR looks pretty strange, because it arises more questions like "How it worked before then"?
I guess we forgot to change something else.

Before that, @Mierdin can you provide more info?
> During e2e testing for 2.2.1, the wrong version of st2mistral was being packaged.

Do you have any info in `v2.2.1` which version/source code of `st2mistral` was packaged?

packages/st2mistral/Makefile
```diff
@@ -8,7 +8,7 @@ PBR_GITURL := git+https://github.com/openstack-dev/pbr@d19459daa8616dd18fde016c2
 
 # We use special additional requirements in our mistral bundle!
 define INJECT_DEPS
-git+https://github.com/StackStorm/st2mistral.git@$(GITREV)#egg=st2mistral
+git+https://github.com/StackStorm/st2mistral.git@st2-$(MISTRAL_VERSION)#egg=st2mistral
```
armab: So here is where `gitrev`: https://github.com/StackStorm/st2-packages/blob/master/rake/build/environment.rb#L85 env is populated for `st2mistral` package.

It already looks good.
So if we need to change the stuff, - change that line.

Let me look at the [`v2.2` change](https://github.com/StackStorm/st2-packages/pull/452)

armab:
From https://stackstorm.slack.com/archives/C03CVRC1A/p1490207760981926 it seems to be resolved.

PR #452, opened 2017-03-21
Modified git URL to include proper mistral version [v2.2]

Mierdin:
During e2e testing for 2.2.1, the wrong version of st2mistral was being packaged. This should resolve this.

Note that this is for v2.2 release branch. #450 will take care of this change in `master`

### DIFF
--- a/packages/st2mistral/Makefile
+++ b/packages/st2mistral/Makefile
@@ -8,7 +8,7 @@ PBR_GITURL := git+https://github.com/openstack-dev/pbr@d19459daa8616dd18fde016c2
 
 # We use special additional requirements in our mistral bundle!
 define INJECT_DEPS
-git+https://github.com/StackStorm/st2mistral.git@$(GITREV)#egg=st2mistral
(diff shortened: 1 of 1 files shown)


packages/st2mistral/Makefile
```diff
@@ -8,7 +8,7 @@ PBR_GITURL := git+https://github.com/openstack-dev/pbr@d19459daa8616dd18fde016c2
 
 # We use special additional requirements in our mistral bundle!
 define INJECT_DEPS
-git+https://github.com/StackStorm/st2mistral.git@$(GITREV)#egg=st2mistral
+git+https://github.com/StackStorm/st2mistral.git@st2-$(MISTRAL_VERSION)#egg=st2mistral
```
armab: So yeah, as @m4dcoder noted, looks it doesn't fix anything.
`GITREV` value should be `st2-2.2.1` from the https://github.com/StackStorm/st2-packages/blob/v2.2/rake/build/environment.rb#L85

Eg. correct value is already provided.

armab:
From the https://stackstorm.slack.com/archives/C03CVRC1A/p1490207760981926 it seems to be resolved.

PR #458, opened 2017-04-25
Share private /tmp folder between actionrunner instances

enykeev:
It's a frequent situation when you putting a file to `/tmp` folder in one action and retrieving it from what you think is the same `/tmp` folder in the following action. Problem is, the expectation is invalid for systemd-based OS with PrivateTmp enabled.

This PR tells different instances of st2actionrunner to share the same network and filesystem namespace.

### DIFF
--- a/packages/st2/debian/st2actionrunner.service
+++ b/packages/st2/debian/st2actionrunner.service
@@ -7,6 +7,7 @@ Type=oneshot
 EnvironmentFile=-/etc/default/st2actionrunner
 ExecStart=/bin/bash /opt/stackstorm/st2/bin/runners.sh start
 ExecStop=/bin/bash /opt/stackstorm/st2/bin/runners.sh stop
+PrivateTmp=true
 RemainAfterExit=true
 
 [Install]
--- a/packages/st2/debian/st2actionrunner@.service
+++ b/packages/st2/debian/st2actionrunner@.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=StackStorm service st2actionrunner
 After=network.target
+JoinsNamespaceOf=st2actionrunner.service
 
 [Service]
 Type=simple
--- a/packages/st2/rpm/st2actionrunner.service
+++ b/packages/st2/rpm/st2actionrunner.service
@@ -7,6 +7,7 @@ Type=oneshot
 EnvironmentFile=-/etc/sysconfig/st2actionrunner
 ExecStart=/bin/bash /opt/stackstorm/st2/bin/runners.sh start
 ExecStop=/bin/bash /opt/stackstorm/st2/bin/runners.sh stop
+PrivateTmp=true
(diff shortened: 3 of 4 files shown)


Kami:
Good catch :+1:

Kami:
> Takes a boolean argument. If true sets up a new file system
    namespace for the executed processes and mounts a private /tmp
    directory inside it, that is not shared by processes outside of
    the namespace. This is useful to secure access to temporary files
    of the process, but makes sharing between processes via /tmp
    impossible. Defaults to false.

(just thinking out loud and trying to understand it)

So an alternative would be to disable it aka set it to false, right?

But we set it to true and have other processes join the same namespaces so they can access it when its set to true, correct (aka more secure then just disabling it)?

enykeev:
Yep, only st2actionrunner processes would be able to access it this way. Less chance of a collision with other processes.

Kami:
Thanks for clarifying, I like the approach :)

armab:
Awesome!
+:100: @enykeev for finding the option to configure it. 

Also it, closes https://github.com/StackStorm/st2/issues/3122 issue previously opened by @AndyMoore111

Kami:
I just spent 40 minutes debugging a problem which I believe is related to this issue.

On Ubuntu 16.04, StackStorm v2.5.0.

File is visible in `/tmp/` if I manually run `ls -la /tmp` as `root` and `stanley` user, but if I run that command as part of local runner action `/tmp` is empty.

The actual file is written inside `/tmp` by st2 action runner action.

Looking at the code changes, shouldn't we do `PrivateTmp=false` or am I missing something?

Kami:
To sum it up - I see the value with having a private copy of `/tmp` just for action runners and sharing it across, but that doesn't seem to work in workflows I tried - `/tmp/` still shows as empty.

If that's the case, I propose changing that option to `PrivateTmp=false`.

I also didn't encounter this problem locally and on another server, because I was first testing it with Ubuntu 14.04.

As a temporary work around for our workflows, I can use a non /tmp/ directory, but Imo we should also fix this since it's a pretty big issue for a lot of actions and workflows which write actions inside /tmp/.

Kami:
Another thing - it might be related to the action - we scp file over to `/tmp` and that file seems to not be visible to action runner even though scp is ran under action runner, it some how doesn't inherit the /tmp directory.

In any case, even if that is the expected behavior, I think it's a bad one and if we don't think we should change it to `PrivateTmp=false` we should at least very clearly and explicitly document it because writing to `/tmp` is a very common pattern and users shouldn't spend 1 hour debugging such behavior.

enykeev:
@Kami are you claiming that `JoinsNamespaceOf` does not work on your system for some reason?

Kami:
I believe it's because of the action in question - it won't work for core.remote and other actions which are not directly spawned by action runner (e.g. python runner ones).

PR #459, opened 2017-05-04
install locales-all to avoid crash #3392

arthurlogilab:


armab:
We've just discussed this internally with StackStorm team and agreed that it's user's responsibility to configure the locale properly.
Installing _ALL_ possible pre-generated locales (ex filelist: https://packages.debian.org/wheezy/amd64/locales-all/filelist) via `locales-all` is not what most of the users want.

So I'm closing this, as well as https://github.com/StackStorm/st2/issues/3392. Additionally, it's not StackStorm issue, but user configuration + mongo.

Anyways, thanks for the report and PR! It provoked some good discussions, your contribution counts! :+1:

Issue #424, opened 2017-02-17
[EL6] st2 services are not started after reboot


armab:
### Before reboot
```sh
[vagrant@centos6 ~]$ st2ctl status
##### st2 components status #####
st2actionrunner PID: 9196
st2actionrunner PID: 9213
st2actionrunner PID: 9234
st2actionrunner PID: 9255
st2actionrunner PID: 9285
st2actionrunner PID: 9313
st2actionrunner PID: 9340
st2actionrunner PID: 9365
st2actionrunner PID: 9400
st2actionrunner PID: 9423
st2api PID: 11749
st2api PID: 11761
st2stream PID: 10154
st2stream PID: 10165
st2auth PID: 9562
st2auth PID: 9573
st2garbagecollector PID: 9608
st2notifier PID: 9640
st2resultstracker PID: 9670
st2rulesengine PID: 9706
st2sensorcontainer PID: 9734
st2chatops PID: 11270
st2chatops PID: 11271
mistral-server PID: 10941
mistral-api PID: 10891
mistral-api PID: 10900
mistral-api PID: 10901
```
### After reboot
```sh
[vagrant@centos6 ~]$ sudo st2ctl status
##### st2 components status #####
st2actionrunner is not running.
st2api is not running.
st2stream is not running.
st2auth is not running.
st2garbagecollector is not running.
st2notifier is not running.
st2resultstracker is not running.
st2rulesengine is not running.
st2sensorcontainer is not running.
st2chatops PID: 2833
st2chatops PID: 2834
mistral-server is not running.
mistral-api is not running.
```

For some reason, `st2` services were not enabled via packaging:
```sh
[vagrant@centos6 ~]$ chkconfig | grep -e mistral -e st2 -e mongo -e rabbit
mistral        	0:off	1:off	2:off	3:off	4:off	5:off	6:off
mistral-api    	0:off	1:off	2:off	3:off	4:off	5:off	6:off
mistral-server 	0:off	1:off	2:off	3:off	4:off	5:off	6:off
mongod         	0:off	1:off	2:on	3:on	4:on	5:on	6:off
rabbitmq-server	0:off	1:off	2:on	3:on	4:on	5:on	6:off
st2actionrunner	0:off	1:off	2:off	3:off	4:off	5:off	6:off
st2api         	0:off	1:off	2:off	3:off	4:off	5:off	6:off
st2auth        	0:off	1:off	2:off	3:off	4:off	5:off	6:off
[...]

PR #461, opened 2017-06-05
Enable services on el6, and check if enabled

warrenvw:
Fixes https://github.com/StackStorm/st2-packages/issues/424

Enable services in el6 on run level 2,3,4,5.

Test if the services are enabled in runlevel 2,3,4,5 for all distributions.


rake/spec/default/60-st2_all-services-ok_spec.rb
```diff
@@ -59,13 +59,17 @@
...
 describe 'st2 services' do
   include_examples 'show service log on failure'
 
   spec[:service_list].each do |name|
     describe service(name), prompt_on_failure: true do
       it { is_expected.to be_running }
+      it { should be_enabled.with_level(2) }
+      it { should be_enabled.with_level(3) }
+      it { should be_enabled.with_level(4) }
+      it { should be_enabled.with_level(5) }
```
armab: Also, just asking do we really need explicitly to check `2, 3, 4, 5` runlevels?
What is the default for a single `should be_enabled`?

Why I'm asking, because this looks too long in the CircleCI build logs:

PR #463, opened 2017-06-06
Remove old st2 build webhook URL

armab:
Removes outdated st2 webhook URL. It shows up in CircleCI UI with 503 status.


warrenvw:
Woot! Thanks @armab

PR #464, opened 2017-06-13
Stop mistral-server with SIGINT instead of SIGTERM

andrew-regan:
I couldn't find any signal handling in Mistral but did find handling for `KeyboardInterrupt` and verified that Ctrl-C would successfully stop mistral-server. This change uses the same SIGINT signal to stop mistral-server from systemd.

### DIFF
--- a/packages/st2mistral/rpm/mistral-server.service
+++ b/packages/st2mistral/rpm/mistral-server.service
@@ -10,6 +10,7 @@ Environment="SERVER_ARGS=--config-file /etc/mistral/mistral.conf --log-file /var
 Environment="COMPONENTS=api,engine,executor"
 EnvironmentFile=-/etc/sysconfig/mistral
 ExecStart=/bin/sh -c ". /opt/stackstorm/mistral/share/sysvinit/helpers; enabled_list -q server || exit 0; exec /opt/stackstorm/mistral/bin/mistral-server --server $(enabled_list server) $SERVER_ARGS"
+KillSignal=SIGINT
 
 [Install]
 WantedBy=multi-user.target


Kami:
Just curious - was there any specific reason for that change?

andrew-regan:
Sorry....should have stared with this:
_`mistral-server` currently does not responding to `systemctl stop mistral` which eventually results in systemd forcefully terminating the process. I believe the root cause of this is Mistral not handling `SIGTERM` which is the default signal used by systemd. _

andrew-regan:
Capture of the Slack conversation on this one:
screen shot 2017-06-15 at 9 52 24 am

Kami:
Thanks for clarifying, that makes sense.

I believe Python should already be handling SIGTERM, but it's possible something else is going on because it's a long term process using eventlet.

Kami:
I was able to reproduce the issue with `mistral-server` on RHEL 7.3, but `mistral-api` seems to be working fine.

Kami:
I was also able to confirm fix is working as intended.

Thanks again for the fix. Will go ahead and merge it.

andrew-regan:
Same observation w/ `mistral-api` stopping fine.

Kami:
Actually, I might have spoken too fast.

I also need to confirm that with SIGINT, mistral-server still correctly waits for all the running tasks to finish before exiting, because that's the desired behavior. If it won't wait, this will cause issues.

Kami:
`2017-06-15 11:42:39.562 12643 INFO oslo_service.service [-] Caught SIGINT signal, instantaneous exiting`

So yeah, it immediately exists even if there are tasks running.

I don't think this is desired, but I will wait for @m4dcoder to confirm this.

Kami:
To clarify - by default, all long running processes should handle `SIGTERM` with a graceful exit.

This means finish all the processing before exiting (finish all the processing, close connections, etc.).

That's also how most service managers work by default - first send `SIGTERM`, wait for some pre-defined (or set by the user) threshold seconds and if process still doesn't exit by then, send `SIGKILL` and kill the process forecefully.

So in short, to me it seems like it's working as intended and designed, but I could also be wrong so will wait on @m4dcoder to chime in.

andrew-regan:
@Kami so do you see Mistral server handle `SIGTERM` currently? I was running `systemctl stop mistral` against what I thought was an idle Mistral server and didn't see anything in the logs indicating it was trying to shutdown and/or waiting for something internal to stop. I waited beyond the 1 minute systemd timer, but can't remember how long exactly.

Kami:
@andrew-regan You are right - I don't see `SIGTERM` being handled (at least from the logs), but it looks like it should be.

We should try to gracefully shut down first, because using sigint and immediately exiting would cause issues with abandoned tasks.

It sounds like we might need a change in mistral-server, let's see what @m4dcoder says.

andrew-regan:
Thanks for confirming and agree with waiting to here what @m4dcoder has to say. 😄

m4dcoder:
Mistral handles SIGTERM. It's not done in Mistral but the OpenStack oslo.service which Mistral inherits from. The code reference is at https://github.com/openstack/oslo.service/blob/master/oslo_service/service.py#L292. I went back and hand tested the systemd service shutdown on both u16 and centos7. On service shutdown, SIGTERM is sent and the stop method is being called.

m4dcoder:
I aslo tested this patch. Even w/ the KillSignal specified as SIGINT in this patch., I'm seeing SIGTERM being process on centos7. So I'm not sure the relevance of this PR.

andrew-regan:
@m4dcoder I'm seeing this behavior with the version of Mistral in StackStorm 2.2.1. Any chance you're testing against a different version?

m4dcoder:
I'm testing this on v2.3 which we will release in a few days if not sooner.  There's no major changes in Mistral from v2.2 to v2.3 and certainly not in how the oslo services work.

m4dcoder:
If you want to try test yourself, you can edit /opt/stackstorm/mistral/lib/python2.7/site-packages/oslo_service/service.py and place more LOG.info statements, recycle the mistral service, and review the mistral log.

Mierdin:
Interestingly this doesn't seem to be limited to systemd - also happening on Ubuntu 14.04:

```
vagrant@st2vagrant:~$ time sudo service mistral-server stop
mistral-server stop/waiting

real    1m0.023s
user    0m0.002s
sys     0m0.010s
vagrant@st2vagrant:~$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.5 LTS"
```

Kami:
@m4dcoder It's also worth noting I was able to reproduce the issue (SIGTERM not being handled) using latest v2.3.0 packages from staging stable.

And once you update init file you need to run systemd reload command for the config change to take an affect.

andrew-regan:
@Mierdin I could reproduce it outside of systemd by just starting the mistral-server process in the shell. Trying to stop it via `kill` which uses `SIGTERM` would not do anything as far as I could tell.

m4dcoder:
@Kami @Mierdin I'll leave this to you since you agree there's a problem here.  As I said, in my tests, I clearly see SIGTERM being triggered (although i agree that it takes a little while to fully shutdown which I don't know why). But I can't reproduce what this PR is concerned.

m4dcoder:
@andrew-regan if you run the process outside of systemd and kill it (ctrl-C), oslo service sees it as SIGINT.  I tested this as well. That I maybe agree needs to be handled as SIGTERM but need more research. But systemd stop sends SIGTERM which is received as SIGTERM by oslo service in my tests.

armab:
FYI there is a patch submitted by @m4dcoder in upstream mistral to address this slow shutdown issue in the right way: https://review.openstack.org/#/c/499853

andrew-regan:
Close this PR?

armab:
This is fixed in mistral upstream and backported into StackStorm `2.4.1` https://stackstorm.com/2017/09/18/quick-one-st2-2-4-1-released/

PR #466, opened 2017-07-01
Remove sleep timeout before Register content

armab:
Bug https://github.com/StackStorm/st2-packages/issues/445 was fixed by @Kami in https://github.com/StackStorm/st2/pull/3542 so we can remove this ugly `sleep 5` timeout from our `curl | bash` installer.

+1 step to improve & stablizie st2 availability/deployments/HA/prod.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -329,10 +329,6 @@ install_st2() {
   sudo crudini --set /etc/st2/st2.conf database password "${ST2_MONGODB_PASSWORD}"
 
   sudo st2ctl start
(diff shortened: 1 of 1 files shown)


armab:
+1 step to improve & stablizie st2 availability/deployments/HA/prod.

So after merging, the only nasty timeout left would be this one: https://github.com/StackStorm/st2-packages/blob/1321cb64bff52a146dc40b42cda06294b37089f9/scripts/st2bootstrap-deb.sh#L556-L560
caused by the following bug related to RabbitMQ: https://github.com/StackStorm/st2/issues/3290

Kami:
:+1:

PR #472, opened 2017-07-21
Add "sudo" to "yum list" for libffi-devel check

Mierdin:
`yum list` requires sudo (at least on the RHEL6 AMI that I was testing with), so this always failed. Using `rpm` doesn't, and should prove a bit more reliable.


scripts/st2bootstrap-el6.sh
```diff
@@ -170,8 +170,7 @@ If you still have questions, please contact support. Alternatively, you can use
 CentOS 6 for evaluation.
 EHD
 )
-  yum list libffi-devel 1>/dev/null 2>&1 || no_libffi_devel=$?
-  if [ ! -z "$no_libffi_devel" ]; then
+  if [ -z "$(rpm -qa libffi-devel)" ]; then
```
armab: I get it now.

Modified version checks if package is installed, while `yum list libffi-devel` checks if package is _available_ to be installed as a dependency.

So I think the previous logic is correct, just add `sudo` if `yum list` didn't work for you.

Mierdin: Sounds good, will revert and apply `sudo`

Kami:
Wonder if that's RHEL 6 specific because on RHEL 7, sudo is not required.

PR #477, opened 2017-08-01
Install MongoDB 3.4 by default

Kami:
Our code has supported 3.4 and we have been running tests with MongoDB 3.4 for a long time so it's time to switch to it in StackStorm v2.4.0.

Resolves https://github.com/StackStorm/st2/issues/3592.


Kami:
@armab Thanks - updated it in this repo.

And looks like we already use 3.4 here - https://github.com/StackStorm/st2-docker/blob/master/docker-compose.yml

PR #480, opened 2017-08-07
Remove 'sleep 30' timeout hack for Xenial

armab:
Xenial stuck bug is fixed in https://github.com/StackStorm/st2/issues/3290 so we can remove nasty `sleep 30` hack from our `curl|bash` installer.

Big thanks @Kami for diving deep to find the root cause & fixing it.


Kami:
:+1:

PR #481, opened 2017-08-11
Fix CI docker volumes cleanup

armab:
Associated change in st2: https://github.com/StackStorm/st2/pull/3668

Don't fail when cleaning up via docker-compose.
See: https://discuss.circleci.com/t/no-space-left-on-device-while-creating-mongo/11532/13


Kami:
:+1:

PR #483, opened 2017-08-15
Switch to st2 pack install

### LINKED ISSUE #482: Installer scripts should use `st2 pack install`
The bootstrap scripts have this line:

```sh
st2 run packs.install packs=st2
```

This should just use the new pack install commands, i.e. `st2 pack install st2`

LindsayHill:
Switch to using `st2 pack install` rather than legacy `st2 run packs.install`

Fixes #482


Kami:
Good catch :+1:

PR #492, opened 2017-09-04
Print a more user-friendly error message in case we fail to retrieve package version

Kami:
Right now we swallow curl error / output so we end up with a lovely totally un-friendly error message like that:

```bash
Get:2 http://us-west-2.ec2.archive.ubuntu.com/ubuntu xenial/universe amd64 jq amd64 1.5+dfsg-1 [144 kB]
Fetched 232 kB in 0s (21.4 MB/s)
Selecting previously unselected package libonig2:amd64.
(Reading database ... 53986 files and directories currently installed.)
Preparing to unpack .../libonig2_5.9.6-1_amd64.deb ...
Unpacking libonig2:amd64 (5.9.6-1) ...
Selecting previously unselected package jq.
Preparing to unpack .../jq_1.5+dfsg-1_amd64.deb ...
Unpacking jq (1.5+dfsg-1) ...
Processing triggers for man-db (2.7.5-1) ...
Setting up libonig2:amd64 (5.9.6-1) ...
Setting up jq (1.5+dfsg-1) ...
Processing triggers for libc-bin (2.23-0ubuntu9) ...
# Failed on step - Install st2 #
```

## TODO

- [ ] Backport to v2.4

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -52,6 +52,7 @@ setup_args() {
           --password=*)
           PASSWORD="${i#*=}"
           shift
+          ;;
           --force-branch=*)
           FORCE_BRANCH="${i#*=}"
           shift
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -16,6 +16,7 @@ PASSWORD=''
 
 # Note: This variable needs to default to a branch of the latest stable release
 BRANCH='v2.4'
+FORCE_BRANCH=""
 
 setup_args() {
   for i in "$@"
@@ -50,6 +51,9 @@ setup_args() {
           ;;
           --password=*)
           PASSWORD="${i#*=}"
+          shift
+          --force-branch=*)
+          FORCE_BRANCH="${i#*=}"
           shift
           ;;
           *)
@@ -119,6 +123,11 @@ fi
 
 if [[ "$DEV_BUILD" != '' ]]; then
   DEV_BUILD="--dev=${DEV_BUILD}"
+fi
+
+if [[ "${FORCE_BRANCH}" != "" ]]; then
+    echo "Using branch ${FORCE_BRANCH}"
+    BRANCH=${FORCE_BRANCH}
 fi
 
 USERNAME="--user=${USERNAME}"
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -127,8 +127,7 @@ if [[ "$DEV_BUILD" != '' ]]; then
 fi
 
 if [[ "${FORCE_BRANCH}" != "" ]]; then
-    echo "Using branch ${FORCE_BRANCH}"
-    BRANCH=${FORCE_BRANCH}
+  BRANCH=${FORCE_BRANCH}
 fi
 
 USERNAME="--user=${USERNAME}"
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -53,6 +53,9 @@ setup_args() {
           PASSWORD="${i#*=}"
           shift
           ;;
+          # Used to specify which branch of st2-packages repo to use. This comes handy when you
+          # need to use a non-master branch of st2-package repo (e.g. when testing installer script
+          # changes which are in a branch)
           --force-branch=*)
           FORCE_BRANCH="${i#*=}"
           shift
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -317,7 +317,19 @@ install_st2() {
     sudo apt-get install -y st2${ST2_PKG_VERSION}
   else
     sudo apt-get install -y jq
+
+    echo "Retrieving package from https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts"
+
+    # Note: We disable global error handler because we want a more user-friendly error message
+    set +e
     PACKAGE_URL="$(curl -Ss -q https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts | jq -r '.[].url' | egrep "${SUBTYPE}/st2_.*.deb")"
+    set -e
+
+    if [ ! ${PACKAGE_URL} ]; then
+        echo "Failed to retrieve package from https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts"
+        exit 2
+    fi
+
     PACKAGE_FILENAME="$(basename ${PACKAGE_URL})"
     curl -Ss -k -o ${PACKAGE_FILENAME} ${PACKAGE_URL}
     sudo dpkg -i --force-depends ${PACKAGE_FILENAME}
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -127,7 +127,7 @@ function get_package_url() {
   fi
 
   PACKAGES_URLS="$(echo ${PACKAGES_METADATA}  | jq -r '.[].url')"
-  PACKAGE_URL=$(echo "${PACKAGES_URLS}" | egrep "${DISTRO}/.*${PACKAGE_NAME_REGEX}")
+  PACKAGE_URL=$(echo "${PACKAGES_URLS}" | egrep "${DISTRO}/${PACKAGE_NAME_REGEX}")
 
   if [ -z "${PACKAGE_URL}" ]; then
       echo "Failed to find url for ${DISTRO} package (${PACKAGE_NAME_REGEX})"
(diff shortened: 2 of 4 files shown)


Kami:
I also pushed a change which allows user to specify which branch of `st2-packages` to use by passing `--force-branch` argument to the installer script.

Right now it's impossibly to specify which branch to use which makes testing changes in branch, etc. cumbersome (you need to hack up the local installer script or similar.

Kami:
With this change the output looks like this:

```bash
Retrieving packages metadata from https://circleci.com/api/v1.1/project/github/StackStorm/st2/5646/artifacts
Failed to find url for xenial deb package
Circle CI response: [ ]
# Failed on step - Install st2 #
```

scripts/st2_bootstrap.sh
```diff
@@ -52,6 +53,10 @@ setup_args() {
           PASSWORD="${i#*=}"
           shift
           ;;
+          --force-branch=*)
```
armab: I think we're plastering this script with more unpopular features used just for a couple of times, but I'm OK if you really need it.

Just please include some basic description in comments, at least will help to know what is `--force-branch` means.

scripts/st2bootstrap-deb.sh
```diff
@@ -317,7 +317,28 @@ install_st2() {
     sudo apt-get install -y st2${ST2_PKG_VERSION}
   else
     sudo apt-get install -y jq
-    PACKAGE_URL="$(curl -Ss -q https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts | jq -r '.[].url' | egrep "${SUBTYPE}/st2_.*.deb")"
+
+    echo "Retrieving packages metadata from https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts"
```
armab: We'll need the same logic for `install_st2mistral()` function and `st2bootstrap-el6.sh` with `st2bootstrap-el7.sh` scripts as well.

scripts/st2_bootstrap.sh
Kami: I think it's a good idea to have this flag to make it possible to test with a non-master branch.

In fact, I believe we had it at some point, but dunno what happened to it.

Kami: I agree comment is a good idea.

Added in f574f47418dc845a8b29d5dc30939b1757e5be39.

scripts/st2bootstrap-deb.sh
Kami: I can do that, but my eventual goal was to move all the common functionality in common.sh or similar and build final script from that (concatenating things together) as opposed to more copy and paste...

Kami: For now, I will refactor this in a function, copy and paste it in all the files and work on the concatenation stuff (long overdue) once this PR is merged to make changes easier to review.

Kami:
I pushed a refactor change in d2a31f89307bd7f678c42f8520022f28d2e348dc, but I'm still a bit confused about the whole grep and code path.

It looks like it never worked, but it's possible I'm missing something.

Let's take https://circleci.com/api/v1.1/project/github/StackStorm/st2/4800/artifacts for an example.


```bash
kami ~/w/stackstorm/st2-packages (git:pkg_version_friendly_error)$ echo "https://4800-19051806-gh.circle-artifacts.com/1/packages/xenial/build/st2_2.4dev-55_amd64.deb" | egrep "xenial/st2_.*.*deb"
kami ~/w/stackstorm/st2-packages (git:pkg_version_friendly_error)$ 
kami ~/w/stackstorm/st2-packages (git:pkg_version_friendly_error)$ echo "https://4800-19051806-gh.circle-artifacts.com/1/packages/xenial/build/st2_2.4dev-55_amd64.deb" | egrep "xenial.*/st2_.*.*deb"
https://4800-19051806-gh.circle-artifacts.com/1/packages/xenial/build/st2_2.4dev-55_amd64.deb

```

It looks like existing regex doesn't match anything because it's too strict (there is no url which matches that pattern since there is always "builds/" suffix after the distro type).

Another possible option would be that Circle CI changed the API response format (added `builds/` suffix to the URL path), but I guess that's unlikely.

@armab @enykeev Any ideas? It looks like egrep change has been made in December (41a0fefae33c7b55bba818e4b169bd5a91f149ba) so it's very unlikely it was broken / unused since then.

Kami:
Changing it to `  PACKAGE_URL=$(echo "${PACKAGES_URLS}" | egrep "${SUBTYPE}/build/${PACKAGE_NAME_REGEX}")` works.

Is it possible that Circle CI changed the API response format and that's also the reason why builds have been failing for 10+ days?

Kami:
Based on the discussion on Slack with @armab it looks like that the URL format has indeed changed a while back, but it should work correctly with the latest packages.

In any case, I still pushed a change so it works with both - old and new url format to avoid surprises.

armab:
So as figured in Slack, the main reason behind the issue is trying to refer to the old build number `#` (1+ month old), which happened during transition time when we experimented with CircleCI 2.0/TravisCI and there were a lot of changes/experiments.

scripts/st2bootstrap-el7.sh
```diff
@@ -111,6 +111,31 @@ install_yum_utils() {
...
+
+  PACKAGES_METADATA=$(curl -Ss -q https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts)
+
+  if [ -z "${PACKAGES_METADATA}" ]; then
+      echo "Failed to retrieve packages metadata from https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts"
+      exit 2
+  fi
+
+  PACKAGES_URLS="$(echo ${PACKAGES_METADATA}  | jq -r '.[].url')"
+  PACKAGE_URL=$(echo "${PACKAGES_URLS}" | egrep "${DISTRO}/.*${PACKAGE_NAME_REGEX}")
```
armab: `"${DISTRO}/.*${PACKAGE_NAME_REGEX}"`

Making `.*` not part of `PACKAGE_NAME_REGEX` looks too greedy to me and can match more unexpected, rather than help.
I'd prefer more explicit match to make sure we don't grab any accidental/temporary package name or dir, outside of expected one.

`tmp-st2-2.5dev-1.x86_64.rpm`, `test/st2_2.5dev-2_amd64.deb`

scripts/st2bootstrap-deb.sh
```diff
@@ -113,6 +113,31 @@ setup_args() {
...
+  # Retrieve direct package URL for the provided dev build, subtype and package name regex.
+  DEV_BUILD=$1 # Repo name and build number - <repo name>/<build_num> (e.g. st2/5646)
+  DISTRO=$2  # Distro name (e.g. trusty,xenial,el6,el7)
+  PACKAGE_NAME_REGEX=$3
+
+  PACKAGES_METADATA=$(curl -Ss -q https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts)
+
+  if [ -z "${PACKAGES_METADATA}" ]; then
+      echo "Failed to retrieve packages metadata from https://circleci.com/api/v1.1/project/github/StackStorm/${DEV_BUILD}/artifacts"
+      exit 2
```
armab: We can use more clean `return` instead of `exit` for a function

scripts/st2bootstrap-deb.sh
```diff
@@ -460,7 +495,17 @@ install_st2mistral() {
...
+
+    # Note: We disable global error handler because we want to print a more user-friendly error message
+    set +e
+    PACKAGE_URL=$(get_package_url "${DEV_BUILD}" "${SUBTYPE}" "st2mistral_.*.deb")
+
+    if [ -z "${PACKAGE_URL}" ]; then
+        exit 2
+    fi
+    # Re-enable a global error handler
+    set -e
```
armab: Definitely need to reconsider `set +e`/`set -e` here and think about cleaner approach.

Maybe `return` from a function (as noted in previous comment) would help to get rid of this disable/enable `e` hack.

scripts/st2bootstrap-deb.sh
Kami: Sadly return won't make a difference with regards to `set +e`. If `-e` mode is enabled, global error handler will still be called in case the function exits with non zero or returns non zero.

Kami:
Pushed a change to utilize global error handler and use return instead of exit - 049db53039570630ac201a964d8dd199e389563e.

It's also worth noting that using either exit / return has the same end result when `set -e` is used.

Kami:
Yeah, will test it on all the distros using those commands:

```bash
wget https://raw.githubusercontent.com/StackStorm/st2-packages/pkg_version_friendly_error/scripts/st2_bootstrap.sh
bash st2_bootstrap.sh --dev=st2/5651 --staging --unstable --user=st2admin --password=Ch@ngeMe --force-branch=pkg_version_friendly_error
```

(see, the `--force-branch` argument comes handy :P)

Kami:
Confirmed it's working on all the supported distros, merging.

PR #493, opened 2017-09-06
Automatically generate final install script files from a template and included script files

Kami:
This one is long overdue.

Right now, there is a bunch of code duplicates and copy and pasted across all the installer script files. This means it's hard to maintain it.

The goal of this PR is to move common functionality and code in common files which are re-used by all the installer script files.

Final installer script files are now automatically generated from a base template and include files. This way the end result is the same - single script file which is easy to wget, etc.

The next iteration will be to use those include snippets inside the documentation - right now those snippets are manually copy and pasted inside the docs (hard to maintain, easy to get up of date / broken).


LindsayHill:
Haven't gone through all the details here, but I am +1 with this approach. Was thinking about it recently in the context of install documentation, where we could make greater use of `include` sections for common stuff. You're right about pulling in some of this content too - then we could get rid of that note saying "read the source". 

Possibly use the `:lines:`, `:start-after:` and `:end-before:` directives with `literalinclude`, to only pull in specific sections as needed.

warrenvw:
On closer inspection (using byte-by-byte compare of sorted functions), it looks like a few changes need to be made. I'll submit a new commit shortly.

 - [ ] `st2bootstrap-deb.sh` is missing `install_st2mistral_dependencies` function.
 - [ ] `st2bootstrap-el7.sh` is missing `sudo yum install -y net-tools`.
 - [ ] `st2bootstrap-el7.sh` fix spelling:  `acces`

Makefile
```diff
@@ -0,0 +1,25 @@
+.PHONY: scripstgen
```
humblearner: s/scripstgen/scriptsgen

scripts/includes/common.sh
```diff
@@ -0,0 +1,248 @@
...
+
+  sudo st2ctl restart-component st2api
+}
+
+
+verify_st2() {
+  st2 --version
+  st2 -h
+
+  st2 auth $USERNAME -p $PASSWORD
```
humblearner: Should we replace `st2 auth` with `st2 login` ?

scripts/includes/common.sh
```diff
@@ -0,0 +1,248 @@
...
+  echo ""
+  echo "  st2 is installed and ready to use."
+  echo ""
+  echo "Head to https://YOUR_HOST_IP/ to access the WebUI"
+  echo ""
+  echo "Don't forget to dive into our documentation! Here are some resources"
+  echo "for you:"
+  echo ""
+  echo "* Documentation  - https://docs.stackstorm.com"
+  echo "* Knowledge Base - https://stackstorm.reamaze.com"
```
humblearner: These are old articles, should not point users to reamaze or update them.

scripts/includes/common.sh
LindsayHill: I am inclined to say yes, we should move to to `st2 login`...but since this PR is already pretty big, maybe it should mostly just be "a better way of generating the existing content", and then we do future PRs to improve the actual content?

warrenvw: Since this PR is all about consolidating common code and auto generating the bootstrap scripts, and we're going to be making many other small changes to these scripts.. I vote that we handle this change in a subsequent PR.

scripts/includes/common.sh
warrenvw: Yep. I considered changing now, but ultimately decided to do in a subsequent PR.

Makefile
warrenvw: 👍

warrenvw:
Thanks @humblearner and @LindsayHill for your feedback. And THANK YOU @Kami ami for the PR itself. :)

scripts/includes/common.sh
Kami: I agree - those changes should be made in another PR once this one is merged.

Kami:
Thanks for approving this - will go ahead and merge it.

Kami:
@warrenvw Thanks for making those changes - be2a389102037783dbad7648d1304507dfac604d.

Kami:
Will test this with dev packages to make sure it works everywhere before merging:

- [ ] Ubuntu 16
- [ ] EL6
- [ ] EL7

Kami:
Everything looked OK so I went ahead and merged it.

warrenvw:
Hurrah!

PR #494, opened 2017-09-07
Added st2-cleanup-db references

nmaludy:
The PR in the main StackStorm repo https://github.com/StackStorm/st2/pull/3669 adds a new command `st2-cleanup-db`. This new command needs to be added to packaging.

DO NOT MERGE UNTIL  https://github.com/StackStorm/st2/pull/3669 IS MERGED


nmaludy:
@armab and i agree that this should not be part of the symlinks created in `/usr/bin/*`  so i'm closing this PR.

Instead the binary will live in `/opt/stackstorm/st2/bin/` and it will need to be invoked from that path.

PR #496, opened 2017-09-09
v2.2: Fix CentOS7 st2chatops installation failure

LindsayHill:
Same as #484, but for v2.2.

Temporary hack to install removed in EPEL 'http-parser' package until proper upstream fix https://bugs.centos.org/view.php?id=13669


armab:
👍

PR #505, opened 2017-09-29
Do not hardcode home paths

### LINKED ISSUE #501: Installation script should handle non-standard HOME location
The installation scripts assume that ~stanley is `/home/stanley`. 

Some environments use a different base path for home directories, e.g. `/export/home/$USER`. This causes problems with the installation script which assumes that ~stanley's SSH key is in `/home/stanley/.ssh`. This should be identified automatically.

warrenvw:
This PR resolves #501 .

Remaining tasks:

 - [x] Testing
 - [x] Addressing review comments


scripts/includes/common.sh
```diff
@@ -1,3 +1,9 @@
+function get_home() {
+  USER=$1
```
armab: let's use local variable here, instead of overriding `$USER`

scripts/includes/common.sh
```diff
@@ -100,23 +106,25 @@ generate_random_passwords() {
...
 configure_st2_user () {
   # Create an SSH system user (default `stanley` user may be already created)
-  if (! id stanley 2>/dev/null); then
-    sudo useradd stanley
+  if (! id $USERNAME 2>/dev/null); then
+    sudo useradd $USERNAME
   fi
 
-  sudo mkdir -p /home/stanley/.ssh
+  HOME=$(get_home ${USERNAME})
```
armab: Same here, please use local variable, instead of overriding `$HOME`

scripts/includes/common.sh
Kami: Good point :+1:

scripts/includes/common.sh
```diff
@@ -100,23 +106,25 @@ generate_random_passwords() {
...
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  sudo ssh-keygen -f ${HOME}/.ssh/${USERNAME}_rsa -P ""
 
   # Authorize key-base access
-  sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'
-  sudo chmod 0600 /home/stanley/.ssh/authorized_keys
-  sudo chmod 0700 /home/stanley/.ssh
-  sudo chown -R stanley:stanley /home/stanley
+  sudo sh -c "cat ${HOME}/.ssh/${USERNAME}_rsa.pub >> ${HOME}/.ssh/authorized_keys"
```
Kami: I would prefer to be more explicit and call this variable `STANLEY_HOME` or similar (becase `$HOME` really represents home of a user which is running the script which is usually not stanley.

scripts/includes/common.sh
```diff
@@ -1,3 +1,9 @@
+function get_home() {
+  USER=$1
+  eval echo "~$USER";
```
armab: Also after some thinking, do we need this function to be flexible and rely on user param?

I mean we always rely on `stanley` in default installer and with that we can just simplify this to:
```diff
- eval echo "~$USER"
+ echo ~stanley
```

and get rid of nasty `eval`

scripts/includes/common.sh
warrenvw: I will make this change.

scripts/includes/common.sh
warrenvw: If we want to always rely on stanley in default installer, then we can simplify this. With this in mind, I'll make the necessary changes.

scripts/includes/common.sh
```diff
@@ -100,23 +105,26 @@ generate_random_passwords() {
 
 configure_st2_user () {
   # Create an SSH system user (default `stanley` user may be already created)
-  if (! id stanley 2>/dev/null); then
-    sudo useradd stanley
+  SYSTEM_USER=stanley
```
armab: I don't think we should assign `stanley` to a var in `curl|bash` installer which is tend to be simple as possible.
I think that `stanley` should be hardcoded everywhere as it was before and keep it simple.

Additionally, keep in mind that `docs` instructions should be modified to be in sync with the installer. So that's another point to have less expressions, constructs and functions.
Think also how you'd modify respective doc section: https://docs.stackstorm.com/install/deb.html#configure-ssh-and-sudo

Not sure what others think? cc @Kami

scripts/includes/common.sh
```diff
@@ -100,23 +105,26 @@ generate_random_passwords() {
...
+  if (! id ${SYSTEM_USER} 2>/dev/null); then
+    sudo useradd ${SYSTEM_USER}
   fi
 
-  sudo mkdir -p /home/stanley/.ssh
+  SYSTEM_HOME=$(get_home ${SYSTEM_USER})
```
armab: Can we go with simpler `SYSTEM_HOME=$(echo ~stanley)` and avoid function and `eval`?

If not, should we include the `get_home` function in respective doc as well?
https://docs.stackstorm.com/install/deb.html#configure-ssh-and-sudo

scripts/includes/common.sh
Kami: +1

scripts/includes/common.sh
```diff
@@ -193,7 +201,7 @@ verify_st2() {
   st2 --version
   st2 -h
 
-  st2 auth $USERNAME -p $PASSWORD
+  st2 login -p $PASSWORD -w $USERNAME
```
Kami: Yeah, we need to confirm login stuff works correctly (e.g. if you run it as sudo as a random user or if you run it directly as root).

Login caches token locally and it writes it for user who executed the script so we need to make sure it works as expected.

scripts/includes/common.sh
warrenvw: Yes, it actually worked this time. I have no idea why my previous tests without the `eval` had failed. I've gone ahead and removed the `get_home()` function.

Issue #503, opened 2017-09-28
Installation script should cope with pre-existing stanley SSH key


LindsayHill:
If there is a pre-existing key for stanley, the installation script will fail. This causes problems when re-running the install after a failed attempt. It should not fail at this point, but should use that existing key.

PR #506, opened 2017-10-03
Generate ssh key if it does not already exist

warrenvw:
Resolves https://github.com/StackStorm/st2-packages/issues/503.

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -107,7 +107,11 @@ configure_st2_user () {
   sudo mkdir -p /home/stanley/.ssh
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  # NOTE: If the file already exists and is non-empty, then assume the key does not need
+  # to be generated again.
+  if [ ! -s /home/stanley/.ssh/stanley_rsa ]; then
+    sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  fi
 
   # Authorize key-base access
   sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'


scripts/includes/common.sh
```diff
@@ -107,7 +107,11 @@ configure_st2_user () {
   sudo mkdir -p /home/stanley/.ssh
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  # NOTE: If the file already exists and is non-empty, then assume the key does not need
+  # to be generated again.
+  if [ ! -s /home/stanley/.ssh/stanley_rsa ]; then
+    sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  fi
```
armab: I think this idempotent `if` block should include more logic from the lines below, related to authorized keys and chmod.

scripts/includes/common.sh
```diff
@@ -107,7 +107,11 @@ configure_st2_user () {
...
 
   # Authorize key-base access
   sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'
```
armab: For example, keeping in mind that previous command could be run several times (keeping in mind idempotence story), here the script will add stanley public key several times which is wrong.

warrenvw:
Sorry, I'd pushed this change without testing. The next commit will be tested first.

scripts/includes/common.sh
```diff
@@ -107,16 +107,28 @@ configure_st2_user () {
   sudo mkdir -p /home/stanley/.ssh
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
-  sudo ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+  # NOTE: If the file already exists and is non-empty, then assume the key does not need
+  # to be generated again.
+  if [ ! -s /home/stanley/.ssh/stanley_rsa ]; then
```
armab: Same issue with `sudo` permissions here.

PR #507, opened 2017-10-03
Generate crypto key only if non-existent

### LINKED ISSUE #504: Installation should cope with pre-existing ST2 datastore encryption key
If there is a pre-existing ST2 datastore encryption key (e.g. from a failed installation attempt), the installation will fail. It should just use the existing key, and not attempt to create a new one.

warrenvw:
Resolves #504

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -181,7 +181,6 @@ generate_symmetric_crypto_key_for_datastore() {
   fi
 
   # Make sure only st2 user can read the file
-  sudo usermod -a -G st2 st2
   sudo chgrp st2 ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
   sudo chmod o-r ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
   sudo chgrp st2 ${DATASTORE_ENCRYPTION_KEY_PATH}
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -173,7 +173,12 @@ generate_symmetric_crypto_key_for_datastore() {
   DATASTORE_ENCRYPTION_KEY_PATH="${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}/datastore_key.json"
 
   sudo mkdir -p ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
-  sudo st2-generate-symmetric-crypto-key --key-path ${DATASTORE_ENCRYPTION_KEY_PATH}
+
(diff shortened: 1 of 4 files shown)


scripts/includes/common.sh
```diff
@@ -173,7 +173,12 @@ generate_symmetric_crypto_key_for_datastore() {
...
-  sudo st2-generate-symmetric-crypto-key --key-path ${DATASTORE_ENCRYPTION_KEY_PATH}
+
+  # If the file ${DATASTORE_ENCRYPTION_KEY_PATH} exists and is not empty, then do not generate
+  # a new key. st2-generate-symmetric-crypto-key fails if the key file already exists.
+  if [ ! -s ${DATASTORE_ENCRYPTION_KEY_PATH} ]; then
+    sudo st2-generate-symmetric-crypto-key --key-path ${DATASTORE_ENCRYPTION_KEY_PATH}
+  fi
 
   # Make sure only st2 user can read the file
   sudo usermod -a -G st2 st2
```
armab: This line could be removed, as we did in `st2docs` just a few days before: https://github.com/StackStorm/st2docs/pull/620

> `st2` user is already part of `st2` group, added by [`deb`](https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/preinst#L36) and [`rpm`](https://github.com/StackStorm/st2-packages/blob/master/packages/st2/rpm/preinst_script.spec#L21) packaging.

scripts/includes/common.sh
```diff
@@ -173,10 +173,14 @@ generate_symmetric_crypto_key_for_datastore() {
   DATASTORE_ENCRYPTION_KEY_PATH="${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}/datastore_key.json"
 
   sudo mkdir -p ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
-  sudo st2-generate-symmetric-crypto-key --key-path ${DATASTORE_ENCRYPTION_KEY_PATH}
+
+  # If the file ${DATASTORE_ENCRYPTION_KEY_PATH} exists and is not empty, then do not generate
+  # a new key. st2-generate-symmetric-crypto-key fails if the key file already exists.
+  if [ ! -s ${DATASTORE_ENCRYPTION_KEY_PATH} ]; then
```
armab: Ouch,
after taking a second look at it, @warrenvw this looks like the same issue like in previous PR.

`if [ ! -s ${DATASTORE_ENCRYPTION_KEY_PATH} ]; then` might not work for non-root user which doesn't use `sudo` for check.

Issue #499, opened 2017-09-28
Write install logs to file as well as stdout


LindsayHill:
When the installation scripts run, they produce large amounts of output to screen. This can be useful information for debugging purposes.

Unfortunately it is only displayed in that session. Sometimes users window buffers don't contain enough info, or they are doing it via the VM console, making it difficult to capture. Or they log out of that session, and lose the content.

Then when we want to debug installation failures, we don't have the content.

During installation, we should write all output to a file, as well as to screen, e.g. using `tee` or similar.

This file should be stored somewhere sensible, perhaps in `/var/log/st2/install.log.<timestamp>`, or maybe `$HOME/st2.install.log.<timestamp>`

warrenvw:
I'm 👍 for `/var/log/st2/install.log.<timestamp>`.

Does it make sense to include the output of commands run prior to installation that identify OS version and any other useful data. e.g. `uname -a`, `cat /etc/lsb-release` or `cat /etc/redhat-release` ..., `netstat -an`, `st2 --version`, ... ? These are always questions that support folks have to ask of people reporting issues.

Kami:
+1.

As far as the file name goes - `st2install.<timestamp>.log` or similar. I believe it should have `.log` suffix so it's consistent with other log files (usually you do find *.log and not find *.log*).

And I agree on including other additional useful information. In a sense, this looks like "installer version" of `st2-submit-debug-info`. There we include OS version, Python version, installed packages, etc.

PR #509, opened 2017-10-04
tee output to file, with timestamp on each line

warrenvw:
Resolves https://github.com/StackStorm/st2-packages/issues/499.

Remaining tasks:

 - [ ] ~~Add system info to the log, similar to that provided by `st2-submit-debug-info`.~~
 - [x] Log standard error.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -188,5 +188,6 @@ else
     echo "Running deployment script for st2 ${VERSION}..."
     echo "OS specific script cmd: bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${USERNAME} ${PASSWORD}"
     TS=$(date +%Y%m%dT%H%M%S)
-    bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${DEV_BUILD} ${USERNAME} ${PASSWORD} 2>&1 | adddate | sudo tee /var/log/st2-install.${TS}.log
+    sudo mkdir -p /var/log/st2
+    bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${DEV_BUILD} ${USERNAME} ${PASSWORD} 2>&1 | adddate | sudo tee /var/log/st2/st2-install.${TS}.log
 fi
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -18,6 +18,12 @@ PASSWORD=''
 BRANCH='v2.4'
 FORCE_BRANCH=""
 
+adddate() {
+    while IFS= read -r line; do
+        echo "$(date +%Y%m%dT%H%M%S%z) $line"
+    done
+}
+
 setup_args() {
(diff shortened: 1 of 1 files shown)


LindsayHill:
Concept looks good. Presumably need to do something similar for BWC install script?

warrenvw:
Please let me know if I'm headed in the right direction. This seemed the easiest way to add timestamps to each line in the `st2bootstrap-{deb|el6|el7}.sh` scripts. However, we're not logging output of `st2_bootstrap.sh`. As it exists right now, we're also not logging standard error.

LindsayHill:
Seems like stderr would be a kinda important thing to capture here

warrenvw:
Yep, it sure does. It's next on my list. :)

warrenvw:
Instead of writing version info to the install log, I'm beginning to think it makes more sense that we get folks to submit the log using `st2-submit-debug-info` (it ships all logs in `/var/log/st2`)? That seems to include all the required information and more. If the script does too much, maybe we could provide an option that ships just the install log file.

Or, we could do as mentioned before and put the required version info in the log itself. But then we're duplicating efforts.

Which of these two options makes the most sense?

scripts/st2_bootstrap.sh
```diff
@@ -181,5 +187,7 @@ else
 
     echo "Running deployment script for st2 ${VERSION}..."
     echo "OS specific script cmd: bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${USERNAME} ${PASSWORD}"
-    bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${DEV_BUILD} ${USERNAME} ${PASSWORD}
+    TS=$(date +%Y%m%dT%H%M%S)
+    sudo mkdir -p /var/log/st2
+    bash ${BOOTSTRAP_FILE} ${VERSION} ${RELEASE} ${REPO_TYPE} ${DEV_BUILD} ${USERNAME} ${PASSWORD} 2>&1 | adddate | sudo tee /var/log/st2/st2-install.${TS}.log
```
armab: Nice trick :+1:

armab: Just please verify if entire script hangs with the correct exit code if something is broken during the main installation.

warrenvw: Good catch. I'll submit a fix for that.

warrenvw: I've verified `exit ${PIPESTATUS[0]}` works as expected.

scripts/st2_bootstrap.sh
```diff
@@ -181,5 +187,8 @@ else
...
+    exit ${PIPESTATUS[0]}
```
armab: Awesome :+1:

warrenvw:
I verified that this works on el6, el7, trusty and xenial.

warrenvw:
I assume that we do not need to add system info to the log, since the log will be part of what's shipped by `st2-submit-debug-info`.

scripts/st2_bootstrap.sh
LindsayHill: I can't decide if it's good or bad that I'm learning new Bash tricks here. Probably good.

PR #510, opened 2017-10-10
Allow installer to work with Proxy

### LINKED ISSUE #500: Installation script should work with proxy
The installation scripts should support the use of a proxy server.

Currently this works if users create the files described [here](https://docs.stackstorm.com/packs.html#installing-packs-from-behind-a-proxy), prior to running the install script. This should be handled automatically by the installer script.

It may also need to support command-line flags to force the use of proxy, or to disable proxies.

armab:
Closes #500 as part of the https://github.com/StackStorm/discussions/issues/262 story

The expected interface to install st2 via proxy and then configure it is by passing `env` variables. An example: 
```sh
export http_proxy=http://192.168.10.123:3128
export https_proxy=http://192.168.10.123:3128
export no_proxy=127.0.0.1,localhost

# which is similar to passing the optional Hubot token
export HUBOT_SLACK_TOKEN=xoxb-abc-123

curl -sSL https://stackstorm.com/packages/install.sh |  bash -s -- --user=st2admin --password=st2admin --unstable
```

## Behavior
Keeping in mind that we're targeting idempotence (https://github.com/StackStorm/discussions/issues/262) in `curl|bash` scripts, the implemented logic honors the fact that installer could be re-ran many times.

- If proxy env vars are set - add them them in `st2api` and `st2actionrunner` service configs
- If passed proxy env vars are changed, - next `curl|bash` run will update the existing var values in service cofnigs
- If proxy env var(s) are unset/removed - `curl|bash` installer with remove the unset vars from service configs

## TODO
- [x] proxy support for community installer
  - [x] fix blocker: https://github.com/StackStorm/st2-packages/pull/511
- [x] st2docs https://github.com/StackStorm/st2docs/pull/633 https://github.com/StackStorm/st2docs/pull/635
- [x] test under the 4 platforms
  - [x] EL6
  - [x] EL7
  - [x] U14
  - [x] U16
- [ ] post-merge
  - [ ] manual tests
  - [ ] verify enterprise installer
  - [ ] verify suites installer

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -1,3 +1,11 @@
+function configure_proxy() {
+  # Allow bypassing 'proxy' env vars via sudo
+  local sudoers_proxy='Defaults env_keep += "http_proxy https_proxy no_proxy proxy_ca_bundle_path"'
+  if ! sudo grep -s -q ^"${sudoers_proxy}" /etc/sudoers.d/st2; then
+    echo "${sudoers_proxy}" | sudo tee -a /etc/sudoers.d/st2
+  fi
+}
+
 function get_package_url() {
   # Retrieve direct package URL for the provided dev build, subtype and package name regex.
   DEV_BUILD=$1 # Repo name and build number - <repo name>/<build_num> (e.g. st2/5646)
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -14,7 +14,7 @@ function configure_proxy() {
     sudo test -e ${service_config} || sudo touch ${service_config}
     for env_var in http_proxy https_proxy no_proxy proxy_ca_bundle_path; do
       # delete line from file if specific proxy env var is unset
-      if sudo test -z ${!env_var}; then
+      if sudo test -z "${!env_var:-}"; then
         sudo sed -i "/^${env_var}=/d" ${service_config}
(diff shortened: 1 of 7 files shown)


lakshmi-kannan:
@armab I cannot get pack install to work without
```
export proxy_ca_bundle_path=/etc/ssl/certs/mitmproxy-ca-cert.pem
```

This is because requests module requires a special flag to pass the CA bundle for verification. It doesn't look up the system certs. 

https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/packs.py#L80

I think you did not hit this because you're using squid and I am using mitmproxy which does SSL proxying differently - http://docs.mitmproxy.org/en/latest/certinstall.html.

Let me try the entire installation again with this flag set. I'll report back.

armab:
Added a note about the `proxy_ca_bundle_path` in installer docs: https://github.com/StackStorm/st2docs/pull/635

lakshmi-kannan:
@armab I report success with U16 and MITMProxy. I think the changes here are good to go and I'll approve. Next step should be to make the bwc-installer work under proxy. https://github.com/StackStorm/bwc-installer/tree/master/scripts 

@LindsayHill - What do you think about bwc-installer supporting proxy setups? Do you agree this is the next step?

LindsayHill:
Yes, bwc-installer needs to support proxies

PR #512, opened 2017-10-19
Ignore known processes using specified ports

warrenvw:
Look at the process that's LISTENING on the ports. If it's a known service, then pass. Otherwise, fail.

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -90,7 +90,7 @@ check_st2_host_dependencies() {
 
   for i in "${ports[@]}"
   do
-    rv=$(port_status $i | sed 's/.*-$\|.*st2.*\|.*nginx.*\|.*postgres.*\|.*mongo.*\|.*rabbitmq.*//')
+    rv=$(port_status $i | sed 's/.*-$\|.*systemd\|.*beam.smp.*\|.*epmd\|.*st2.*\|.*nginx.*\|.*python.*\|.*postgres\|.*postmaster.*\|.*mongod//')
     if [ "$rv" != "Unbound" ] && [ "$rv" != "" ]; then
       used+=("$rv")
     fi
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -90,7 +90,7 @@ check_st2_host_dependencies() {
 
   for i in "${ports[@]}"
   do
-    rv=$(port_status $i | sed 's/.*-$\|.*systemd\|.*beam.smp.*\|.*epmd\|.*st2.*\|.*nginx.*\|.*python.*\|.*postgres\|.*postmaster.*\|.*mongod//')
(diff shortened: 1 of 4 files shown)


armab:
The fix for current build failure is https://github.com/StackStorm/st2-packages/pull/513.
Needs to be merged to have it :white_check_mark:

armab:
Does it handle init processes under systemd U16 and EL7?

```
sudo netstat -tunlp --inet | awk -v port=:9100 '$4 ~ port && $6 ~ /LISTEN/ { print $4 " " $7 }'
127.0.0.1:9100 1/init
```

warrenvw:
I tested the functions on all platforms and behaviour is as expected.

scripts/includes/common.sh
```diff
@@ -90,7 +90,7 @@ check_st2_host_dependencies() {
 
   for i in "${ports[@]}"
   do
-    rv=$(port_status $i)
+    rv=$(port_status $i | sed 's/.*-$\|.*systemd\|.*beam.smp.*\|.*epmd\|.*st2.*\|.*nginx.*\|.*python.*\|.*postgres\|.*postmaster.*\|.*mongod//')
```
armab: Why the rule `.*-$` is needed?
Can you provide any example matching that regex?

warrenvw: I receive the following running

```
sudo netstat -tunlp --inet | awk -v port=:9100 '$4 ~ port && $6 ~ /LISTEN/ { print $4 " " $7 }'
```

on each of the platforms:

u14: `127.0.0.1:9100 15018/python`
u16: `127.0.0.1:9100 1/systemd`
el6: `127.0.0.1:9100 12283/python`
el7: `127.0.0.1:9100 1/systemd`
st2-docker u14: `127.0.0.1:9100 -`

warrenvw: I need the `.*-$` pattern when running in st2-docker. I suppose it can be simplified to `-$`.

armab:
This is from our AWS infra, Ubuntu16 AMI: 
```
arma@test001:~$ sudo netstat -tunlp --inet | awk -v port=:9100 '$4 ~ port && $6 ~ /LISTEN/ { print $4 " " $7 }'
127.0.0.1:9100 1/init
```

So merging output from our both environments, adding `1/init` along with `1/systemd` would be good.

scripts/includes/common.sh
armab: This is from our AWS infra, Ubuntu16 AMI: 
```
arma@test001:~$ sudo netstat -tunlp --inet | awk -v port=:9100 '$4 ~ port && $6 ~ /LISTEN/ { print $4 " " $7 }'
127.0.0.1:9100 1/init
```

So merging output from our both environments, adding `1/init` along with `1/systemd` would be good.

warrenvw:
@armab Thanks for the extra data point. 👍

PR #514, opened 2017-10-23
Logrotate mistral logs

### LINKED ISSUE #468: st2mistral package needs logrotate configuration file
The st2mistral package should ship with a logrotate configuration file, similar to https://github.com/StackStorm/st2/blob/master/conf/logrotate.conf

Otherwise the Mistral logs get out of hand, e.g:

```
brocade@BWC:/var/log/mistral$ du -hs *
8.0K    mistral-api.log
457M    mistral-server.log
brocade@BWC:/var/log/mistral$
```

This should have the same rotation policy as the st2 logrotate configuration. 

Important:
- file should not get deleted on RPM upgrade
- Ensure mistral gracefully handles re-opening log files.

armab:
Fixes #468 

Add logrotate configs for `st2mistral` deb/rpm package to prevent filling up the disk space by logs and overall to be a better packaging citizens.

### DIFF
--- a/packages/st2mistral/conf/logrotate.conf
+++ b/packages/st2mistral/conf/logrotate.conf
@@ -13,7 +13,7 @@
 
 ## Mistral Server
 /var/log/mistral/mistral-server.log {
-  size 10M
+  size 50M
   rotate 10
   compress
   notifempty
--- a/packages/st2mistral/conf/logrotate.conf
+++ b/packages/st2mistral/conf/logrotate.conf
@@ -7,7 +7,7 @@
   missingok
   create 644 mistral mistral
   postrotate
-    pkill -USR1 -f mistral.api.wsgi
+    pkill -USR1 -f mistral.api.wsgi > /dev/null 2>&1 || true
   endscript
 }
 
--- a/packages/st2mistral/conf/logrotate.conf
+++ b/packages/st2mistral/conf/logrotate.conf
@@ -1,0 +1,25 @@
+## Mistral API
+/var/log/mistral/mistral-api.log {
+  size 1M
+  rotate 5
+  compress
+  notifempty
+  missingok
+  create 644 mistral mistral
+  postrotate
+    pkill -USR1 -f mistral.api.wsgi
+  endscript
+}
+
+## Mistral Server
+/var/log/mistral/mistral-server.log {
+  size 10M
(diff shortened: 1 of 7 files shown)


packages/st2mistral/conf/logrotate.conf
```diff
@@ -0,0 +1,25 @@
...
+  size 10M
+  rotate 10
+  compress
+  notifempty
+  missingok
+  # Mistral doesn't handle USR1 signal right to re-open log files,
+  # Sending HUP will re-open the logs, but will stop the running workflows
+  copytruncate
+  delaycompress
+}
```
armab: Will run this logorate config for a several days on live `st2cicd` server to understand if it behaves OK.

packages/st2mistral/rpm/st2mistral.spec
```diff
@@ -59,6 +60,7 @@ Summary: st2 Mistral workflow service
   %{_bindir}/mistral
   /opt/stackstorm/mistral
   %config(noreplace) %{_sysconfdir}/mistral/*
+  %config(noreplace) %{_sysconfdir}/logrotate.d/mistral
```
armab: With this, `/etc/logrotate.d/mistral` file is deleted automatically on mistral remove what is good :+1: 

Checking the specifics on `deb` side.

packages/st2mistral/debian/conffiles
```diff
@@ -0,0 +1 @@
+/etc/logrotate.d/mistral
```
armab: For `deb` the expected behavior is different: `/etc/logroate.d/mistral` is not removed on `apt-get remove st2mistral`, but it's removed on `apt-get purge st2mistral`, which removes the package with configs.

packages/st2mistral/conf/logrotate.conf
```diff
@@ -0,0 +1,25 @@
...
+  create 644 mistral mistral
+  postrotate
+    pkill -USR1 -f mistral.api.wsgi > /dev/null 2>&1 || true
+  endscript
+}
+
+## Mistral Server
+/var/log/mistral/mistral-server.log {
+  size 50M
+  rotate 10
```
armab: Only these actual logrotate settings are under question.
If it's good enough and balanced or we need different numbers here.

PR #522, opened 2017-11-18
v2.5.1: Pin ruby version to 2.0.0 for package_cloud

humblearner:


armab:
Already included in https://github.com/StackStorm/st2-packages/pull/526

PR #533, opened 2018-02-20
Add netbase as a dependency for st2 deb package

Kami:
Some minimal container images (notably our xenial packagingtest image) don't contain this package and because of that, our tests fail.

We decided best way to ensure this package is present is to have netbase as a dependency of st2 package.

The issue only affects our xenial packagingtest build and image, but it's still a good idea to define it there.

For context and details, see https://github.com/StackStorm/st2/pull/4007#issuecomment-366909531.


Kami:
Looks like this file is provided by netcfg package on RHEL based distros.

Kami:
Added same change for RPM package - 547c242.

Good call, @armab.

PR #537, opened 2018-02-27
Add notifier component to the mistral service files

m4dcoder:
Add the notifier component to the mistral service files. The notifier component is required for the mistral to st2 callback integration.


m4dcoder:
I manually tested the service files.

PR #538, opened 2018-03-28
Fix nginx to install from 'nginx' repo on EL7, instead of 'EPEL'

armab:
Originally we install `nginx` from the `nxing.org` official repo which has `1:1.12.2-1` version.
A few days ago EPEL released new version of nginx `1:1.12.2-2.el7` which now takes precedence over the nginx.org version and gets installed instead.

EPEL version of nginx comes with other dependencies, different directory structure, nginx config which resulted in a broken EL7 installation.

```
20180328T173450+0000 ############### ERROR ###############
20180328T173450+0000 # Failed on Install st2web #
20180328T173450+0000 #####################################
```

> Related Ansible-st2 fix: https://github.com/StackStorm/ansible-st2/pull/191

### DIFF
--- a/scripts/st2bootstrap-el6.sh
+++ b/scripts/st2bootstrap-el6.sh
@@ -673,8 +673,10 @@ gpgcheck=1
 enabled=1
 EOT"
 
-  # Install st2web and nginx
-  sudo yum install -y ${ST2WEB_PKG} nginx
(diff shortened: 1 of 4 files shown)


armab:
Confirming the fix by verifying the scripted StackStorm installer.

PR #543, opened 2018-04-03
Fix CI build failure, move to CircleCI2.0

### LINKED ISSUE #22: Docker Containers Caching in CircleCI?
The pull stage in CircleCI when Docker containers dependencies are downloaded takes up to `2 mins` (usually 1.5).

Workaround to cache Docker Containers for CircleCI described here:
https://circleci.com/docs/docker#caching-docker-layers
https://circleci.com/docs/docker#some-known-caching-issues

Looks like dirty hack, probably requires crazy debugging to make it working for `docker-compose`.

### LINKED ISSUE #31: CircleCI systemd (centos7/xenial) container is not operational

### LINKED ISSUE #38: [INFO] Moar Parallel Docker Operations
Opening this issue for informational purposes (not to forget).

2 docker-related tickets to track, that would be useful for us in future to further improve [CI pipeline](https://circleci.com/gh/StackStorm/st2-packages) (speed improvements where possible):
- [Concurrent "docker build" takes longer than sequential builds](https://github.com/docker/compose/issues/1652)
- [Epic: Improve performance by running operations in parallel](https://github.com/docker/docker/issues/9656)

What we alredy do:
- Push images to Docker Hub in parallel

### LINKED ISSUE #536: Move build to CircleCI 2.0
We've moved to `CircleCI 2.0` for `StackStorm/st2` repository a long time ago.

Logically, packages build in `st2` and `st2-packages` repositories do the same thing.

The bigger difference is in mechanics: `st2-packages` is on `CircleCI 1.0`, while `st2` is on `CircleCI 2.0`. Considering that `v1.0` gradually abandoned, we see more and more diff in build reliability & results.

----

Move `st2-packages` to CircleCI 2.0, following the `st2` repo example:
https://github.com/StackStorm/st2/blob/master/circle.yml#L108-L248
As a bonus, switching to new CircleCI infra will bring faster build times.

### LINKED ISSUE #542: Fatal: destination path '/tmp/mistral-...' already exists and is not an empty directory
We started seeing this Rake error (race condition case) in `st2-packages` CircleCI build: 

```
[56a78cea][DEBUG]  Running /usr/bin/env nproc as root@centos6build
[56a78cea][DEBUG]      2
[56a78cea][DEBUG]  Finished in 0.160 seconds with exit status 0 (successful).
[checkout: st2mistral]  Running /usr/bin/env mkdir -p $ARTIFACT_DIR as root@centos6build
[checkout: st2]  Running /usr/bin/env mkdir -p $ARTIFACT_DIR as root@centos6build
[checkout: st2]  Finished in 0.015 seconds with exit status 0 (successful).
[checkout: st2mistral]  Finished in 0.021 seconds with exit status 0 (successful).
[checkout: st2]  Running /usr/bin/env git clone --depth 1 -b $GITREV $GITURL $GITDIR as root@centos6build
[checkout: st2mistral]  Running /usr/bin/env git clone --depth 1 -b $GITREV $GITURL $GITDIR as root@centos6build
[checkout: st2]      Initialized empty Git repository in /tmp/mistral-20180402-9-gunys9/.git/
[checkout: st2mistral]      fatal: destination path '/tmp/mistral-20180402-9-gunys9' already exists and is not an empty directory.

.circle/docker-compose.sh build ${DISTRO} returned exit code 1

Action failed: .circle/docker-compose.sh build ${DISTRO}
```

Example: https://circleci.com/gh/StackStorm/st2-packages/2901#tests/containers/2
Happens only on EL6/EL7 Docker guests.

armab:
An effort to fix [`Fatal: destination path '/tmp/mistral-...' already exists and is not an empty directory #542`](https://github.com/StackStorm/st2-packages/issues/542) CI build failure appeared a while ago leaded directly to liquidating technical debt of long delayed CircleCI 1.0 -> CircleCI 2.0 migration.

### One PR to fix them all: 
- Fixes #542 - original CI build failure which is blocking an st2 release.
- Closes #536, - we migrate to CircleCI 2.0. Outdated 1.0 environment is the actual reason of this failure (see #536 for more). Also Circle 1.0 is going to be deprecated this summer, so killing two :rabbit: in one shot.
- Fixes #31 so the integration tests now can work in systemd-based Ubuntu Xenial and CentOS7. They didn't work because of Docker + systemd limitation in CircleCI 1.0.
- Fixes #22, - Docker container caching is a new feature of CircleCI2.0 and we benefit from it from now on.
- Closes #38, - with CircleCI 1.0 we were locked to max `docker v1.9.0` which is outdated as hell. With CircleCI2.0 we can use latest Docker engine (currently `18.03`) and benefit from more parallel docker-compose operations (pull, build), better reliability and build stability.
- Closes #408, - first effort of migrating to 2.0


LindsayHill:
I thought that CircleCI 2.0 preferred to have config file at `/.circleci/config.yml`, rather than `/circle.yml`?

warrenvw:
Yes, it's true.

> On CircleCI 2.0, this file must be called config.yml and must be in a hidden folder called .circleci

https://circleci.com/docs/2.0/#adding-a-yml-file

armab:
@LindsayHill We rely on old `circle.yml` in our `st2ci/cd` build workflows to perform version-related changes in stable `vX.Y` branches.
https://github.com/StackStorm/st2-packages/blob/51cd103f6ddd606faf07f693747ebca0e02c2814/circle.yml#L23-L28

So leaving it as is. I don't feel there is a big diff if we place it here or there.

LindsayHill:
Just don't want us to get caught out in future if CircleCI stops supporting that file location. Wonder if it's a transition/legacy thing from them?

armab:
We'll see what happens. If they deprecate old location, - we'll have to change the `st2` repo `circle.yml` as well in one go, which relies on `2.0` schema too.

As part of work to "fix the st2 builds asap", I'm not convinced to update the file paths which also force changing the `st2ci/cd` workflows (and handling corner cases for the new way in `master` vs old way in `vX.Y` branches leading to another rabbit hole).

Kami:
Thanks for working on this.

And yeah, I would personally also prefer to avoid changing the location right now, if possible, so we don't need to update ci/cd workflows.

circle.yml
```diff
@@ -3,104 +3,165 @@
...
+    environment:
+      DISTROS: "trusty xenial el6 el7"
+      ST2_PACKAGES: "st2 st2mistral"
+      BASH_ENV: ~/.buildenv
+      # These should be set to an empty string, so that st2cd prep tasks are able to replace these
+      # with real gitrevs during releases. Note that they are commented out, so that they do not interfere
+      # with build parameters. st2cd prep tasks will uncomment these on a branch, and replace with proper
+      # gitrefs.
+      # ST2_GITREV: ""
+      # ST2MISTRAL_GITREV: ""
```
armab: Branch needs version pinning, see
https://github.com/StackStorm/st2-packages/blob/631e8a47eefc8191089d3723e223b3f199e9d44e/circle.yml#L23-L28

PR #545, opened 2018-04-06
Add back Circle CI 1.0 related files still used by mistral repo

Kami:
Those files are still used by the mistral repo (https://github.com/StackStorm/mistral/blob/master/circle.yml) which hasn't been ported to Circle CI 2.0 yet.

This is a temporary solution to try to get the build to pass and the packages out. Eventually we should also move that repo to Circle CI 2.0.


armab:
:+1:

PR #549, opened 2018-04-17
Fix dh-virtualenv packages build, avoid updating pip to latest v10.0

armab:
> Related to: https://github.com/StackStorm/st2-dockerfiles/pull/58 https://github.com/StackStorm/st2-dockerfiles/pull/59 https://github.com/StackStorm/st2-dockerfiles/pull/60

pip recently released new `v10.0` which broke the `dh-virtualenv` packages build.

So here is the rabbit hole.
While we use & [pin `virtualenv 15.1.0`](https://github.com/StackStorm/st2/blob/e8f179c0023bd1a11b332126cfe493c197284486/fixed-requirements.txt#L34-L35) which advertizes to install `pip 9.0.1` (https://virtualenv.pypa.io/en/stable/changes/#id1) it doesn't happen in reality.

Obviously, at it happens in software world, it's not a bug but a feature, - `virtualenv` since `14.0.1` installs _latest available_ (!) pip version by default when user creates a new virtualenv (see https://github.com/pypa/virtualenv/issues/1044 and https://github.com/pypa/virtualenv/issues/1157).

To avoid this behavior in virtualenv and stick with expected associated pip version, `--no-download` flag was added (see https://virtualenv.pypa.io/en/stable/changes/#id12):
> Download new releases of the preinstalled software from PyPI when there are new releases available. This behavior can be disabled using --no-download.

Well, it means that once we updated to newer virtualenv version (14.0.1+) which installs latest pip by default, we opened a can of worms.

### TODO
- [x] Find the root cause
- [x] Provide `--no-download` flag to virtualenv in st2-packages
   - [x] Investigate new fails: we rely on [our outdated fork](https://github.com/StackStorm/dh-virtualenv/tree/fix/py-custom-shebang) of `dh-virtualenv` which doesn't have [`--extra-virtualenv-arg`](https://github.com/spotify/dh-virtualenv/pull/146)
     - [x] Update our `dh-virtualenv` fork, re-implement custom changes https://github.com/StackStorm/dh-virtualenv/commit/bf43b0ae436c8caa67f24d545df8cfca70d3cb9a (thanks @Kami https://github.com/StackStorm/st2-dockerfiles/pull/61)
       - [x] `trusty` build started to fail due to updated `dh-virtualenv` fork changes, fix: https://github.com/StackStorm/dh-virtualenv/commit/bab5d2ef8c9402c3d203bfb3bf757b39a22e1f1d
- [x] Check if `st2` repo needs `--no-download` as well for `st2 pack install` (thanks @Kami https://github.com/StackStorm/st2/pull/4085)

### DIFF
--- a/.circle/docker-compose2.sh
+++ b/.circle/docker-compose2.sh
@@ -23,7 +23,7 @@ case "$1" in
   # Perform fake command invocation, technically provides images "pull" phase.
   pull)
     echo Pulling dependent Docker images for $2 ...
-    docker-compose -f docker-compose.circle2.yml -f docker-compose.override.yml run $2 /bin/true
+    docker-compose -f docker-compose.circle2.yml -f docker-compose.override.yml pull --include-deps $2
   ;;
   build)
     echo Starting Packages Build for $2 ...
--- a/packages/st2/debian/rules
+++ b/packages/st2/debian/rules
@@ -56,9 +56,13 @@ override_dh_installdeb:
 	dh_installdeb
 
 override_dh_virtualenv:
-	# NP! Wheels must be pre-populated by now, we use --no-index to skip
+	# NB! Wheels must be pre-populated by now, we use --no-index to skip
 	# querying pypi and rely only on --find-links.
-	dh_virtualenv --extra-pip-arg '--find-links=$(WHEELDIR)' \
+	#
+	# NB! Use '--no-download' arg for 'virtualenv' is required,
+	# otherwise it downloads latest PIP version instead of bundled/pinned one.
+	dh_virtualenv --extra-virtualenv-arg='--no-download' \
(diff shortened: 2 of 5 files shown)


Kami:
As far as `st2 pack install` goes - I confirmed it results in the same behavior since we also use latest version of virtualenv there.

So while it does install latest version of pip (10.0.0) it doesn't seem to break packs. In theory it shouldn't, because (in ideal world) packs should never import from pip or do anything like that.

Having said that, I still think it's probably safer and better (and results in more reproducible virtualenv creations) to pass `--no-download` flag to `virtualenv` binary while creating a pack virtual environment.

Kami:
Nice work - thanks!

I will merge two st2 PRs into master, test them on staging unstable on all the distros (really just "st2 pack install" needs to be tested) and start working on 2.7.1 when everything is in order :)

PR #553, opened 2018-04-20
Fix st2mistral smoke tests not running

### LINKED ISSUE #529: Smoke tests for mistral not running
https://github.com/StackStorm/st2-packages/pull/527 fixed a postgres upstream issue, and sometime before that PR was merged and the previous packages were generated, one of mistral's upstream dependencies made a change that was causing mistral to not start. This was fixed in https://github.com/StackStorm/st2-packages/pull/528.

The question now is, why did the tests pass for #527? There are smoke tests that are executed as part of https://github.com/StackStorm/st2-packages/blob/master/rake/spec/default/60-st2_all-services-ok_spec.rb#L88-L93, but for some reason those weren't run - guessing due to a misconfiguration of `mistral_enabled `. Need to look into this.

armab:
For quite a long time our `st2mistral` smoke tests were not running.

Fixes #529

### DIFF
--- a/rake/spec/default/10-package_st2-consistency_spec.rb
+++ b/rake/spec/default/10-package_st2-consistency_spec.rb
@@ -13,10 +13,14 @@ shared_examples 'os package' do |name, _opts|
 
   # Check for presence of users
   #
-  get_users do |u, opts|
-    describe user(u) do
-      it { is_expected.to exist }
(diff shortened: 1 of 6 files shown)


armab:
A bunch of tests were not running, including some for `st2` like directories and users expected to exist.
All mistral tests were absent.
Surprising.

PR #555, opened 2018-05-09
Increasing timeout for service startup on CentOS 6

nmaludy:
Closes StackStorm/st2/issues/4100

On CentOS/RHEL 6 in `2.7.0` we experienced the `st2auth` service failing to start due to a timeout waiting for the process to initialize. 

This change increases the process startup timeout on CentOS/RHEL 6 to allow for longer startup times from the `st2auth` service.

### DIFF
--- a/packages/st2/conf/rhel-functions-sysvinit
+++ b/packages/st2/conf/rhel-functions-sysvinit
@@ -124,7 +124,7 @@ daemon() {
 # By default wait for the pid up to 3 seconds.
 pgrep_waitforpid() {
   local pid= pid_file= base= trailer=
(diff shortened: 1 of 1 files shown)


packages/st2/conf/rhel-functions-sysvinit
```diff
@@ -124,7 +124,7 @@ daemon() {
 # By default wait for the pid up to 3 seconds.
```
armab: Need to change a comment here `3 seconds` as well

nmaludy:
Comment is fixed :+1:

Kami:
LGTM :+1: 

Thanks @nmaludy and @armab.

PR #561, opened 2018-06-01
Fix sharing st2 logs when CircleCI job failed

armab:
See https://github.com/StackStorm/st2/pull/4158 for more context


armab:
The CI failures were caused by updated Docker images (pulling new Serverspec) via https://github.com/StackStorm/st2-dockerfiles/pull/65 that automatically were built/pushed to Docker Hub and fixed in https://github.com/StackStorm/st2-dockerfiles/pull/66

PR #563, opened 2018-06-14
Fix configuration directory and file permissions

blag:
Companion PR to [st2#4173](https://github.com/StackStorm/st2/pull/4173).

Fixes configuration directory permissions to `2770` and the configuration file permissions to `660`.


scripts/includes/common.sh
```diff
@@ -201,11 +199,9 @@ EOT"
     sudo mkdir -p ${CURRENT_USER_CLI_CONFIG_DIRECTORY}
   fi
 
-  sudo sh -c "cat <<EOT > ${CURRENT_USER_CLI_CONFIG_PATH}
-[credentials]
-username = ${USERNAME}
-password = ${PASSWORD}
-EOT"
+  sudo sh -c st2 login --config ${CURRENT_USER_CLI_CONFIG_PATH} \
```
armab: Any reason why using `sudo` and `sh` is required? Ideally is to avoid something like that, making sure that `st2` does the right thing.

Just `st2 login` would be enough.

scripts/includes/common.sh
```diff
@@ -185,11 +185,9 @@ configure_st2_cli_config() {
     sudo mkdir -p ${ROOT_USER_CLI_CONFIG_DIRECTORY}
   fi
 
-  sudo sh -c "cat <<EOT > ${ROOT_USER_CLI_CONFIG_PATH}
-[credentials]
-username = ${USERNAME}
-password = ${PASSWORD}
-EOT"
+  sudo sh -c st2 login --config ${ROOT_USER_CLI_CONFIG_PATH} \
```
armab: Any reason why do we need explicit `ROOT_USER_CLI_CONFIG_PATH`?

With that, the logic defined in `configure_st2_cli_config()` needs a cleanup. All dir/config file creation should be a `st2 login -w` responsibility.

blag: > Do we really need explicit `ROOT_USER_CLI_CONFIG_PATH`?

Technically, no, but it makes it easier to verify that the two calls to `st2 login` (one for root, one for the user) are doing similar things.

I have found it easier to make more variables in Bash than to hard code everything.

> All dir/config file creation should be a `st2 login -w` responsibility.

Yep, fixed.

armab: Thanks for explaining that.

To ensure st2 login config is created for both `user` and `root`, would it be enough to run
1) `st2 login -w` (for user)
2) `sudo st2 login -w` (for root)

Can st2 login handle that correctly?
---
Additionally, there is no such `--config` option for `st2 login` command.
We shouldn't hardcode the dirs in every installer. It's a st2 responsibility to decide on config dirs.

scripts/includes/common.sh
armab: Alright, as described in another comment, 1 run `sudo st2 login -w` is needed to write config for root and another `st2 login -w` is needed to write user's config.

scripts/includes/common.sh
blag: > Can st2 login handle that correctly?

Should be able to, yes.

> Additionally, there is no such --config option for st2 login command.

Good catch, thanks! Fixed.

> We shouldn't hardcode the dirs in every installer. It's a st2 responsibility to decide on config dirs.

That's true, but I'm just trying to maintain backwards compatibility. We can take care of that in a different PR.

scripts/st2bootstrap-deb.sh
```diff
@@ -298,34 +298,22 @@ configure_st2_cli_config() {
...
-  if [ ! -d ${CURRENT_USER_CLI_CONFIG_DIRECTORY} ]; then
-    sudo mkdir -p ${CURRENT_USER_CLI_CONFIG_DIRECTORY}
-  fi
-
-  sudo sh -c "cat <<EOT > ${CURRENT_USER_CLI_CONFIG_PATH}
-[credentials]
-username = ${USERNAME}
-password = ${PASSWORD}
-EOT"
+  st2 login --config-file ${CURRENT_USER_CLI_CONFIG_PATH} \
```
armab: As said in previous comment, there is no such flag for `st2 login` as `--config-file`

scripts/st2bootstrap-deb.sh
```diff
@@ -298,34 +298,22 @@ configure_st2_cli_config() {
...
+            --write-password \
+            --username ${USERNAME} --password ${PASSWORD}
 
   # Fix the permissions
-  sudo chown -R ${CURRENT_USER}:${CURRENT_USER} ${CURRENT_USER_CLI_CONFIG_DIRECTORY}
+  chown -R ${CURRENT_USER}:${CURRENT_USER} ${CURRENT_USER_CLI_CONFIG_DIRECTORY}
```
armab: Don't need to `chown` anymore.

If we do care about backwards compatibility, need to keep in mind that `curl|bash` installer is not idempotent and runs only once (first install), and so the config files/dirs are created on first run making `chown` excessive in this case.

Also, that's why we moved this permissions logic in `st2 login` itself, see: https://github.com/StackStorm/st2/pull/4173

blag:
@armab I fixed the issues you pointed out. Let me know if there's anything else I need to fix.

armab:
Thanks @blag! The logic LGTM.
The only requirement is to test the installer script end-to-end once we'll have the st2 https://github.com/StackStorm/st2/pull/4173 merged.

blag:
@armab Waiting on your approval to merge this.

armab:
@blag Per previous discussion, did you end-to-end test this change manually?
We need to ensure that this change doesn't break anything prior merging.

Kami:
@armab @blag We merged st2 change, but not this one so now if you install StackStorm using installer script it will produce warnings by default.

I would say merge that if it's working so we get rid of warnings on a clean installation. Everything needs to be tested end to end though, of course.

Kami:
I will merge it and test it out.

If it doesn't work, I will revert it since we don't want to delay the release any longer.

Kami:
I pushed a fix, yet things are still breaking (38fdd4e6eafdcf6c1417178021bc18fb7af12232). This requires more work.

@blag for future reference, that's how you tested installer script changes from a branch:

```bash
wget https://stackstorm.com/packages/install.sh
bash install.sh --user=st2admin --password=Ch@ngeMe --staging --unstabl --force-branch=fix-config-permissions
```

The problem seems to be that ``st2 login`` command depends on the config directory and config file to exist.

This means we will probably still need to manually create and chmod directories and config inside the installer script.

armab:
> Either that, or change st2 CLI to create directories and file if they don't exist yet, but that's a lot more involved.

Agree, that's the point which I'd like us to go with.
But it's easier, because `st2 login` can already create files and dirs if they don't exist yet.

We just need to get rid of all the config/dir paths in scripted installer and let the st2 handle it correctly. This leaves scripted installer with less logic to handle which is a good part.

armab:
I've made a commit to this PR to have a working st2 config behavior for scripted installer.

```
curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password=st2admin --staging --unstable --force-branch=fix-config-permissions

[...]

$ ls -la ~/.st2/
total 16
drwxrws--- 2 vagrant vagrant 4096 Jun 29 16:09 .
drwxr-xr-x 5 vagrant vagrant 4096 Jun 29 16:11 ..
-rw-rw---- 1 vagrant vagrant   55 Jun 29 16:09 config
-rw-rw---- 1 vagrant vagrant   77 Jun 29 16:09 token-st2admin

$ sudo ls -la /root/.st2/
total 16
drwxrws--- 2 root root 4096 Jun 29 16:09 .
drwx------ 5 root root 4096 Jun 29 16:10 ..
-rw-rw---- 1 root root   55 Jun 29 16:09 config
-rw-rw---- 1 root root   77 Jun 29 16:09 token-st2admin
```


@Kami Please take a look again.
@blag Please help to test it end-to-end, @Kami provided instructions above.

PR #566, opened 2018-06-21
Add st2workflowengine service

m4dcoder:
Add systemd, upstart, and related service files for st2workflowengine.

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -49,6 +49,7 @@ post_install:
 	install -m644 ../st2actions/conf/logging.conf $(DESTDIR)/etc/st2/logging.actionrunner.conf
 	install -m644 ../st2actions/conf/syslog.conf $(DESTDIR)/etc/st2/syslog.actionrunner.conf
 	install -m644 ../st2actions/conf/logging.workflowengine.conf $(DESTDIR)/etc/st2/logging.workflowengine.conf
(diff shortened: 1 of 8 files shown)


packages/st2/debian/rules
```diff
@@ -37,6 +37,7 @@ ifeq (true, $(IS_SYSTEMD))
 	dh_systemd_enable --name=st2rulesengine st2rulesengine.service
 	dh_systemd_enable --name=st2sensorcontainer st2sensorcontainer.service
 	dh_systemd_enable --name=st2garbagecollector st2garbagecollector.service
+    dh_systemd_enable --name=st2workflowengine st2workflowengine.service
```
armab: Use `tabs` instead of spaces in `Makefile` for indentation as that's the only supported way by their syntax.

packages/st2/Makefile
```diff
@@ -48,6 +48,7 @@ post_install:
 	install -m644 ../conf/logrotate.conf $(DESTDIR)/etc/logrotate.d/st2
 	install -m644 ../st2actions/conf/logging.conf $(DESTDIR)/etc/st2/logging.actionrunner.conf
 	install -m644 ../st2actions/conf/syslog.conf $(DESTDIR)/etc/st2/syslog.actionrunner.conf
+	install -m644 ../st2actions/conf/logging.workflowengine.conf $(DESTDIR)/etc/st2/logging.workflowengine.conf
```
armab: Apart of `logging.workflowengine.conf`, need to copy also `syslog.workflowengine.conf`, similar to other services.

PR #568, opened 2018-06-27
Install and use special gunicorn logging config file for all gunicorn API services

Kami:
This pull request installs gunicorn worker logging config files and updates service manager files to pass ``--logging-config`` argument to the gunicorn process.

For context and details, see https://github.com/StackStorm/st2/pull/4206.

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -51,10 +51,13 @@ post_install:
 	install -m644 ../st2actions/conf/logging.workflowengine.conf $(DESTDIR)/etc/st2/logging.workflowengine.conf
 	install -m644 ../st2actions/conf/syslog.workflowengine.conf $(DESTDIR)/etc/st2/syslog.workflowengine.conf
 	install -m644 ../st2api/conf/logging.conf $(DESTDIR)/etc/st2/logging.api.conf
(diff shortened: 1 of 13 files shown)


armab:
How `st2api/conf/logging.gunicorn.conf` is different from `st2api/conf/logging.conf`
when one is used and when another? I'm curious if st2.conf logging settings we set make any sense.
Which part is logged via gunicorn config and which part via old logging config?

Is it possible to have one logging config for services like api stream auth, instead of two configs?

At this moment sounds like to change the service to log to `stdout` we'll need to edit 2 files?

Kami:
Responded in Slack, to recap:

We can't re-use logging config file from a st2 service directly because StackStorm services inject new `AUDIT` log level which is not known to gunicorn workers so it would break if we used st2 logging config file directly for gunicorn (in fact, that's was my first attempt - trying to re-use the same configs for gunicorn).

We could perhaps try to do some deduplication and move some of duplicated stuff to build time. This would perhaps decrease duplication a bit, but it would make it more opaque (harder to see where things are set and changed).

Kami:
Note: Circle CI build will keep failing until st2 changes are merged into master.

PR #569, opened 2018-06-27
Adding orchestra runner to requirements

bigmstone:


packages/st2/debian/postinst
```diff
@@ -50,6 +50,11 @@ register_runners() {
     st2ctl reload --register-runners --register-fail-on-failure
 }
 
+install_orchestra_runner() {
+    echo "Installing orchestra runner"
+    (cd /opt/stackstorm/runners/orchestra_runner; /opt/stackstorm/virtualenv/python setup.py install)
```
Kami: Hm, shouldn't this path be `/opt/stackstorm/st2/bin/python` and not `/opt/stackstorm/virtualenv/python`?

Kami: Also, I wonder how this "double installation" will affect things (well, in theory we don't really double install them now, we just mess with the PYTHONPATH, but end result is similar).

PR #572, opened 2018-06-28
Adding Orchestra Runner To Build

bigmstone:
Building runner wheels and adding orchestra runner to install requirements

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -3,6 +3,7 @@ ST2_COMPONENT := $(notdir $(CURDIR))
 ST2PKG_RELEASE ?= 1
 ST2PKG_VERSION ?= $(shell python -c "from $(ST2_COMPONENT) import __version__; print __version__,")
 CHANGELOG_COMMENT ?= "automated build, version: $(ST2PKG_VERSION)"
+RUNNERS := $(shell ls ../contrib/runners)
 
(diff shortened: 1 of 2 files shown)


packages/st2/Makefile
```diff
@@ -84,13 +85,24 @@ ifeq ($(DEBIAN),1)
...
 
-wheelhouse: .stamp-wheelhouse
+wheelhouse: .build-runners .stamp-wheelhouse
 .stamp-wheelhouse: | populate_version requirements inject-deps
 	cat requirements.txt
 	pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt || \
 		pip wheel --wheel-dir=$(WHEELDIR) --find-links=$(WHEELDIR) -r requirements.txt
 	touch $@
 
+.build-runners:
```
Kami: Thanks, LGTM :+1:

packages/st2/in-requirements.txt
```diff
@@ -7,4 +7,5 @@ st2reactor
 st2exporter
 st2debug
 st2tests
+stackstorm-runner-orchestra
```
Kami: Eventually, when st2 code changes are in place, we should do the same with all the runners.

We can also do it dynamically inside `inject-deps` Make target, similar as we do for ``st2client``.

armab:
Awesome!

Thanks for digging in and going this path :+1:

PR #574, opened 2018-07-06
Make sure we restart st2sensorcontainer and st2workflowengine service after generating encryption key (v2.8)

Kami:
I noticed this bug / issue while testing v2.8.0 packages.

We generate symmetric encryption key for encrypted datastore values after StackStorm has been installed and all the services started.

This means that if we don't restart all the affected services (st2api, st2workflowengine, st2sensorcontainer), those services won't be aware that the key exists until they have been restarted. As such, encryption / decryption will fail.

For example (from `st2workflowengine` log file):

```bash
2018-07-06 11:08:17,633 140130336387472 INFO keyvalue [-] Checking if encryption is enabled for key-value store.
2018-07-06 11:08:17,633 140130336387472 INFO keyvalue [-] Encryption enabled. Looking for key in path 
2018-07-06 11:08:17,633 140130336387472 ERROR keyvalue [-] Encryption key file does not exist in path .
```

I originally noticed this issue in `st2workflowengine`, but the same issue exists with `st2sensorcontainer` which spawns long running sensor processes on start up.

Keep in mind that the same issue doesn't affect `st2actionrunner`, because decrypting / encrypting values is only supported by Python runner actions and those actions run as subprocesses which are spawned when action execution is requested.

Having said that, I would still recommend to restart it as well - mostly for consistency reasons and to avoid simialr issues in the future if we ever change how st2actionrunner works or add support for encrypting / decrypting to other runner actions.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -350,7 +350,7 @@ generate_symmetric_crypto_key_for_datastore() {
   # set path to the key file in the config
   sudo crudini --set /etc/st2/st2.conf keyvalue encryption_key_path ${DATASTORE_ENCRYPTION_KEY_PATH}
 
-  # NOTE: We need to restart all the affected services so they pick the key nad load it in memory
+  # NOTE: We need to restart all the affected services so they pick the key and load it in memory
   sudo st2ctl restart-component st2api
   sudo st2ctl restart-component st2sensorcontainer
   sudo st2ctl restart-component st2workflowengine
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -351,6 +351,8 @@ generate_symmetric_crypto_key_for_datastore() {
   sudo crudini --set /etc/st2/st2.conf keyvalue encryption_key_path ${DATASTORE_ENCRYPTION_KEY_PATH}
 
   sudo st2ctl restart-component st2api
+  sudo st2ctl restart-component st2sensorcontainer
+  sudo st2ctl restart-component st2workflowengine
 }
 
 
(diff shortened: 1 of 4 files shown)


Kami:
As per discussions with @enykeev, we decided it's safer and more future proof to simply restart all the services (e.g. if we do something with encrypting values in other services in the future, this would bite us again).

Restarting all the services does slow everything up for 5 seconds or so, mostly because Mistral, but I don't think it's a big deal.

scripts/st2bootstrap-deb.sh
```diff
@@ -350,7 +350,11 @@ generate_symmetric_crypto_key_for_datastore() {
   # set path to the key file in the config
   sudo crudini --set /etc/st2/st2.conf keyvalue encryption_key_path ${DATASTORE_ENCRYPTION_KEY_PATH}
 
-  sudo st2ctl restart-component st2api
+  # NOTE: We need to restart all the affected services so they pick the key and load it in memory
+  # NOTE: Only st2api, st2sensorcontainer, st2workflowengine and st2actionrunner work with
+  # encrypted datastore values, but we restart all the services to be on the safe side and for
+  # future proofing reasons.
+  sudo st2ctl restart
```
armab: I'd like to still avoid restarting everything

Also see for example where we follow this granular strategy, instead of restart-all:
https://github.com/StackStorm/st2-packages/blob/9249e5cdff0d6d8ead2d6921e7bc5028e35c9c61/scripts/st2bootstrap-deb.sh#L579-L585

Kami: @armab Fine, I'll revert it to restarting just the affected services for now. It just means we need to be careful in the future or it will bite us again.

I honestly don't think 10-15s slow down is a big deal, but fine.

enykeev: Changing config without reloading it in every component is very dangerous. None of us could be able to tell from the top of our head whether any particular line of config used in one component or a several of them.

enykeev: Also, when adding new component or a feature to existing component you don't usually think about all the cases this component needs to be restarted during install.

armab: Then we have to keep in mind about this in future,
because in Docker we're giving granular access to specific services that _really_ need that access, not "free-for-all".

Ex: pack content and virtualenvs only for `st2actionrunner`, `st2sensorcontainers` and for register-content one-time task, (now `st2workflowengine` may need it as well ?).

Have to figure out the borders for each service.

Issue #576, opened 2018-07-08
Install script attempts to restart non-existent st2workflowengine component


tcosprojects:
OS: Ubuntu 16.04.4 LTS
StackStorm Version: st2 2.7.2, on Python 2.7
Installed with: install.sh, st2bootstrap-deb.sh

Using the quick install instructions, the following error occurs:

```
20180708T114925+0000 Error: Invalid component provided: st2workflowengine
20180708T114925+0000
20180708T114925+0000 Usage: st2ctl {start, stop, restart, status}
20180708T114925+0000
20180708T114925+0000 Usage: st2ctl {restart-component}
20180708T114925+0000 positional arguments:
20180708T114925+0000   component           Name of the st2 service to restart.
20180708T114925+0000                       st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops mistral
20180708T114925+0000
20180708T114925+0000 Usage: st2ctl {reopen-log-files}
20180708T114925+0000 positional arguments:
20180708T114925+0000   component           Name of the st2 service to reopen the log files for.
20180708T114925+0000                       st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops
20180708T114925+0000
20180708T114925+0000 Usage: st2ctl {reload, clean}
20180708T114925+0000 optional arguments:
20180708T114925+0000   --register-all                Register all.
20180708T114925+0000   --register-triggers           Register all triggers.
20180708T114925+0000   --register-sensors            Register all sensors.
20180708T114925+0000   --register-rules              Register all rules.
20180708T114925+0000   --register-runners            Register all runners.
20180708T114925+0000   --register-actions            Register all actions.
20180708T114925+0000   --register-aliases            Register all aliases.
20180708T114925+0000   --register-policies           Register all policies.
20180708T114925+0000   --register-configs            Register all configuration files.
20180708T114925+0000   --register-setup-virtualenvs  Create Python virtual environments for all the registered packs.
20180708T114925+0000   --register-fail-on-failure    Exit with non-zero if some resource registration fails. Deprecated. This is now a default behavior.
20180708T114925+0000   --register-no-fail-on-failure Don't exit with non-zero if some resource registration fails.
20180708T114925+0000   --verbose                     Output additional debug and informational messages.
20180708T114925+0000
[...]

PR #577, opened 2018-07-09
Point to v2.7 branch until v2.8 release is out.

Kami:
We should point to v2.7 branch of st2-packages otherwise things will break because v2.8.0 is not out yet.

Should resolve https://github.com/StackStorm/st2-packages/issues/576.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -15,7 +15,7 @@ USERNAME=''
 PASSWORD=''
 
 # Note: This variable needs to default to a branch of the latest stable release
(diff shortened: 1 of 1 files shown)


Kami:
Note - we need to revert this change as soon as v2.8.0 packages are out / promoted.

As mentioned here - https://github.com/StackStorm/st2-packages/issues/576#issuecomment-403399043, I think we should make that change as last step of our release finalize workflow. This way we won't experience such issues in the future.

Kami:
Confirmed this fix is working for the time being.

PR #579, opened 2018-07-11
Also include runners in-requirements.txt when generating final requirements files

Kami:
Found yet another issue while working on a big rabbit hole called new winrmi runner tests and related work.

Before those changes, we didn't have any runner with a dependency / requirement which is not part of any other component so things worked out of the box.

Now new winrm runner depends on the `pywin` library and we need to take this into account when generating final `requirements.txt` otherwise this requirement will be skipped and packages won't be completed (`pywinrm` requirement will be missing).

Closes #562

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -76,7 +76,7 @@ populate_version: .stamp-populate_version
 	touch $@
 
 requirements:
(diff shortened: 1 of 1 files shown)


armab:
Looks like a duplicate of https://github.com/StackStorm/st2-packages/pull/562

Kami:
@arma thanks, I some how missed that one :/

PR #584, opened 2018-09-22
Removed default site on RHEL/CentOS hosts

nmaludy:
Fixes https://github.com/StackStorm/st2/issues/4358

In order to prevent the default splash screen from showing, we need to remove the default config on RHEL/CentOS hosts.


nmaludy:
@armab this should be he fix for one-liner installs and packer-st2

PR #585, opened 2018-09-25
Install all runner packages into st2 virtual environment

Kami:
This pull request updates requirements file so we now install all runner packages into the generated StackStorm virtualenv.

Other includes changes:

* ``/opt/stackstorm/runners`` directory is not used anymore so it has been removed from package definition files
* ``sudo st2ctl reload --register-runners`` command is not needed on upgrade anymore so it has been removed from postinst step

Part of https://github.com/StackStorm/st2/pull/4217.


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       # with real gitrevs during releases. Note that they are commented out, so that they do not interfere
       # with build parameters. st2cd prep tasks will uncomment these on a branch, and replace with proper
       # gitrefs.
-      # ST2_GITREV: ""
+      ST2_GITREV: "stevedore-runners"
```
armab: Let's please avoid this approach in future as it fails all the PRs in st2 repo as well as release automation. And so packages which should be built + deployed to staging-unstable are not there because this PR pins target branch.

It's OK to pin the branch for one-time CI check in PR to verify if stings work, but not good to block it for everyone else.

See alternative approach in https://github.com/StackStorm/st2-packages/pull/588 which does the needful with less sacrifice.

Kami: It was chicken and the egg / cyclic dependency problem and the only other solution was to force merge without all the tests passing.

This was supposed to be a temporary change, but then I got blocked by CI server being in a broken state so I couldn't merge other PR to get it resolved.

PR #486, opened 2017-08-18
Suppress mistral db errors in bootstrap script

Mierdin:
Lots of errors when populating Mistral's DB:

```
......
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base [-] Failed to create action: neutron.update_policy_profile: AttributeError: 'Client' object has no attribute 'update_policy_profile'
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base Traceback (most recent call last):
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base   File "/opt/stackstorm/mistral/lib/python2.7/site-packages/mistral/actions/openstack/action_generator/base.py", line 143, in create_actions
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base     client_method = class_.get_fake_client_method()
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base   File "/opt/stackstorm/mistral/lib/python2.7/site-packages/mistral/actions/openstack/base.py", line 75, in get_fake_client_method
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base     return cls._get_client_method(cls._get_fake_client())
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base   File "/opt/stackstorm/mistral/lib/python2.7/site-packages/mistral/actions/openstack/base.py", line 59, in _get_client_method
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base     attribute = getattr(attribute, attr)
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base AttributeError: 'Client' object has no attribute 'update_policy_profile'
2017-08-18 17:08:02.272 17431 ERROR mistral.actions.openstack.action_generator.base
2017-08-18 17:08:02.273 17431 ERROR mistral.actions.openstack.action_generator.base [-] Failed to create action: neutron.show_policy_profile: AttributeError: 'Client' object has no attribute 'show_policy_profile'
2017-08-18 17:08:02.273 17431 ERROR mistral.actions.openstack.action_generator.base Traceback (most recent call last):
[...]

PR #586, opened 2018-09-28
Remove verbose warnings from mistral cli alembic upgrades

bigmstone:


scripts/st2bootstrap-deb.sh
```diff
@@ -617,7 +617,7 @@ install_st2mistral() {
   sudo crudini --set /etc/mistral/mistral.conf database connection "postgresql+psycopg2://mistral:${ST2_POSTGRESQL_PASSWORD}@127.0.0.1/mistral"
 
   # Setup Mistral DB tables, etc.
-  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
+  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head | grep -v "ironicclient" || true
```
armab: I'm :+1: for filtering those warnings out, but :-1: for `|| true` construct which will suppress real mistral DB upgrade failures.

I think following similar approach we did before would work OK.
See: https://github.com/StackStorm/st2-packages/pull/486/files

PR #590, opened 2018-11-01
Add support for Ubuntu Bionic for StackStorm using Python 3

Kami:
This pull request add support for building StackStorm packages for Ubuntu Bionic using Python 3.

## Depends on

* https://github.com/StackStorm/st2packaging-dockerfiles/pull/69

## Related PRs

* https://github.com/StackStorm/st2web/pull/624
* https://github.com/StackStorm/st2chatops/pull/117
* https://github.com/StackStorm/bwc-enterprise-packages/pull/26

## TODO

- [x] New dh virtualenv debian rule for Bionic which uses python 3 binary
- [x] Support for Bionic in installer script (only for staging unstable until it's officially supported)

### DIFF
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e
+set -x
 
 my_dir="$(dirname "$0")"
 source "$my_dir/buildenv_common.sh"
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -40,6 +40,7 @@ st2_giturl() {
 # ST2_GITREV - st2 branch name (ex: master, v1.2.1). This will be used to determine correct Docker Tag: `latest`, `1.2.1`
 # ST2PKG_VERSION - st2 version, will be reused in Docker image metadata (ex: 1.2dev)
 # ST2PKG_RELEASE - Release number aka revision number for `st2` package, will be reused in Docker metadata (ex: 4)
+# ST2_PACKAGES - Which packages to build (defautls to st2 mistral).
 
 ST2_GITURL=${ST2_GITURL:-https://github.com/StackStorm/st2}
 ST2_GITREV=${ST2_GITREV:-master}
@@ -57,6 +58,13 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
+# NOTE: We don't build Mistral package on Ubuntu Bionic
+if [ "${DISTRO}" = "bionic"} ]; then
+    ST2_PACKAGES="st2"
+else
+    ST2_PACKAGES=${ST2_PACKAGES:-st2 mistral}
+fi
+
+write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL ST2_PACKAGES
 
 cat ~/.buildenv
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -58,13 +58,6 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-# NOTE: We don't build Mistral package on Ubuntu Bionic
-if [ "${DISTRO}" = "bionic"} ]; then
-    ST2_PACKAGES="st2"
-else
-    ST2_PACKAGES=${ST2_PACKAGES:-st2 mistral}
-fi
-
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL ST2_PACKAGES
+write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
 
 cat ~/.buildenv
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -37,11 +37,11 @@ st2_giturl() {
 # ST2_GITREV - st2 branch name (ex: master, v1.2.1). This will be used to determine correct Docker Tag: `latest`, `1.2.1`
 # ST2PKG_VERSION - st2 version, will be reused in Docker image metadata (ex: 1.2dev)
 # ST2PKG_RELEASE - Release number aka revision number for `st2` package, will be reused in Docker metadata (ex: 4)
-# ST2_PACKAGES - Which packages to build (defautls to st2 mistral).
 
 ST2_GITURL=${ST2_GITURL:-https://github.com/StackStorm/st2}
 ST2_GITREV=${ST2_GITREV:-master}
 ST2PKG_VERSION=$(fetch_version)
+
 # for PackageCloud
 if [ -n "$PACKAGECLOUD_TOKEN" ]; then
   ST2PKG_RELEASE=$(.circle/packagecloud.sh next-revision ${DISTRO} ${ST2PKG_VERSION} st2)
@@ -55,13 +55,6 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-# NOTE: We don't build Mistral package on Ubuntu Bionic
-if [ "${DISTRO}" = "bionic"} ]; then
-    ST2_PACKAGES="st2"
-else
-    ST2_PACKAGES=${ST2_PACKAGES:-st2 mistral}
-fi
-
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL ST2_PACKAGES
+write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
 
 cat ~/.buildenv
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -7,17 +7,20 @@ source "$my_dir/buildenv_common.sh"
 distros=($DISTROS)
 DISTRO=${distros[$CIRCLE_NODE_INDEX]}
 
-echo "Using distros: ${DISTRO}"
+echo "Using distro: ${DISTRO}"
+echo "Using Python: $(python --version 2>&1)"
+echo "ST2_GITURL: ${ST2_GITURL}"
+echo "ST2_GITREV: ${ST2_GITREV}"
 
 fetch_version() {
   if [ -f ../st2/st2common/st2common/__init__.py ]; then
     # Get st2 version based on hardcoded string in st2common
     # build takes place in `st2` repo
-    python -c 'execfile("../st2/st2common/st2common/__init__.py"); print __version__'
+    python -c 'execfile("../st2/st2common/st2common/__init__.py"); print(__version__)'
   else
     # build takes place in `st2-packages` repo
     curl -sSL -o /tmp/st2_version.py ${ST2_GITURL}/raw/${ST2_GITREV}/st2common/st2common/__init__.py
-    python -c 'execfile("/tmp/st2_version.py"); print __version__'
+    python -c 'execfile("/tmp/st2_version.py"); print(__version__)'
   fi
 }
 
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -58,6 +58,13 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
+# NOTE: We don't build Mistral package on Ubuntu Bionic
+if [ "${DISTRO}" = "bionic"} ]; then
+    ST2_PACKAGES="st2"
+else
+    ST2_PACKAGES=${ST2_PACKAGES:-st2 mistral}
+fi
+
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -107,6 +105,67 @@ Current community packages are hosted on https://packagecloud.io/StackStorm. For
 - [RHEL7/CentOS7](https://docs.stackstorm.com/install/rhel7.html)
 - [RHEL6/CentOS6](https://docs.stackstorm.com/install/rhel6.html)
 
+## Adding Support For a New Distribution
+
+If you are adding support for a new distribution for which ``packagingbuild`` and ``packagingtest``
+images are not yet published to Docker Hub and you want to test the build pipeline locally, you
+need to update ``docker-compose.yml`` file to use locally built Docker images.
+
+For example:
```
armab: Please include references to files and high-level ideas abut what needs to be done, but not the code samples as it's gets outdated quickly and every distro could have their own specifics.

scripts/st2bootstrap-deb.template.sh
```diff
@@ -74,6 +75,14 @@ setup_args() {
     fi
   fi
 
+  # Right now Bionic is not officially supported yet so we only support using staging unstable packages
```
armab: Works to me :+1:

README.md
Kami: It's not meant to be copied directly, just to give user some general idea on what needs to be done.

docker-compose.circle.yml
```diff
@@ -22,6 +22,26 @@ xenial:
...
+  extends:
+    file: docker-compose.override.yml
+    service: suite-compose
+  environment:
+    - BUILDNODE=bionicbuild
+    - TESTNODE=bionictest
+    # NOTE: We don't build st2mistral package so we don't mix Python 3
+    # (StackStorm components) and Python 2 (Mistral)
+    # Bionic will be first release without a Mistral
+    - ST2_GITREV=fix_py3_issues
```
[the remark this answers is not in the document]
Kami: Yeah, the plan was to get that one merged before this PR.

packages/st2/Makefile
```diff
@@ -4,10 +4,13 @@ ST2PKG_RELEASE ?= 1
 ST2PKG_VERSION ?= $(shell python -c "from $(ST2_COMPONENT) import __version__; print __version__,")
 CHANGELOG_COMMENT ?= "automated build, version: $(ST2PKG_VERSION)"
 RUNNERS := $(shell ls ../contrib/runners)
+DEB_DISTRO := $(shell lsb_release -cs)
```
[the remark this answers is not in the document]
Kami: Good catch, will update it.

Kami:
Re CI issues and packaging not being produced yet - yeah, they should get produced once this PR is merged (https://github.com/StackStorm/st2packaging-dockerfiles/pull/69).

Will follow it through tomorrow once that PR is merged. In theory, it should work out of the box, might just need to tag a release in that repository, similar to what it looks like we do for other distros and then I assume Docker Hub should automatically build the image and tag it appropriately.

README.md
Kami: Will take a look at it again tomorrow and re-think / re-word it if I can, worst case, I will just remove it.

My goal was to just add something which helped me and I thought it may perhaps or also add others.

docker-compose.circle.yml
```diff
@@ -22,6 +22,25 @@ xenial:
...
+  extends:
+    file: docker-compose.override.yml
+    service: suite-compose
+  environment:
+    - BUILDNODE=bionicbuild
+    - TESTNODE=bionictest
+    # NOTE: We don't build st2mistral package so we don't mix Python 3
+    # (StackStorm components) and Python 2 (Mistral)
+    # Bionic will be first release without a Mistral
+    - ST2_PACKAGES=st2
```
armab: Try to set an `ST2_PACKAGES` ENV in docker-compose definition for both `bionicbuild` and `bionictest` (instead of `buildenv_st2.sh`)

Kami: Thanks - I was actually thinking about trying that, but then I checked at the code and assumed buildenv stuff still has priority based on the way how it's passed to docker compose.

Kami: It looks like it indeed doesn't work - https://circleci.com/gh/StackStorm/st2-packages/3180#queue-placeholder/containers/2 and I will need to go with my original approach.

As I mentioned above, I wanted to try this first then I checked docker compose script and I assumed it won't work because of the way strict is structured.

Kami:
Alright, the Bionic packages are building now.

I needed to make various changes to make it work:

1. Change Circle CI parallelism from 4 to 5 so we build packages for all the distros (once we drop support for Trusty we will be back to 4 again).
2. Update `packagecloud.sh` script so it supports Bionic. Right now we only upload Bionic packages for staging unstable (intentional, until we officially announce support for it).
3. Update `buildenv_st2.sh` script so it correctly sets `ST2_PACKAGES` environment variable so we only build st2 package and not mistral on Bionic. @armab and I had a discussion on this above. Sadly due to the way how our `docker-compose2.sh` script is structured, I can't just specify those values in docker compose definition file, but need to utilize this script so it works correctly on Circle CI (small discrepancy between local and Circle CI build environment).
4. Various other changes.

Next step is testing and getting bootstrap script to work on Bionic. One that works, I think we should review this and merge it into master.

And after that (or perhaps before), I will deal with Docker Hub repo recreation drama.

.circle/buildenv_st2.sh
```diff
@@ -51,6 +58,13 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
+# NOTE: We don't build Mistral package on Ubuntu Bionic
+if [ "${DISTRO}" = "bionic"} ]; then
+    ST2_PACKAGES="st2"
+else
+    ST2_PACKAGES=${ST2_PACKAGES:-st2 mistral}
```
armab: `buildenv_st2.sh` is intended to be st2-focused only, while `buildenv_mistral.sh` is about mistral only and so they're responsible for different packages but not alltogether.

This part I definitely don't like.

Will try different things to workaround it, - but in a meantime we have `buildenv_common.sh` which can be favored instead.

docker-compose.circle.yml
armab: And so we use the one taking most precedence via `-e` :/

.circle/buildenv_st2.sh
```diff
@@ -1,21 +1,27 @@
 #!/bin/bash
 set -e
+set -x
```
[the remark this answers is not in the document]
Kami: Yeah, although the default output is very unfriendly - if there is an issue in this script it will just show "script exited with 1" which provides no helpful / debug info on failure.

.circle/buildenv_st2.sh
```diff
@@ -1,21 +1,27 @@
...
 my_dir="$(dirname "$0")"
 source "$my_dir/buildenv_common.sh"
 
 distros=($DISTROS)
 DISTRO=${distros[$CIRCLE_NODE_INDEX]}
 
+echo "Using distro: ${DISTRO}"
+echo "Using Python: $(python --version 2>&1)"
+echo "ST2_GITURL: ${ST2_GITURL}"
+echo "ST2_GITREV: ${ST2_GITREV}"
```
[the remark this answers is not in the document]
Kami: Yep, but I wanted to print it before if some step in the script fails (which happened in my case and see comment below - default output is useless and provides no info which would help with debugging).

.circle/buildenv_st2.sh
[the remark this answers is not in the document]
armab: We really touch this part of st2-packages once in a year.

With that much verbosity what's more important like "export" statements with actual values, - they become less visible and harder to quick-read.
This noise makes it harder to see what's important in a day-to-day builds.

Alternative is to improve script error reporting for the edge cases, instead of using debug mode `-x`.

.circle/buildenv_st2.sh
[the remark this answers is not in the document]
armab: But we're done with the debugging stage, right?

packages/st2/debian/control
```diff
@@ -4,16 +4,16 @@ Priority: optional
...
-               dh-virtualenv (>= 0.8)
+               dh-virtualenv (>= 0.8),
 Standards-Version: 3.9.5
 Homepage: http://stackstorm.com/
 Vcs-Git: git://github.com/stackstorm/st2.git
 Vcs-Browser: https://github.com/stackstorm/st2
 
 Package: st2
 Architecture: any
-Pre-Depends: dpkg (>= 1.16.16), python2.7, ${misc:Pre-Depends}, adduser
```
armab: I wouldn't touch such things like `${misc:Pre-Depends}` as it's something that's autogenerated by the debhelper based on dependencies he'll find package relies on.

Kami: Yeah, I was just testing it - I will put it back once I confirm everything is working.

.circle/buildenv_st2.sh
Kami: Yeah, I'm also not a big fan of it either. For now, I will move it into ``buildenv_common.sh``.

Maybe eventually we can refactor how our docker-compose2.sh script works so it will only pass environment variable to docker compose if it's set.

.circle/buildenv_st2.sh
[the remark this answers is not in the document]
Kami: I will remove it for now since I don't have the motivation to re-do all the error handling in that script.

.circle/buildenv_st2.sh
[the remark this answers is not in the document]
Kami: I will remove last two, but leave Python version and distro in place since it can still come handy.

packages/st2/debian/rules
```diff
@@ -13,16 +13,41 @@ DH_VIRTUALENV_INSTALL_ROOT := /opt/stackstorm
...
+# Trusty, Xenial and Bionic depend on different libraries.
+# Most important, Bionic uses Python 3 and others use Python 2.7
+BIONIC_PRE_DEPENDS := python3 (>= 3.6)
+BIONIC_DEPENDS := python3-distutils, python3-dev
+
+OTHER_PRE_DEPENDS := python2.7
+OTHER_DEPENDS = libpython-dev | python-dev
+
+ifeq ($(DEB_DISTRO),bionic)
+	BUILD_PRE_DEPENDS := $(BIONIC_PRE_DEPENDS)
```
armab: Because `BIONIC_PRE_DEPENDS` and similar vars are not used everywhere more than 1 time, can we just simplify it to something like

```suggestion
	BUILD_PRE_DEPENDS := python3 (>= 3.6)
```

This will move real values closer to context, making it easier to understand the flow.
Otherwise var->var->var dependency is too Pythonic :)

packages/st2/debian/control
```diff
@@ -4,16 +4,16 @@ Priority: optional
...
-Depends: ${shlibs:Depends}, ${misc:Depends}, sudo, adduser, libpython-dev | python-dev, libssl-dev, libffi-dev, git, libpam0g, openssh-server, openssh-client, bash, netbase
+Pre-Depends: dpkg (>= 1.16.16), ${pre:Depends}, ${misc:Pre-Depends}, adduser
+Depends: ${shlibs:Depends}, ${misc:Depends}, sudo, adduser, ${Depends}, libssl-dev, libffi-dev, git, libpam0g, openssh-server, openssh-client, bash, netbase
```
armab: :+1:

.circleci/config.yml
```diff
@@ -122,6 +121,7 @@ jobs:
       - image: circleci/ruby:2.4
     working_directory: /tmp/deploy
     environment:
+      # TODO: Add support for deploying Bionic to Staging Unstable
```
armab: As I understand this could be removed and bionic is OK to add to the deployment list as you already covered the Packagecloud.sh changes

.circle/buildenv_st2.sh
```diff
@@ -51,6 +55,13 @@ re="\\b$DISTRO\\b"
 
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
-write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
+# NOTE: We don't build Mistral package on Ubuntu Bionic
```
armab: As code was moved to `buildenv_common.sh` now and can revert injecting this logic in `buildenv_st2.sh`?

Kami: Done.

.circleci/config.yml
Kami: Correct - will do.

Kami:
@armab I believe I addressed all the comment and this PR is now ready for a final review.

I think it would be good to merge st2web, st2chatops and st2enterprise changes to those Ubuntu Bionic packages are built and uploaded to packagecloud.io.

This way I can test Bionic installer script end to end - basically I tested everything except st2web and st2chatops and enterprise for which we don't have packages at the moment so I can't test that.

Kami:
For the approach mentioned above to work, I will need to merge ``packagecloud.sh`` changes into master before other changes in this branch.

I will create a separate PR with changes in https://github.com/StackStorm/st2-packages/pull/590/files#diff-c1a0d8b9f26079f47f9a3565b51a2f89R1 which can be merged immediately (it's fully backward and forward compatible, has no negative implications).

scripts/st2bootstrap-deb.template.sh
```diff
@@ -114,14 +128,17 @@ install_st2_dependencies() {
   sudo apt-get update
 
   # Note: gnupg-curl is needed to be able to use https transport when fetching keys
-  sudo apt-get install -y gnupg-curl
+  if [[ "$SUBTYPE" != 'bionic' ]]; then
+    sudo apt-get install -y gnupg-curl
```
armab: This specific package doesn't present in Bionic?

Kami: Yeah, it's not (https://packages.ubuntu.com/search?keywords=gnupg-curl) and https transport for curl adding keys to APT works by default.

Kami:
Now that all the other repo (st2chatops, st2web, bwc-enterprise) PRs have been merged, I confirmed this works end to end:

1. Installer script works
2. self check works
3. Basic functionality works

I will go ahead and merge this into master and start working on hooking this up to cicd.

PR #591, opened 2018-11-07
Update Node.js version to 10

blag:
Update Node.js for new st2chatops packages.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -653,8 +653,26 @@ EOT"
 }
 
 install_st2chatops() {
-  # Add NodeJS 10 repo
-  curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
+  INSTALLABLE_ST2CHATOPS_VERSION=$(apt show st2chatops | grep 'Version:' | sed 's/^Version: \([[:digit:]]\+\)\.\([[:digit:]]\).*/\1.\2')
+
+  # Check if the st2chatops version is >= 3.0
+  python -c <<EOF
+import sys
+if '${INSTALLABLE_ST2CHATOPS_VERSION}' >= '3.0':
+    sys.exit(0)
+else:
+    sys.exit(1)
+EOF
+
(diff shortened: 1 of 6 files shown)


Kami:
Related PR: https://github.com/StackStorm/st2chatops/pull/116

scripts/st2bootstrap-deb.sh
```diff
@@ -653,8 +653,26 @@ EOT"
...
 
 install_st2chatops() {
-  # Add NodeJS 6 repo
-  curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+  INSTALLABLE_ST2CHATOPS_VERSION=$(apt show st2chatops | grep 'Version:' | sed 's/^Version: \([[:digit:]]\+\)\.\([[:digit:]]\).*/\1.\2')
+
+  # Check if the st2chatops version is >= 3.0
+  python -c <<EOF
+import sys
+if '${INSTALLABLE_ST2CHATOPS_VERSION}' >= '3.0':
```
Kami: I don't think this will work as expected.

We either need to parse it into a tuple or use `StrictVersion` from distutils or similar.

scripts/st2bootstrap-deb.sh
```diff
@@ -653,8 +653,26 @@ EOT"
...
+    sys.exit(0)
+else:
+    sys.exit(1)
```
Kami: Python interpreter exits with 1 by default if there is a syntax error in a file or similar so we should use some other, more unique exit code which won't have false positives.

scripts/st2bootstrap-deb.sh
LindsayHill: I don't think we need to do this at all. The only time you would want to install older node version is if you're passing flags to the install script to tell it to install an older ST2 version. 

But in that case it will retrieve the OS-specific install script from the relevant branch - see https://github.com/StackStorm/st2-packages/blob/master/scripts/st2_bootstrap.sh#L170

So we don't need to put this code in here. Anyone pulling this code will be installing from dev, or in future it will be >= 3.x

blag:
@Kami Yes, all of these PRs should go in at roughly the same time.

PR #592, opened 2018-11-07
Add support for Ubuntu Bionic to packagecloud.sh

Kami:
This pull request adds support for Ubuntu Bionic to packagecloud.sh script.

We need to merge that PR before we can merge other PRs in st2web, st2chatops, etc. to avoid chicken and the egg problem.

Cherry picked from https://github.com/StackStorm/st2-packages/pull/590.

### DIFF
--- a/.circle/packagecloud.sh
+++ b/.circle/packagecloud.sh
@@ -122,7 +122,7 @@ function deploy() {
 
     # NOTE: At the moment we only support staging unstable Bionic packages
     if [ "${PKG_OS}" = "bionic" ]; then
-      if [ "${PACKAGECLOUD_REPO}" != "staging-unstable"]; then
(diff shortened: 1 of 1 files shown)


.circle/packagecloud.sh
```diff
@@ -120,6 +120,14 @@ function deploy() {
      continue
     fi
 
+    # NOTE: At the moment we only support staging unstable Bionic packages
+    if [ "${PKG_OS}" = "bionic" ]; then
+      if [ "${PACKAGECLOUD_REPO}" != "staging-unstable" ]; then
```
armab: It's not enough, since there is also related `staging-enterprise-unstable` repo

Kami: https://github.com/StackStorm/st2-packages/pull/593

PR #598, opened 2018-11-19
Don't bundle and include Mistral runner in Ubuntu Bionic packages

Kami:
This pull request updates Makefile so we don't bundle and include Mistral runner on Ubuntu Bionic packages which ship without Mistral.

This will also allow us to more dynamically check if Mistral is available on a particular installation by checking if Mistral runner is installed and available.

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -3,7 +3,6 @@ ST2_COMPONENT := $(notdir $(CURDIR))
 ST2PKG_RELEASE ?= 1
 ST2PKG_VERSION ?= $(shell python -c "from $(ST2_COMPONENT) import __version__; print __version__,")
 CHANGELOG_COMMENT ?= "automated build, version: $(ST2PKG_VERSION)"
-RUNNERS := $(shell ls ../contrib/runners)
 
 ifneq (,$(wildcard /etc/debian_version))
 	DEBIAN := 1
@@ -21,6 +20,13 @@ else ifeq ($(DEB_DISTRO),bionic)
 	PYTHON_BINARY := /usr/bin/python3
 else
 	PYTHON_BINARY := python
+endif
+
+# Don't include Mistral runner on Bionic
+ifeq ($(DEB_DISTRO),bionic)
+	RUNNERS := $(shell ls -I mistral* ../contrib/runners)
+else
+	RUNNERS := $(shell ls ../contrib/runners)
 endif
 
 # Note: We dynamically obtain the version, this is required because dev


Kami:
It needed a bit more changes, but it's working now.

I confirmed generate package for Bionic doesn't contain ``python-mistralclient`` and ``stackstorm-runner-mistral-v2`` package in bundled Python virtual environment.

Kami:
Actually, it looks like that for the time being we will need to leave mistralclient in there because some StackStorm code depends on it.

rake/spec/default/70-st2_actions-integrity_spec.rb
```diff
@@ -4,7 +4,7 @@
     its(:exit_status) { is_expected.to eq 0 }
   end
 
-  describe command("st2 pack install github") do
+  describe command("st2 pack install xml") do
```
armab: https://github.com/stackstorm-exchange/stackstorm-github contains `actions`, `aliases`, `rules` and `sensors`, opposing to `xml` contains `actions` only.

armab:
Sadly, this becomes even uglier and uglier with mistral yes/no conditionals :/

Kami:
@armab That's something we discussed in the past. It's a fact of life. No way to avoid it until we fully remove Mistral.

Down the road though, I would like to make this a bit more readable by introducing a new variable e.g. "INCLUDES_MISTRAL" or similar so It's a bit more distro agnostic.

For now that should work though - and the "plan" was so we can hopefully drop Mistral all together before we will need to make this more distro agnostic.

rake/spec/default/70-st2_actions-integrity_spec.rb
Kami: In this case I will change it to use ``github`` if ``mistral_enabled`` is True, otherwise use a different pack.

armab: What's wrong with github pack?

Kami: It includes a Mistral workflow.

Kami: It looks like this pack contains all the resources (minus sensors) and doesn't rely on a Mistral workflow - https://github.com/StackStorm-Exchange/stackstorm-hubot

armab: `hubot` would be good enough

Kami: Pushed that change - I also think it's better than installing two different packs based if Mistral if available or not.

Kami:
It turns out this change needs quite a bit more work on StackStorm side.

In the worst case, I will postpone it for the v3.1.0 release.

Kami:
Another thing worth noting - runners are now full blown Python packages.

That means that if some user wanted to install Mistral runner (Mistral server itself is a different story and out of scope of this change) on a server / deployment without Mistral, that would be quite easy:

```bash
/opt/stackstorm/st2/bin/pip install "git+https://github.com/stackstorm/st2.git#egg=stackstorm-runner-mistral-v2&subdirectory=contrib/runners/mistral_v2"
```

PR #600, opened 2018-12-14
Fix --version installer script argument on Ubuntu Bionic

Kami:
Ubuntu Bionic packages don't include and support Mistral so we should skip st2mistral package check when using ``--version`` installer script argument on Bionic (e.g. ``bash install.sh --user=st2admin --password=Ch@ngeMe --staging --unstable --version=2.10dev ``).


Kami:
@armab Thanks!

PR #602, opened 2019-01-08
Don't install nginx from EPEL repos

blag:
Implements @LindsayHill's suggestion from StackStorm/st2docs#815.

Companion PR to StackStorm/st2docs#838.


armab:
I'm fine with the change and good to see that revived.
As a recommended measure to prevent breaking the one-line installer, can you please check that script works end-to-end by running it against EL instance?
I don't think we run something like that as a PR check.

blag:
Ran it against CentOS 7 with Vagrant - added sudo to commend, now it's good.

PR #611, opened 2019-05-14
Only call set_permissions at upgrade action

igorfernandes:
With "-ge" option will call set permission even on rpm install. Need to test if $1 is greater than 1 and call only on upgrade.

### DIFF
--- a/packages/st2/rpm/preinst_script.spec
+++ b/packages/st2/rpm/preinst_script.spec
@@ -54,6 +54,6 @@ set_permissions() {
 create_users
 
 # We perform upgrade (when install count > 1)
-if [ "$1" -ge 1 ]; then
+if [ "$1" -gt 1 ]; then
   set_permissions "$RESET_PERMS"
 fi


armab:
Thank you, that's a good catch!

packages/st2/rpm/preinst_script.spec
```diff
@@ -54,6 +54,6 @@ set_permissions() {
 create_users
 
 # We perform upgrade (when install count > 1)
```
armab: :man_facepalming:

packages/st2/rpm/preinst_script.spec
```diff
@@ -54,6 +54,6 @@ set_permissions() {
...
-if [ "$1" -ge 1 ]; then
+if [ "$1" -gt 1 ]; then
```
armab: :man_facepalming:

Kami:
Indeed, good catch, thanks!

We should also cherry pick this into v3.0 branch. I will be working on v3.0.1 in the next couple of days so that fix is timely as well :)

armab:
I guess we might have something like that in other repos as well, as copy-pasted from this one :)
Will check to be sure

Kami:
Looking at the code again - it doesn't seem to hurt if we also call that set permission stuff on install as well.

In any case, we need to test it and make sure it works correctly. And if we stay with the current version (do it on install and on upgrade) we need to update comment in the code to reflect that.

Kami:
So, yeah, looking at the code (https://github.com/StackStorm/st2-packages/blob/f91c3001c0e95b197f4057407ef9e6a5453d5782/packages/st2/rpm/preinst_script.spec#L49) we only reset permissions if they are not already set correctly so in theory it doesn't hurt to also run it on installation.

In personally fine either way. 

The question also is if this change is even still need. We added it in the past for backward compatibility reasons, so it's likely not even needed anymore :)

EDIT: It looks like we do the same thing on postinst so indeed only doing it in here on upgrade should be fine.

armab:
Will cherry pick to stable branch too

PR #613, opened 2019-05-22
Make sure we restart st2auth service after configuring auth in st2.conf

Kami:
I noticed that Ubuntu Bionic tests fail every now and then on run bootstrap script task.

It turns out it fails because ``st2auth`` service is not started.

After digging in, it turns out there is a race between the time when we configure authentication in st2.conf and staring / restarting st2 services.

If st2auth is started before auth is configured in st2.conf, it won't work correctly.


Kami:
I will take the liberty to merge this into master and v3.0 because I'm blocked on v3.0.1 release and there are tons of other changes and fixes I need to make, to make progress on the release.

PR #614, opened 2019-06-07
Update pip to latest version

armab:
> Part of https://github.com/StackStorm/st2/issues/4681

When PIP introduced breaking change `v10.0` months ago it broke too much on our CI side.
See the rabbit hole: https://github.com/StackStorm/st2-packages/pull/549 where we had to to pin `pip` to working `v9.x` in many repos.

This is a re-try to see what stabilized and what else still fails like upstream pip packages we use.
Hopefully update our build env to latest PIP tech.


armab:
Looks like majority of the fixes we needed were already landed into `st2` repo during the maintenance work and also py3 efforts.

PR #616, opened 2019-06-17
Fix Ubuntu Bionic install failure caused by updated 'libssl' that requires user prompt

armab:
See related https://github.com/StackStorm/st2-packages/pull/615 and https://github.com/StackStorm/st2packaging-dockerfiles/pull/71

New `libssl` version which is a dependency for `st2` package now asks for user prompt asking if it should restart affected services or not.
Answering default option, otherwise scripted Ubuntu Bionic installation is waiting for user input and no `apt-get install -y` flag can help.

PR #617, opened 2019-06-18
Make sure we also set DEBIAN_FRONTEND environment variable when installing dev version of packages

Kami:
#616 fixed Bionic Build for stable packages (afaik) and I believe this should fix it for dev versions (scenario where we directly wget the package).


Kami:
Another option would be to just set ``DEBIAN_FRONTEND=noninteractive`` globally. In theory, it shouldn't hurt anything since the whole installer script should be non-interactive anyway.

Kami:
I confirmed it's working by running ``bash st2_bootstrap.sh --force-branch=bionic_fix --staging --unstable --dev=st2/10367 --user=st2admin --password=Ch@ngeMe``.

I will go ahead and merge it to unblock the builds (and also re-run failed executions).

Kami:
I confirmed it fixes installer script timeout issue by re-running the build, but now some other step failed with a different error...

armab:
:+1:

PR #620, opened 2019-07-29
Use dist_utils module which doesn't rely on internal pip API

Kami:
Same change as in https://github.com/StackStorm/st2/pull/4750, but for this repo.


Kami:
There are actually more repos which use that file, so needs some more love :)

Issue #624, opened 2019-09-19
Bug in echo command on st2bootstrap-deb.sh


AshleyDumaine:
* [x] Operating system: Ubuntu 16.04.6 LTS
  * [x] StackStorm version: st2 3.1.0, on Python 2.7.12

# Issue details
I'm trying to install Stackstorm with the one-line install command here, however it's failing: https://docs.stackstorm.com/install/index.html#ref-one-line-install

I noticed on line 500 (https://raw.githubusercontent.com/StackStorm/st2-packages/v3.1/scripts/st2bootstrap-deb.sh) there's a bad echo command:
```
  sudo sh -c 'echo "security:\n  authorization: enabled" >> /etc/mongod.conf'
```
This puts the \n (not an actual newline) into the mongodb config and makes it so mongodb can't start. Instead of `echo` this should probably be `printf` so the \n gets properly formatted.

armab:
```sh
$ sudo sh -c 'echo "security:\n  authorization: enabled"'
security:
  authorization: enabled
```

What's your OS/environment?

AshleyDumaine:
Very odd, though I do notice adding `-e` on the echo resolves it:
```
$ sudo sh -c 'echo "security:\n  authorization: enabled"'  
security:\n  authorization: enabled
$ sudo sh -c 'echo -e "security:\n  authorization: enabled"'
security:
  authorization: enabled
```

armab:
I'm guessing the behavior depends on different versions used. Sounds like `-e` might work or not too (https://stackoverflow.com/questions/8467424/echo-newline-in-bash-prints-literal-n).

I'm good with your suggestion replacing it with `printf`:+1:
Mind creating a PR to fix this?

AshleyDumaine:
Sounds good, will do!

PR #625, opened 2019-09-19
Switch echo to printf for proper formatting of \n

AshleyDumaine:
Resolves https://github.com/StackStorm/st2-packages/issues/624

For reference: https://stackoverflow.com/a/8467449

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -497,7 +497,7 @@ quit();
 EOF
 
   # Require authentication to be able to acccess the database
-  sudo sh -c 'printf "security:\n  authorization: enabled" >> /etc/mongod.conf'
+  sudo sh -c 'printf "security:\n  authorization: enabled\n" >> /etc/mongod.conf'
 
   # MongoDB needs to be restarted after enabling auth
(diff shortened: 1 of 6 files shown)


scripts/st2bootstrap-deb.sh
```diff
@@ -497,7 +497,7 @@ quit();
 EOF
 
   # Require authentication to be able to acccess the database
-  sudo sh -c 'echo "security:\n  authorization: enabled" >> /etc/mongod.conf'
+  sudo sh -c 'printf "security:\n  authorization: enabled" >> /etc/mongod.conf'
```
armab: Looks like there is some small diff in results, when echo adds new line at the end, while printf doesnt:

```
vagrant@stackstorm:~$ sudo sh -c 'echo "security:\n  authorization: enabled"'
security:
  authorization: enabled
vagrant@stackstorm:~$ 
```

```
$ sudo sh -c 'printf "security:\n  authorization: enabled"'
security:
  authorization: enabledvagrant@stackstorm:~$ 
```

This could be a problem modifying `/etc/mongod.conf` this way. Let's please also add missing new line at the end to preserve the behavior.

AshleyDumaine: Good catch, I'll add that now

PR #629, opened 2019-12-30
WIP: temp EL8 Py2 Package Builds

punkrokk:
Goals:

- Get st2 packages working on Centos8/EL8 using py2. Least amount of changes
- Isolate the changes needed to later build st2 EL8 packages with py3 only
- debug some st2.spec issues related to latest RPM build and new conflicts


packages/st2/Makefile
```diff
@@ -90,7 +103,7 @@ post_install:
 	sed -i "/\[logger_root\]/,/\[.*\]\|\s*$$/ {s/level=DEBUG/level=INFO/}" $(DESTDIR)/etc/st2/logging.*conf
 	sed -i "/\[logger_root\]/,/\[.*\]\|\s*$$/ {s/level=DEBUG/level=INFO/}" $(DESTDIR)/etc/st2/syslog.*conf
 ifeq ($(REDHAT),1)
-	if [ `/root/scripts/platform_major_version.py` = 7 ]; then sed -i -r 's/virtualenv_opts\s*=.*/virtualenv_opts =/' $(DESTDIR)/etc/st2/st2.conf; fi
+	if [ `/root/scripts/platform_major_version.py` = 7 ] || [ `/root/scripts/platform_major_version.py` = 8 ]; then sed -i -r 's/virtualenv_opts\s*=.*/virtualenv_opts =/' $(DESTDIR)/etc/st2/st2.conf; fi
```
armab: The issue with:
```
Command /opt/stackstorm/virtualenvs/hubot/bin/python -m pip config list had error code 1
Installing setuptools, pip, wheel...
  Complete output from command /opt/stackstorm/virtualenvs/hubot/bin/python - setuptools pip wheel:
  Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/usr/lib64/python2.7/tempfile.py", line 32, in <module>
    import io as _io
  File "/usr/lib64/python2.7/io.py", line 51, in <module>
    import _io
ImportError: No module named _io
```

you mentioned is Slack comes from this line not working properly.

In both EL7 and EL8 `st2.conf` parameter should look this way:
```
[actionrunner]
virtualenv_opts =
```

You'll probably find `virtualenv_opts = --always-copy` which shouldn't be there in your `st2.conf` after you install the el8 package.

armab: Also, please format Makefile properly with `tabs` (which is part of Makefile syntax) instead of spaces which may cause problems as mentioned many times before in another PR to make sure we don't have any other issues.

https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html
https://stackoverflow.com/questions/28712585/when-to-use-space-or-tab-in-makefile

punkrokk: This is all set. But in the py2 EL8 branch

PR #631, opened 2020-01-21
WIP: EL8 installer

punkrokk:
### Done so far
- [x] Create EL8 Packages in `staging-unstable`
- [x] Prep Vagrantfile
- [x] Update st2bootstrap scripts (This PR)
- [x] Author (@punkrokk) testing
- [x] You test

### Updates for EL8:
- Currently need the rabbit packagecloud repo:
```
# Install rabbit from packagecloud
  curl -s https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | sudo bash
  sudo yum makecache -y --disablerepo='*' --enablerepo='rabbitmq_rabbitmq-server'
```
- selinux policy utils package name is different: `policycoreutils-python -> policycoreutils-python-utils`
- modify `[includes/common.sh](https://github.com/StackStorm/st2-packages/blob/f0e1e650d17ccee9c8e4ab05bfdbd1a664378089/scripts/st2bootstrap-el8.sh#L256-L260)` to force `PEM` key type. 
- MongoDB to 4.0 (Readme not updated yet)

See https://github.com/StackStorm/st2docs/pull/955 more differences from EL7 -> EL8.

### How I am testing

Here is a reasonably plain Vanilla `Vagrantfile` that I'm using. Main difference is 4 CPUs and 4GB of RAM and a Private Network static IP. YMMV. 
https://gist.github.com/punkrokk/36db930d50d0a272c93ae6049919bf5c

#### Vagrant setup
```
mkdir el8test && cd el8test
[copy above or your own Vagrantfile in this dir
vagrant up --provider virtualbox && vagrant ssh
```
In Vagrant box or your CentOS 8 VM: :
```
sudo yum -y install git
git clone https://github.com/StackStorm/st2-packages.git
cd st2-packages
git checkout el8-install-testing 
cd scripts
bash st2bootstrap-el8.sh  --unstable --staging  --user=st2admin --password=Ch@ngeMe
```

Please comment here if you:
- Succeed (Thumbs up this PR)
- Find a bug (Comment or request a change and explain why/what)

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -140,7 +140,6 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
-  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
     # added PEM to enforce PEM ssh key type in EL8 to maintain consistency
     PEM="-m PEM"
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -142,10 +142,8 @@ configure_st2_user () {
   # to be generated again.
   ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    if [[ "$ELMAJVER" == 8 ]]; then
-        # EL8 added -m PEM to force RSA PEM format
-        PEM="-m PEM"
-    fi
+    # added PEM to enforce PEM ssh key type in EL8 to maintain consistency
+    PEM="-m PEM"
     sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P "" ${PEM}
   fi
 
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -140,8 +140,13 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
+  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P ""
+    if [[ "$ELMAJVER" == 8 ]]; then
+        # EL8 added -m PEM to force RSA PEM format
+        PEM="-m PEM"
+    fi
+    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P "" ${PEM}
   fi
 
   if ! sudo grep -s -q -f ${SYSTEM_HOME}/.ssh/stanley_rsa.pub ${SYSTEM_HOME}/.ssh/authorized_keys;
--- a/scripts/st2bootstrap-el6.sh
+++ b/scripts/st2bootstrap-el6.sh
@@ -252,7 +252,6 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
-  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
     # added PEM to enforce PEM ssh key type in EL8 to maintain consistency
     PEM="-m PEM"
--- a/scripts/st2bootstrap-el6.sh
+++ b/scripts/st2bootstrap-el6.sh
@@ -252,8 +252,13 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
+  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P ""
+    if [[ "$ELMAJVER" == 8 ]]; then
+        # EL8 added -m PEM to force RSA PEM format
+        PEM="-m PEM"
+    fi
+    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P "" ${PEM}
   fi
 
   if ! sudo grep -s -q -f ${SYSTEM_HOME}/.ssh/stanley_rsa.pub ${SYSTEM_HOME}/.ssh/authorized_keys;
--- a/scripts/st2bootstrap-el6.sh
+++ b/scripts/st2bootstrap-el6.sh
@@ -254,10 +254,8 @@ configure_st2_user () {
   # to be generated again.
   ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    if [[ "$ELMAJVER" == 8 ]]; then
-        # EL8 added -m PEM to force RSA PEM format
-        PEM="-m PEM"
-    fi
+    # added PEM to enforce PEM ssh key type in EL8 to maintain consistency
+    PEM="-m PEM"
     sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P "" ${PEM}
   fi
 
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -517,7 +517,7 @@ install_mongodb() {
 
   # Need yum perl module enabled on RHEL 8
   if [[ "$RHEL" == "1" ]]; then
-    yum module enable perl:5.26
+    sudo yum -y module enable perl:5.26
   fi
 
   # Add key and repo for the latest stable MongoDB (4.0)
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -666,7 +666,7 @@ EOT"
 
   # RHEL 8 runs firewalld so we need to open http/https
   # Don't run this on ec2
-  if [[ "$RHEL" == "1" ]] && [[ $(grep -q "ec2" /sys/hypervisor/uuid)  ]]; then
+  if [[ "$RHEL" == "1" ]] && [[ $(grep -q "ec2" /sys/hypervisor/uuid) == 1 ]]; then
     sudo firewall-cmd --zone=public --add-service=http --add-service=https
     sudo firewall-cmd --zone=public --permanent --add-service=http --add-service=https
   fi
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -471,7 +471,7 @@ get_full_pkg_versions() {
 adjust_selinux_policies() {
   if getenforce | grep -q 'Enforcing'; then
     # SELINUX management tools, not available for some minimal installations
-    sudo yum install -y policycoreutils-python
+    sudo yum install -y policycoreutils-python-utils
 
     # Allow rabbitmq to use '25672' port, otherwise it will fail to start
     sudo semanage port --list | grep -q 25672 || sudo semanage port -a -t amqp_port_t -p tcp 25672
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -80,14 +80,14 @@ setup_args() {
   echo "          Installing st2 $RELEASE $VERSION              "
   echo "########################################################"
 
-  if [ "$REPO_TYPE" == "staging" ]; then
+  if [[ "$REPO_TYPE" == "staging" ]]; then
     printf "\n\n"
     echo "################################################################"
     echo "### Installing from staging repos!!! USE AT YOUR OWN RISK!!! ###"
     echo "################################################################"
   fi
 
-  if [ "$DEV_BUILD" != '' ]; then
+  if [[ "$DEV_BUILD" != '' ]]; then
     printf "\n\n"
     echo "###############################################################################"
     echo "### Installing from dev build artifacts!!! REALLY, ANYTHING COULD HAPPEN!!! ###"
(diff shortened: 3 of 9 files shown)


punkrokk:
https://github.com/hashicorp/vagrant/issues/11330

scripts/st2bootstrap-el8.sh
```diff
@@ -507,13 +511,13 @@ install_st2_dependencies() {
 install_mongodb() {
   # Add key and repo for the latest stable MongoDB (3.4)
```
armab:
```suggestion
  # Add key and repo for the stable MongoDB (4.0)
```

scripts/st2bootstrap-el8.sh
```diff
@@ -492,6 +491,11 @@ install_st2_dependencies() {
   if [[ -z "$is_epel_installed" ]]; then
     sudo dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
   fi
+
+  # Install rabbit from packagecloud
```
armab: Any background, eg why default Rabbitmq package didn't work? 
I think a better code comment would describe the reasons "why" rather than "what".

scripts/includes/common.sh
```diff
@@ -140,8 +140,13 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
+  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P ""
+    if [[ "$ELMAJVER" == 8 ]]; then
+        # EL8 added -m PEM to force RSA PEM format
+        PEM="-m PEM"
```
armab: After looking at it once again, I don't think it's good idea to include special case with "PEM" `ssh-keygen` param for EL8 installer.

The reason is that `curl|bash` script by idea is a demo installer.
With that, default, easy, clean and the most popular straightforward settings would be ideal to use.
Additionally, consistency across the platforms in other installers needs to be preserved and deviation is not recommended.

I would assume hardening cases like that are perfectly acceptable for other installers, with more extended  production-friendly configuration settings and options. For curl|bash let's please keep it simple.

scripts/st2bootstrap-el6.sh
```diff
@@ -252,8 +252,13 @@ configure_st2_user () {
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
   # NOTE: If the file already exists and is non-empty, then assume the key does not need
   # to be generated again.
+  ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
   if ! sudo test -s ${SYSTEM_HOME}/.ssh/stanley_rsa; then
-    sudo ssh-keygen -f ${SYSTEM_HOME}/.ssh/stanley_rsa -P ""
+    if [[ "$ELMAJVER" == 8 ]]; then
+        # EL8 added -m PEM to force RSA PEM format
+        PEM="-m PEM"
```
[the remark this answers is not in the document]
punkrokk: @armab - since there is an include file, I didn't have time/knowledge on how to upgrade the script `generate_file_installer_scripts.py` to only update the EL8 script. If necessary, I could build a new include or attack somehow, looking for suggestions. 

As to the reason why we need `-m PEM` on EL8, upon reviewing the [ssh-keygen manpage](http://man7.org/linux/man-pages/man1/ssh-keygen.1.html) 

`ssh-keygen will by default write keys in an OpenSSH-specific format.
     This format is preferred as it offers better protection for keys at
     rest as well as allowing storage of key comments within the private key
     file itself.  The key comment may be useful to help identify the key.
     The comment is initialized to “user@host” when the key is created, but
     can be changed using the -c option.

     It is still possible for ssh-keygen to write the previously-used PEM
     format private keys using the -m flag.  This may be used when generat‐
     ing new keys, and existing new-format keys may be converted using this
     option in conjunction with the -p (change passphrase) flag.`

Other ST2 install scripts all use older versions of `ssh-keygen` which default to `PEM`, so I added the flag to maintain consistency ( the change requires other changes to move to the new openssl key format ), which I assumed would be better left to a different PR.

scripts/st2bootstrap-el8.sh
```diff
@@ -472,7 +476,7 @@ get_full_pkg_versions() {
 adjust_selinux_policies() {
```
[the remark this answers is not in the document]
punkrokk: Yes. Will do.

scripts/st2bootstrap-el6.sh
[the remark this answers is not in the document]
armab: That makes a lot of sense now, thanks for explaining!

I think a quick reasoning behind this would be great to have as code comment.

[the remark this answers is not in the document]
armab: Also, in fact if other installers already relying on `PEM` by default, we can just add that flag globally for everything. This will remove EL8 conditional.

scripts/includes/common.sh
punkrokk: pushing

scripts/st2bootstrap-el8.sh
```diff
@@ -492,6 +496,13 @@ install_st2_dependencies() {
   if [[ -z "$is_epel_installed" ]]; then
```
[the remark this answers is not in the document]
punkrokk: @armab I think that the situation is Rabbit will be in EPEL 8 it just isn't yet, it just isn't yet. Are you thinking just removing EPEL for now and putting it back in? 

I was deferring to the minimal changes (and keeping EPEL so that we are accommodating for the repo being present), and moving back to the package once it's released.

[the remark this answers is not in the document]
armab: I'm good with leaving it as is, considering it has a reason.

Can we document that we're waiting for RabbitMQ to appear in EPEL?
TODO Issue in this repo + Code comment linking to the issue would work best, just in case if we forget to revise it later.

scripts/st2bootstrap-el8.sh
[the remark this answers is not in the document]
punkrokk: Confirmed locally. Just pushed a change to detect if in ec2 to not run firewalld commands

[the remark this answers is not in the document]
punkrokk: @armab Here is the log of a successful install: 
https://gist.github.com/punkrokk/ebceed24d2ee3a6261ae170844e98d0d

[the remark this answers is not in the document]
punkrokk: Confirmed on ec2 also. Same gist.

scripts/st2bootstrap-el8.template.sh
```diff
@@ -371,6 +359,13 @@ configure_st2chatops() {
 }
 
 trap 'fail' EXIT
+
+# check for RHEL 8
+ELMAJVER=$(cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/')
```
armab: Why do we need to check for specific `8` version considering we already running `st2bootstrap-el8` script?
This part looks redundant. I think `RHEL` check if sufficient for the logic below.

scripts/st2bootstrap-el8.sh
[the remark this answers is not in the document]
armab: Thanks! :+1:

scripts/st2bootstrap-el8.sh
```diff
@@ -492,6 +493,15 @@ install_st2_dependencies() {
...
     sudo dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
   fi
+
+  # Install rabbit from packagecloud
+  # Package are not in EPEL or CentOS repos -
+  # recommended by rabbit: https://www.rabbitmq.com/install-rpm.html#package-cloud
+  # TODO: Migrate install back to EPEL rpm when available
+  # https://github.com/StackStorm/st2-packages/issues/632
```
armab: :+1:

armab:
Looks we're close with the EL8 scripted installer :+1: 

Next stage would be asking existing @StackStorm/maintainers  + Contributors to test the EL8 bootstrap script manually and report back if anything weird, bugs or improvements found and if installation succeeds or not in different environments.
I'll try to run the script myself as well.

The instructions to run it on EL8 (RHEL8 or CentOS8) just:
```
curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --staging --unstable
```

scripts/st2bootstrap-el8.template.sh
```diff
@@ -145,15 +153,21 @@ install_st2_dependencies() {
 }
 
 install_mongodb() {
-  # Add key and repo for the latest stable MongoDB (3.4)
-  sudo rpm --import https://www.mongodb.org/static/pgp/server-3.4.asc
-  sudo sh -c "cat <<EOT > /etc/yum.repos.d/mongodb-org-3.4.repo
-[mongodb-org-3.4]
+
+  # Need yum perl module enabled on RHEL 8
+  if [[ "$RHEL" == "1" ]]; then
```
armab: Just ran the install script and it's failing on this line with the following error:

```
Last metadata expiration check: 0:02:58 ago on Mon 10 Feb 2020 11:41:16 AM UTC.
Dependencies resolved.
===================================================================================================================================================================
 Package                                Architecture                          Version                                    Repository                           Size
===================================================================================================================================================================
Installing:
 crudini                                noarch                                0.9.3-1.el8                                epel                                 29 k

Transaction Summary
===================================================================================================================================================================
Install  1 Package

Total download size: 29 k
Installed size: 61 k
Downloading Packages:
crudini-0.9.3-1.el8.noarch.rpm                                                                                                      26 kB/s |  29 kB     00:01    
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                               11 kB/s |  29 kB     00:02     
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                           1/1 
  Installing       : crudini-0.9.3-1.el8.noarch                                                                                                                1/1 
  Running scriptlet: crudini-0.9.3-1.el8.noarch                                                                                                                1/1 
  Verifying        : crudini-0.9.3-1.el8.noarch                                                                                                                1/1 

Installed:
  crudini-0.9.3-1.el8.noarch                                                                                                                                       

Complete!
main: line 520: RHEL: unbound variable
# Failed on Install st2 dependencies (MongoDB) #
[vagrant@centos8 ~]$ 
```

scripts/st2bootstrap-el8.template.sh
```diff
@@ -145,15 +153,21 @@ install_st2_dependencies() {
 }
 
 install_mongodb() {
-  # Add key and repo for the latest stable MongoDB (3.4)
-  sudo rpm --import https://www.mongodb.org/static/pgp/server-3.4.asc
-  sudo sh -c "cat <<EOT > /etc/yum.repos.d/mongodb-org-3.4.repo
-[mongodb-org-3.4]
+
+  # Need yum perl module enabled on RHEL 8
```
armab: Can you please clarify this part about yum perl module, which package or component requires it?

scripts/st2bootstrap-el8.template.sh
punkrokk: I agree. Pushing update

scripts/st2bootstrap-el8.template.sh
punkrokk: My above update will resolve this.

scripts/st2bootstrap-el8.template.sh
punkrokk: correct, we don't. Since the base script does this. Removed

scripts/st2bootstrap-el8.template.sh
punkrokk: I fixed this by adding `RHEL=0` if not RHEL.

scripts/st2bootstrap-el8.template.sh
armab: The old version with 8 version check is still in the codebase. Needs updating.

armab:
Fun fact: `rabbitmq-server` is pulling 160 other packages via dependencies, including Java.
This experience is similar to nodejs that's pulling python for dependency in `st2chatops` package :face_with_head_bandage:

scripts/st2bootstrap-el8.template.sh
punkrokk: I removed this now. Thanks. I also moved the RHEL check to a function

scripts/st2bootstrap-el8.template.sh
punkrokk: @armab When installing rabbitmq, on RHEL, I was getting this error: 

```
 From       : https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey
Generating yum cache for rabbitmq_rabbitmq-server-source...

The repository is setup! You can now install packages.
Updating Subscription Management repositories.
rabbitmq_rabbitmq-server                        548  B/s | 833  B     00:01    
Modular dependency problems:

 Problem 1: conflicting requests
  - nothing provides module(perl:5.26) needed by module perl-DBD-SQLite:1.58:8010020190322125518:073fa5fe-0.x86_64
 Problem 2: conflicting requests
  - nothing provides module(perl:5.26) needed by module perl-DBI:1.641:8010020190322130042:16b3ab4d-0.x86_64
Metadata cache created.
Updating Subscription Management repositories.
```

The perl modules resolve it. The reason I installed it here, is because I was trying to get them enabled as early as possible.

punkrokk: I have moved this to where it should have been (right before rabbitmq install) and explained the need.

punkrokk: @armab I just double checked this and tested it, and the perl modules are needed for RHEL

punkrokk:
@armab FYI - epel is required for rabbitmq dependencies (erlang specifically)

scripts/st2bootstrap-el8.sh
```diff
@@ -19,10 +19,17 @@ DEV_BUILD=''
...
 ST2WEB_PKG='st2web'
 ST2CHATOPS_PKG='st2chatops'
 
+check_for_rhel() {
+  if [[ $(cat /etc/os-release | grep 'ID="rhel"') ]]; then
+    RHEL=1
+  else
+    RHEL=0
+  fi
+}
```
armab: Nitpick comment.

Just noticed we're using it in `STEP='Check for RHEL' && check_for_rhel`, where "STEPS" are more about running important operations like installing st2 + dependencies, configuring them, etc.

How about construct like this, wouldn't it be easier?
```bash
# Check if the current OS is Red Hat Enterprise Linux
is_rhel() {
  return $(cat /etc/os-release | grep 'ID="rhel"')
}
```

And using it in code:
```bash
  # RabbitMQ on RHEL8 requires module(perl:5.26
  if is_rhel; then
    sudo yum -y module enable perl:5.26
  fi
```

punkrokk: ok. I can refactor it like that.

scripts/st2bootstrap-el8.template.sh
```diff
@@ -11,10 +11,17 @@ DEV_BUILD=''
 USERNAME=''
 PASSWORD=''
 ST2_PKG='st2'
-ST2MISTRAL_PKG='st2mistral'
 ST2WEB_PKG='st2web'
 ST2CHATOPS_PKG='st2chatops'
 
+is_rhel() {
```
armab: If we use this function like `if is_rhel; then`, - it should return the status code to work properly.

punkrokk: resolved in new PR: https://github.com/StackStorm/st2-packages/pull/635

armab: :+1:

PR #633, opened 2020-02-08
Make sure "postrm purge" command correctly cleans up

Kami:
This pull request updates ``postrm`` script to correctly remove all the StackStorm related configfiles and directories when the script is called with ``purge`` argument (aka when ``apt-get remove --purge st2*`` command is used).

I know we don't officially support re-installing StackStorm on the same host, but we should still comply with the deb package``postrm purge`` "API" and correctly clean up after the package if ``apt-get remove --purge`` flag is used.

I believe this change should be saved since ``postrm`` is only called with the ``purge`` argument if purge is explicitly requested by the user.
## TODO

- [ ] Tests


punkrokk:
@Kami should we consider something similar for RPM?

PR #634, opened 2020-02-10
Installer script should fail if any of the "mongo shell" steps fail

Kami:
Currently, installer script doesn't correctly fail if any of the mongo shell commands we run fails (e.g. admin user creation fails, user already exists, etc.).

This means at the end, user may end up with a broken / half configured installation.

---

This pull request fixes that by removing ``quit()`` function call.

When piping data to mongo shell, it will already exit with the correct status code, even if the ``quit()`` statement is not present.

For example:

```bash
vagrant@ubuntu-xenial:~$ bash 2.sh ; echo $?
MongoDB shell version v3.4.23
connecting to: mongodb://127.0.0.1:27017
MongoDB server version: 3.4.23
switched to db admin
2020-02-10T09:08:03.491+0000 E QUERY    [thread1] Error: couldn't add user: User "admin@admin" already exists :
_getErrorWithCode@src/mongo/shell/utils.js:25:13
DB.prototype.createUser@src/mongo/shell/db.js:1292:15
@(shell):1:1
0

vagrant@ubuntu-xenial:~$ cat 1.sh
mongo <<EOF
use admin;
db.createUser({
    user: "admin",
    pwd: "wwq",
    roles: [
        { role: "userAdminAnyDatabase", db: "admin" }
    ]
});
EOF
vagrant@ubuntu-xenial:~$ bash 1.sh ; echo $?
MongoDB shell version v3.4.23
connecting to: mongodb://127.0.0.1:27017
MongoDB server version: 3.4.23
switched to db admin
2020-02-10T09:09:29.551+0000 E QUERY    [thread1] Error: couldn't add user: User "admin@admin" already exists :
_getErrorWithCode@src/mongo/shell/utils.js:25:13
DB.prototype.createUser@src/mongo/shell/db.js:1292:15
@(shell):1:1
bye
1
```

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -472,6 +472,10 @@ install_mongodb() {
   sleep 5
 
   # Create admin user and user used by StackStorm (MongoDB needs to be running)
+  # NOTE: mongo shell will automatically exit when piping from stdin. There is
+  # no need to put quit(); at the end. This way last command exit code will be
+  # correctly preserved and install script will correctly fail and abort if this
+  # command fails.
   mongo <<EOF
 use admin;
 db.createUser({
@@ -481,7 +485,6 @@ db.createUser({
         { role: "userAdminAnyDatabase", db: "admin" }
     ]
 });
-quit();
 EOF
 
   mongo <<EOF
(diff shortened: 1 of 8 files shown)


punkrokk:
@armab @Kami Now that the EL8 scripts have been shipped, once tests pass I'd like to get this merged before I reach out to the #development channel to request EL8 testing.

armab:
+1 go for it

PR #639, opened 2020-04-06
Various fixes to the el8 installer script

m4dcoder:
Add -y to flag to the repoquery commands so that it doesn't prompt for y/ok during the install. Currently, it prompts for y/ok silently and the installer script will hang awhile and then fail. Create the .ssh directory for the stanley user if it does not exists.

### DIFF
--- a/scripts/includes/common.sh
+++ b/scripts/includes/common.sh
@@ -136,7 +136,8 @@ configure_st2_user () {
   SYSTEM_HOME=$(echo ~stanley)
 
   if [ ! -d "${SYSTEM_HOME}/.ssh" ]; then
-  	sudo mkdir ${SYSTEM_HOME}/.ssh
+    sudo mkdir ${SYSTEM_HOME}/.ssh
+    sudo chmod 700 ${SYSTEM_HOME}/.ssh
   fi
 
   # Generate ssh keys on StackStorm box and copy over public key into remote box.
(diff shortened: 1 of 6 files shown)


scripts/includes/common.sh
```diff
@@ -135,6 +135,9 @@ configure_st2_user () {
 
   SYSTEM_HOME=$(echo ~stanley)
 
+  if [ ! -d "${SYSTEM_HOME}/.ssh" ]; then
+  	sudo mkdir ${SYSTEM_HOME}/.ssh
```
armab, line 139:
```suggestion
    sudo mkdir ${SYSTEM_HOME}/.ssh
```

Please use 2 spaces for indentation to be consistent with the rest of the code in file.

armab: @punkrokk Do you think we'll also need to set the permissions on `~/.ssh/` after dir creation if it didn't exist by default?

m4dcoder: Fixed

m4dcoder:
I haven't encountered needing repoquery -y before on el6/7. It's only testing on centos8 that I have to explicitly call that out. i'm also using a different el8 image due to disk size issue.

armab:
FYI this PR became a regression for EL6 and EL7 as they don't support `repoquery -y`: https://github.com/StackStorm/st2-packages/issues/641

PR #640, opened 2020-04-06
Skip st2mistral check for el8 installer

m4dcoder:
The package st2mistral is not supported on el8 because st2 runs on python3 and the StackStorm fork is not tested on python3. The RHEL installer script is modified to ignore the st2mistral warning for RHEL8.

### DIFF
--- a/scripts/st2bootstrap-el7.sh
+++ b/scripts/st2bootstrap-el7.sh
@@ -436,8 +436,9 @@ get_full_pkg_versions() {
     fi
     ST2_PKG=${ST2_VER}
 
+    local RHMAJVER=`cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/'`
     local ST2MISTRAL_VER=$(repoquery -y --nvr --show-duplicates st2mistral | grep -F st2mistral-${VERSION} | sort --version-sort | tail -n 1)
-    if [ -z "$ST2MISTRAL_VER" ]; then
+    if [ -z "$ST2MISTRAL_VER" && "$RHMAJVER" != '8' ]; then
       echo "Could not find requested version of st2mistral!!!"
       sudo repoquery -y --nvr --show-duplicates st2mistral
       exit 3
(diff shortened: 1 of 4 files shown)


scripts/st2bootstrap-el7.sh
```diff
@@ -436,8 +436,9 @@ get_full_pkg_versions() {
     fi
     ST2_PKG=${ST2_VER}
 
+    local RHMAJVER=`cat /etc/redhat-release | sed 's/[^0-9.]*\([0-9.]\).*/\1/'`
     local ST2MISTRAL_VER=$(repoquery -y --nvr --show-duplicates st2mistral | grep -F st2mistral-${VERSION} | sort --version-sort | tail -n 1)
-    if [ -z "$ST2MISTRAL_VER" ]; then
+    if [ -z "$ST2MISTRAL_VER" && "$RHMAJVER" != '8' ]; then
```
armab, line 441: Can we try to avoid this kind of unnecessary construct in `el6` and `el7` scripts and make `el8` script more customized to exclude mistral in general?

m4dcoder: This part is auto-generated by make scriptsgen from a common file `scripts/includes/rhel.sh` that is shared between el6-8.  If we want a customized el8 script without mistral, it's a larger undertaking.  With mistral that is going to be deprecated by v3.3 and likely the norm in the shared script, it's not worth the effort IMHO.

punkrokk: I see both sides here. It's like 50/50 value to fix now vs. fix later. @armab if @m4dcoder is blocked by this, I'd be ok with a #TODO to flag this afterwards.

m4dcoder: The check of el8 whether to skip Mistral is also the only concrete requirement here that I can work with on generalizing the autogen and allowing for customization between el distros. The installer scripts are going to be touched in v3.3. again when we deprecate Mistral which will make the immediate change pointless.  Therefore, this is a simple decision for me here.  We can revisit refactoring of the autogen when there are concrete requirements for v3.3+ and that is not related to Mistral.

armab: Makes sense, considering we deprecate mistral completely in next `v3.3` release.
Thanks for more context.

I think at this point autogen/templating scripts are bringing more harm and confusion than value.

PR #644, opened 2020-04-28
Add -y option to yum list http-parser command

m4dcoder:
Add the -y option to the yum list http-parser command during st2chatops install to prevent install script from hanging while waiting for user to answer prompt.

### DIFF
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -692,7 +692,7 @@ EOT"
 
 install_st2chatops() {
   # Temporary hack until proper upstream fix https://bugs.centos.org/view.php?id=13669
-  if ! yum list http-parser 1>/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


scripts/st2bootstrap-el8.sh
```diff
@@ -692,7 +692,7 @@ EOT"
 
 install_st2chatops() {
   # Temporary hack until proper upstream fix https://bugs.centos.org/view.php?id=13669
```
armab, line 694: BTW @punkrokk when adding EL8 scripts, did you verify if this is still an issue?

Looks like this block was just copy-pasted from the EL7 scripts.
https://bugs.centos.org/view.php?id=13669

armab: At least CentOS8 has `http-parser` in the default AppStream repository.

punkrokk: What do you mean by `verify if this is still an issue`? I was able to install it when I shipped this.

PR #658, opened 2020-08-03
Remove EL6 and st2python

armab:
Finishing the work related to `EL6` deprecation in st2 packages: https://github.com/StackStorm/discussions/issues/39


punkrokk:
Phew, that was a lot of code removed.

armab:
It may be hard to do a research and package it properly with all the corner cases researched (EL6 st2python code might be very different for EL7).

However the hardest thing is really maintain and update the python package properly.
EL6 st2python package wasn't updated for 4 years I guess? I'm sure it's a can of worms now in terms of security. I'm happy to finally get rid of it.

punkrokk:
Not suggesting we do it, just pointing out that it’s not impossible :P

PR #660, opened 2020-08-31
Add strip to .so libraries

amanda11:
For some reason based on https://app.circleci.com/pipelines/github/StackStorm/st2/374/workflows/6bb47baf-c2a5-4b11-be22-5f86e318d9f9/jobs/11646 a few  `.so` files for  `EL8` rpm packages started to contain `BUILDROOT` in them which resulted in failed builds:

```
[package: st2]      + xargs '-I{}' -n1 sed -i s@/tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64@@ '{}'
[package: st2]      + find /tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64 -name RECORD -o -name '*.egg-link' -o -name '*.pth'
[package: st2]      + /usr/lib/rpm/check-buildroot
[package: st2]      Binary file /tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64/opt/**********/st2/lib/python3.6/site-packages/cryptography/hazmat/bindings/_padding.abi3.so matches
[package: st2]      Binary file /tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64/opt/**********/st2/lib/python3.6/site-packages/cryptography/hazmat/bindings/_constant_time.abi3.so matches
[package: st2]      Binary file /tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64/opt/**********/st2/lib/python3.6/site-packages/cryptography/hazmat/bindings/_openssl.abi3.so matches
[package: st2]      Found '/tmp/st2/st2/build/BUILDROOT/st2-3.3dev-51.x86_64' in installed files; aborting
[package: st2]      error: Bad exit status from /var/tmp/rpm-tmp.OtVxc8 (%install)
```

Call strip for any `.so` libraries, as per logic used in https://github.com/kevinconway/rpmvenv/issues/48 or https://stackoverflow.com/questions/52766091/removing-paths-from-so-files-so-that-rpm-check-buildroot-succeeds.

### DIFF
--- a/packages/st2/rpm/st2.spec
+++ b/packages/st2/rpm/st2.spec
@@ -50,7 +50,7 @@ Conflicts: st2common
 # We build cryptography for EL8, and this can contain buildroot path in the
 # built .so files. We use strip on these libraries so that there are no
 # references to the buildroot in the st2 rpm
-%if 0%{?rhel} >= 8
+%if 0%{?rhel} == 8
   %cleanup_so_abspath
 %endif
   %cleanup_python_abspath
--- a/packages/st2/rpm/st2.spec
+++ b/packages/st2/rpm/st2.spec
@@ -47,7 +47,12 @@ Conflicts: st2common
   %service_install st2scheduler
   make post_install DESTDIR=%{buildroot}
 
+# We build cryptography for EL8, and this can contain buildroot path in the
+# built .so files. We use strip on these libraries so that there are no
+# references to the buildroot in the st2 rpm
+%if 0%{?rhel} >= 8
   %cleanup_so_abspath
+%endif
   %cleanup_python_abspath
 
 %prep
--- a/packages/st2/rpm/st2.spec
+++ b/packages/st2/rpm/st2.spec
@@ -47,6 +47,7 @@ Conflicts: st2common
   %service_install st2scheduler
   make post_install DESTDIR=%{buildroot}
 
+  %cleanup_so_abspath
   %cleanup_python_abspath
 
 %prep
--- a/rpmspec/helpers.spec
+++ b/rpmspec/helpers.spec
@@ -40,6 +40,11 @@
       xargs -I{} -n1 sed -i 's@%{buildroot}@@' {} \
 %{nil}
 
+#Cleanup .so files that may contain buildroot
+%define cleanup_so_abspath \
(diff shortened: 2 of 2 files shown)


armab (approved):
:+1: Good stuff! Thanks for the fix.

@amanda11so based on your research, after looking deeper sounds like rpm build doesn't actually strip the `.so` files?

amanda11:
@armab -   Because our spec files has a post-install then it only does the stripping by the bro-strip -
And the build root  in the cryptography binding .so files Isn’t getting stripped out by that.
There is still the question of why we didn’t get this issue before, so I want to do a little more digging before I merge (unless this is needed urgently)

armab:
@amanda11 That makes sense, thanks for the deeper investigation on this issue!

amanda11:
@armab @punkrokk I've revised the code so that its less invasive. Now only performs strip on:
1) EL8 (as that's where we build crypography only for)
2) Only strip on .so files where the build root is actually found
I've still not located the root cause, could you re-check you are happy with code files.

Also is there a way of getting a E2E run using the packages from this branch? If not, then I'll fire up an EL8 StackStorm and use the st2 package manually, but if we have it setup so that it can use the packages from this branch that would be even better.

packages/st2/rpm/st2.spec
```diff
@@ -47,6 +47,12 @@ Conflicts: st2common
   %service_install st2scheduler
   make post_install DESTDIR=%{buildroot}
 
+# We build cryptography for EL8, and this can contain buildroot path in the
+# built .so files. We use strip on these libraries so that there are no
+# references to the buildroot in the st2 rpm
+%if 0%{?rhel} >= 8
```
armab, line 53: What if we'll add RHEL 9 in the future?

This would be a more correct check: 
```suggestion
%if 0%{?rhel} == 8
```

armab:
I'd say merge the PR and see what happens. We can revert it if e2e tests go wrong or any other regressions spot.

Packages are initially deployed to `staging-unstable` package repository. Nightly builds check new packages periodically and if e2e tests succeeded, - promote them to `unstable` packagecloud repository.

PR #663, opened 2020-09-17
No-longer need to exclude mistral runner, as it is removed from st2

amanda11:
Tidy-up Makefile now that mistral runner is removed from st2 code base.


armab (approved):
Thanks for the follow-up :+1:

PR #664, opened 2020-10-06
account for unique user group

lattisimo:
use 'id -gn' to get the user group

### DIFF
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -289,6 +289,7 @@ configure_st2_cli_config() {
   # Configure CLI config (write credentials for the root user and user which ran the script)
   ROOT_USER="root"
   CURRENT_USER=$(whoami)
+  CURRENT_GROUP=$(id -gn)
 
   ROOT_HOME=$(eval echo ~${ROOT_USER})
   : "${HOME:=$(eval echo ~${CURRENT_USER})}"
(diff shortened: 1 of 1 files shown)


nmaludy (changes requested):
@lattisimo Looks like the build is failing here https://app.circleci.com/pipelines/github/StackStorm/st2-packages/77/workflows/75cabd81-6e11-43dd-8c88-a8cce7ae7904/jobs/3628 

Can you please update https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el8.template.sh

Then run `make scriptsgen` and include the changed files?

Thanks!

armab:
Would be good to know the more context around this issue: what led you to this solution, what's the issue, error message, etc.

scripts/st2bootstrap-el8.sh
```diff
@@ -289,6 +289,7 @@ configure_st2_cli_config() {
   # Configure CLI config (write credentials for the root user and user which ran the script)
   ROOT_USER="root"
   CURRENT_USER=$(whoami)
+  CURRENT_GROUP=$(id -gn)
```
armab, line 292: Was the issue only on EL8 platform or do we need to apply the same patch for other OS systems?

lattisimo: @armab The issue I had was installing on a box that had LDAP configured and my user group is 'domain users' since the script just assumed the group matched the user so the chown command fails and the script stops executing. I can avoid the issue by 'sudo su' but I realized that if I get the user's group with 'id -gn' then there is no assumption of what that value should be. This would likely be relevant on the other scripts.

@nmaludy I am reading up on how to perform the steps you requested. I am very new to development so it may take me a bit before I can update. please standby.

armab: Thanks for more detailed context, that's helpful! :+1:

lattisimo:
Since this is my first interaction of this sort. I want to follow up if there is any additional action needed on my part? I am not confident that I accomplished what @nmaludy has asked of me.

nmaludy:
@lattisimo Your changes look good, we're just in a change freeze right now due to a pending 3.3.0 release. I'm going to target this for the 3.4.0 milestone, does that sound good?

lattisimo:
Sounds good. Thanks.

PR #665, opened 2020-10-09
-v doesn't allow you to set version

amanda11:
The install.sh supports -v and --version to set the version.

Using --version=3.2.0 then it allows you to configure the specify the version to install.

However if you use -v=3.2.0 or -v3.2.0 then it ignores the version specified and uses the default value set to BRANCH. If you use "-v 3.2.0" then it complains that the version -v is not valid.

I believe that you should have been able to use -v=3.2.0 or --version=3.2.0 - and have amended the setup_args so that it will extract the version number out on -v same as --version.

NB. This appears to be legacy behaviour, so should not be a blocker for 3.3 release.

### DIFF
--- a/scripts/st2_bootstrap.sh
+++ b/scripts/st2_bootstrap.sh
@@ -28,7 +28,7 @@ setup_args() {
   for i in "$@"
     do
       case $i in
-          -v|--version=*)
+          -v=*|--version=*)
           VERSION="${i#*=}"
           shift
           ;;
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -32,7 +32,7 @@ setup_args() {
   for i in "$@"
     do
       case $i in
-          -v|--version=*)
+          -v=*|--version=*)
           VERSION="${i#*=}"
           shift
           ;;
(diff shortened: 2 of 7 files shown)


m4dcoder (changes requested):
@amanda11 This is a good catch. Can you also update https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L35 and other el7/el8 scripts? There are also the template that generated those script files at https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.template.sh#L27. The main script probably just pass `--version` but we should fix the other scripts just in case.  I think you should update the deb/el7/el8 templates and then regenerate the deb/el7/el8 scripts. Thanks!

amanda11:
> @amanda11 This is a good catch. Can you also update https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L35 and other el7/el8 scripts? There are also the template that generated those script files at https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.template.sh#L27. The main script probably just pass `--version` but we should fix the other scripts just in case.  I think you should update the deb/el7/el8 templates and then regenerate the deb/el7/el8 scripts. Thanks!

Thanks @m4dcoder for the review - I have updated the extra files, ready for your re-review whenever you have time.

blag (approved):
LGTM

m4dcoder (approved):

PR #669, opened 2020-11-06
Install python3 selinux library in preparation for Python 3 move on EL7

amanda11:
Install python3 SELinux library on EL7.

Part of https://github.com/StackStorm/discussions/issues/54


amanda11:
Can be reviewed and merged before python3 package changes.
If SE Linux enabled then it will install the python 3 selinux lib (which by dependanices will pull in python3).
For non-SE Linux, then the python3 install will happen when ST2 RPM is installed as it will have python3 as a dependency. (PR for package change to be done separately)

armab (approved):

PR #671, opened 2020-11-10
EL7 to use python3

amanda11:
Update the EL7 packages so that they use python3 not python2.

Part of StackStorm/discussions#54

### DIFF
--- a/packages/st2/Makefile
+++ b/packages/st2/Makefile
@@ -27,6 +27,10 @@ else ifeq ($(EL_VERSION),8)
 	PYTHON_BINARY := /usr/bin/python3
 	PIP_BINARY := /usr/local/bin/pip3
 	PYTHON_ALT_BINARY := python3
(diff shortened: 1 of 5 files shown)


amanda11:
Comments:

- Tested with docker images built by https://github.com/StackStorm/st2packaging-dockerfiles/pull/92
- In EL8 we build cryptography. I haven't built cryptography for EL7, and it appears to use the same version of cryptography. Tested on a deployed EL7 with new packages that st2 key set/get with encrypted secrets worked.

amanda11:
Tested 3.3 EL7 -> 3.4dev EL7 upgrade using new package. Manuallly install of libselinux-python3 before upgrade. Upgrade successful and self-check passed. 
`$ st2 --version
st2 3.4dev (632a07e), on Python 3.6.8`

armab (approved):
Worked flawlessly for me when trying the el7 packages manually.

Great work :+1:

PR #675, opened 2020-11-20
Install python3-devel from rhel-server-optional-rpms

amanda11:
On CentOS 7 python3-devel is available from enabled repos. In RHEL 7 then the python3-devel is in the rhel-server-optional-rpms repo (name differs depending on which version of RHEL 7 on AWS, or standalone server). The optional-rpms repo is disabled by default.
Add commands to bash installer to check if python3-devel is available from enabled repos, and if not then install it with temporary enablement of repos. 

Part of StackStorm/discussions#54


armab (approved):
:+1: Thanks for the fix!

PR #678, opened 2020-11-27
Remove libselinux-python3 for EL7

amanda11:
Remove libselinux-python3 as not needed for ST2 on EL7 (when running python3)


armab (approved):
:+1:

PR #680, opened 2020-12-10
Remove st2debug component

amanda11:
Companion to https://github.com/StackStorm/st2/pull/5103. 
The only item in st2debug component is the submit-debug-info component, so when this is removed the st2debug component also is removed. So remove it from the packages.


armab (approved):
Thanks a lot for keeping st2 clean from the outdated legacy parts! :+1:

PR #681, opened 2020-12-14
Ubuntu Xenial Warning: Block python3.6 PPA installation waiting on user's agreement

armab:
Per previous discussion in https://github.com/StackStorm/st2-packages/pull/679, adding an option to install 3rd party PPA based on user's input.

On Ubuntu Xenial the installer would stop, show insecure warning and wait for user's input to continue:

>WARNING!
>python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it.
>We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS.
>
>Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa.
>By continue you're aware of security risk associated with using unofficial 3rd party PPA repository.

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -121,11 +121,11 @@ setup_args() {
     if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
       echo ""
       echo "WARNING!"
-      echo "python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it."
-      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
+      echo "The python3.6 package is a required dependency for the StackStorm st2 package but that is not installable from any of the default Ubuntu 16.04 repositories."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS. Support for Ubuntu 16.04 will be removed with future StackStorm versions."
       echo ""
-      echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
-      echo "By continuing you're aware of security risk associated with using unofficial 3rd party PPA repository."
+      echo "Alternatively we'll try to add python3.6 from the 3rd party 'deadsnakes' repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continuing you are aware of the support and security risks associated with using unofficial 3rd party PPA repository, and you understand that StackStorm does NOT provide ANY support for python3.6 packages on Ubuntu 16.04."
       echo ""
       read -p "Press [y] to continue or [n] to cancel adding it: " choice
       case "$choice" in
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -112,6 +112,34 @@ setup_args() {
         exit 1
     fi
   fi
+
+  # Python 3.6 package is not available in Ubuntu Xenial
+  # Installer can add it via 3rd party PPA based on user agreement
+  if [[ "$SUBTYPE" = 'xenial' ]]; then
+    sudo apt-get update > /dev/null 2>/dev/null
+    # check if python3.6 is available
+    if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
+      echo ""
+      echo "WARNING!"
+      echo "python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
+      echo ""
+      echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continue you're aware of security risk associated with using unofficial 3rd party PPA repository."
+      echo ""
+      read -p "Press [y] to continue or [n] to cancel adding it: " choice
+      case "$choice" in
+        y|Y )
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys F23C5A6CF475977595C89F51BA6932366A755776
+          echo "deb http://ppa.launchpad.net/deadsnakes/ppa/ubuntu xenial main" | sudo tee /etc/apt/sources.list.d/deadsnakes-ubuntu-ppa-xenial.list
+          ;;
+        * )
+          echo "python3.6 PPA installation aborted"
+          exit 1
+          ;;
+      esac
+    fi
+  fi
 }
 
 
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -125,7 +125,7 @@ setup_args() {
       echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
       echo ""
       echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
-      echo "By continue you're aware of security risk associated with using unofficial 3rd party PPA repository."
+      echo "By continuing you're aware of security risk associated with using unofficial 3rd party PPA repository."
       echo ""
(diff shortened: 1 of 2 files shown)


scripts/st2bootstrap-deb.sh
```diff
@@ -112,6 +112,34 @@ setup_args() {
...
+    if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
+      echo ""
+      echo "WARNING!"
+      echo "python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
+      echo ""
+      echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continue you're aware of security risk associated with using unofficial 3rd party PPA repository."
+      echo ""
+      read -p "Press [y] to continue or [n] to cancel adding it: " choice
```
armab, line 130: I decided to not provide a dedicated installation flag for Ubuntu Xenial installer as everything could be already done with `yes |`  or `echo 'y' |` from the user side anyways.

scripts/st2bootstrap-deb.template.sh
```diff
@@ -104,6 +104,34 @@ setup_args() {
...
+    sudo apt-get update > /dev/null 2>/dev/null
+    # check if python3.6 is available
+    if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
+      echo ""
+      echo "WARNING!"
+      echo "python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
+      echo ""
+      echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continue you're aware of security risk associated with using unofficial 3rd party PPA repository."
```
amanda11, line 120: Minor reword suggestion from
"By continue you're aware of"
to
"By continuing you're aware of "

amanda11 (commented):
LGTM - just a minor comment on wording in warning.

Also, are we adding the --use-deadsnakes-repo-for-python3.6 option in under a separate PR to avoid the question?

scripts/st2bootstrap-deb.sh
armab: @amanda11 ^^ I believe that should answer your question about running installer in non-interactive mode.

Even if we add `--install-py3-from-insecure-ppa` flag for extra security, - it makes little sense as users would be able to use `yes | curl|bash` or `echo y | curl|bash` to workaround it.

scripts/st2bootstrap-deb.template.sh
armab: Thanks, fixed!

scripts/st2bootstrap-deb.template.sh
```diff
@@ -104,6 +104,34 @@ setup_args() {
...
+    sudo apt-get update > /dev/null 2>/dev/null
+    # check if python3.6 is available
+    if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
+      echo ""
+      echo "WARNING!"
+      echo "python3.6 package is required as a dependency for StackStorm and none of your software repositories provide it."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS."
+      echo ""
+      echo "Alternatively we'll try to add 3rd party python3.6 APT repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continuing you're aware of security risk associated with using unofficial 3rd party PPA repository."
```
blag, line 120:
```suggestion
      echo "The python3.6 package is a required dependency for the StackStorm st2 package but that is not installable from any of the default Ubuntu 16.04 repositories."
      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS. Support for Ubuntu 16.04 will be removed with StackStorm 3.5."
      echo ""
      echo "We will use python3.6 from the 3rd party 'deadsnakes' PPA: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
      echo ""
      echo "By continuing you are aware of the support and security risks associated with using unofficial 3rd party PPA repository, and you understand that StackStorm does NOT provide ANY support for python3.6 packages on Ubuntu 16.04."
      echo ""
      echo "To bypass this check in the future and automatically install and use the deadsnakes PPA, you can run this script with:"
      echo ""
      echo "echo 'y' | $(0) $@"
```

scripts/st2bootstrap-deb.template.sh
```diff
@@ -104,6 +104,34 @@ setup_args() {
...
+      echo ""
+      read -p "Press [y] to continue or [n] to cancel adding it: " choice
+      case "$choice" in
+        y|Y )
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys F23C5A6CF475977595C89F51BA6932366A755776
+          echo "deb http://ppa.launchpad.net/deadsnakes/ppa/ubuntu xenial main" | sudo tee /etc/apt/sources.list.d/deadsnakes-ubuntu-ppa-xenial.list
```
blag, line 126: Instead of doing this manually, I would use the `add-apt-repository` command, which will take care of this for you:

```bash
sudo apt-get install --yes software-properties-common
sudo add-apt-repository --yes ppa:deadsnakes/ppa
sudo apt-get update
```

That should reduce our maintenance burden if the PPA changes at all.

blag (changes requested):
Some suggestions.

scripts/st2bootstrap-deb.template.sh
armab: I tried that, but installing `software-properties-common` is not what I wanted in the foremost pre-req place of the installer script. See `setup_args` for more context.

I decided to go with adding the string manually as we already do that in other places.
I don't expect us to ever change this repository to another, but rather remove it alltogether in a half of the year.

scripts/st2bootstrap-deb.sh
amanda11: OK - so we just need to update the E2E tests to install with an echo "y" | installer then?
Fine with me.

scripts/st2bootstrap-deb.template.sh
blag: It's a big enough check that we could probably put it in its own Bash function/step.

scripts/st2bootstrap-deb.sh
blag: Yeah that's fine with me as well.

scripts/st2bootstrap-deb.template.sh
armab: Thanks, I took some of the fixes you suggested: https://github.com/StackStorm/st2-packages/pull/681/commits/aaeea508253b5c41314f932635928988ecb25d86

I won't include a command example with `y` and I'd prefer to avoid a wall of text to keep the information amount somewhat balanced. I liked more explicit wording around security risks, - that's very helpful and more clear :+1:

scripts/st2bootstrap-deb.template.sh
```diff
@@ -104,6 +104,34 @@ setup_args() {
...
+      echo "WARNING!"
+      echo "The python3.6 package is a required dependency for the StackStorm st2 package but that is not installable from any of the default Ubuntu 16.04 repositories."
+      echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS. Support for Ubuntu 16.04 will be removed with future StackStorm versions."
+      echo ""
+      echo "Alternatively we'll try to add python3.6 from the 3rd party 'deadsnakes' repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
+      echo "By continuing you are aware of the support and security risks associated with using unofficial 3rd party PPA repository, and you understand that StackStorm does NOT provide ANY support for python3.6 packages on Ubuntu 16.04."
+      echo ""
+      read -p "Press [y] to continue or [n] to cancel adding it: " choice
+      case "$choice" in
+        y|Y )
```
blag, line 124: Users might also use `yes`, `Yes`, or `YES`.

blag (approved):

PR #691, opened 2021-02-23
Fix PPA check else

amanda11:
When the insecure_ppa flag was set, it was skipping the install of deadsnakes ppa, due to incorrect else condition


amanda11:
Related to https://github.com/StackStorm/discussions/issues/66

armab (approved):
:100: 

Thanks for finding this!

Kami:
We should hook up shell check or something like that to those scripts - I imagine shell check should be able to catch such error automatically.

PR #687, opened 2021-02-17
[BUGFIX] Use pip==20.0.2 to match core st2 pin

cognifloyd:
We have been running tests with pip==20.0.2 in v3.3.0 and v3.2.0. But we have still been installing pip 19 in production.

That is very likely to cause installation errors for packs in production that would not exist during testing (eg if a pack's dep uses a pip feature that is not available in pip 19, which is becoming more common).

See these commits in StackStorm/st2:
- a6c4b89373e01f1c38396ce65d9dda3e3be0e842
- c9f6638c491383e7587ab44f76673c81805aaa13
- e1804fb666cd9d41a77a1e776fe6bb0a1e1b949b
- 40827e7fdb04588bfc12ce4da7c7a9bc7e9c4ae3

We have not been testing with anything newer than pip 20.0.2, however, so upgrading to a newer version is dangerous. Since we've already been testing with 20.0.2, we should start with making pip consistent across all of our infrastructure and then we can worry about upgrading.

Here is a list of places that need to be updated when we do upgrade pip:
https://github.com/StackStorm/st2/wiki/Where-all-to-update-pip-and-or-virtualenv

PR #695, opened 2021-03-04
Install pip 20.0.2 in rpm package venv

Kami:
This pull request updates rpmspec for package venv so we use the same pip version as we use elsewhere and in deb packages.

NOTE: I'm trying to get build with orjson to work but it's a pita since it looks like we are using python2.7 to build dh virtualenv so it will need more work :/


Kami:
Just noticed it's duplicate of #687.

PR #697, opened 2021-03-28
[WIP] Speed up builds by caching /tmp/wheelhouse directory

Kami:
This pull request is an attempt to speed up builds by caching ``/tmp/wheelhouse`` directory.

This way we can avoid re-building wheels for all the dependencies for each run which is quite slow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
           exclusive: true   # default - true
           # Temporary workaround for Circle CI issue
           # https://discuss.circleci.com/t/setup-remote-docker-connection-failures/26434
-          version: 18.05.0-ce 
+          version: 18.05.0-ce
       - run:
           name: Ensure installation scripts are synced with their templates
           command: make .generated-files-check
@@ -70,21 +70,54 @@ jobs:
       - run:
           name: Pull dependent Docker Images
           command: .circle/docker-compose2.sh pull ${DISTRO}
+      - restore_cache:
+          name: Restore Wheelhouse Cache
+          keys:
+            - wheelhouse-{{ checksum "/tmp/distro-version" }}-{{ checksum "/tmp/st2-version" }}
+            - wheelhouse-
       - run:
           name: Build the ${DISTRO} Packages
           command: |
-            .circle/docker-compose2.sh build ${DISTRO}
-            # Once build container finishes we can copy packages directly from it
+            # Create necessary directories
             mkdir -p ~/st2-packages/build/${DISTRO}/log/
+            mkdir -p ~/st2-packages/wheelhouse/
+
+            # List volume content to see if there is any cached dasta
+            echo "local wheelhouse cache directory content"
+            echo ""
+            ls -la ~/st2-packages/wheelhouse/
+            du -hs ~/st2-packages/wheelhouse/
+            echo ""
+
+            # Copy over cached wheelhouse packages (if any exist)
+            docker cp ~/st2-packages/wheelhouse/. st2-packages-vol:/tmp/wheelhouse
+
+            # List volume content to see if there is any cached data
+            echo "docker volume wheelhouse cache directory content"
+            echo ""
+            docker run --rm -i -v=st2-packages-vol:/tmp/wheelhouse busybox find /tmp/wheelhouse
+            echo ""
+
+            # Run the build
+            .circle/docker-compose2.sh build ${DISTRO}
+
+            # Once build container finishes we can copy packages directly from it
+
+            # Copy built packages
             docker cp st2-packages-vol:/root/build/. ~/st2-packages/build/${DISTRO}
-#      # TODO: It works! (~0.5-1min speed-up) Enable CircleCI2.0 cache for pip and wheelhouse later
-#      - run:
-#          name: Build the ${DISTRO} Packages 2nd time (compare with pip/wheelhouse cached)
-#          command: |
-#            .circle/docker-compose2.sh build ${DISTRO}
-#            # Once build container finishes we can copy packages directly from it
-#            docker cp st2-packages-vol:/root/build /tmp/st2-packages
-#          working_directory: ~/st2-packages
+
+            # Copy all the files from /tmp/wheelhouse so we can cache it and substantially speed
+            # up subsequent builds
+            docker cp st2-packages-vol:/tmp/wheelhouse/. ~/st2-packages/wheelhouse/
+
+            echo "wheelhouse cache directory content"
+            ls -la ~/st2-packages/wheelhouse/
+            du -hs ~/st2-packages/wheelhouse/
+      - save_cache:
+          name: Store Wheelhouse Cache
+          key: wheelhouse-{{ checksum "/tmp/distro-version" }}-{{ checksum "/tmp/st2-version" }}
+          paths:
+            - ~/st2-packages/wheelhouse/
       - run:
           name: Test the Packages
           command: .circle/docker-compose2.sh test ${DISTRO}
--- a/.circle/buildenv_st2.sh
+++ b/.circle/buildenv_st2.sh
@@ -56,5 +56,8 @@ re="\\b$DISTRO\\b"
 ST2_CIRCLE_URL=${CIRCLE_BUILD_URL}
 
 write_env ST2_GITURL ST2_GITREV ST2PKG_VERSION ST2PKG_RELEASE DISTRO TESTING ST2_CIRCLE_URL
-
 cat ~/.buildenv
+
+# Workaround for cache key since we can't reference arbitrary environment variables for cache keys
+echo "${DISTRO}" > /tmp/distro-version
+echo "${ST2PKG_VERSION}" > /tmp/st2-version


Kami:
@armab I didn't dig into TODO yet since I'm trying to get caching to work, but do you have any ideas why RHEL 8 build is so much slower compared to other distros?

Kami:
Also, one of the "bottlenecks" is re-building wheels for all the git repo dependencies on each run (orquesta, ldap backend, rbac, etc.).

We should also figure out a way to cache those since there is no need to re-build it if the actual repo / git sha doesn't changes.

armab:
Yeah, EL8 takes the longest time to build since the beginning when it was added, we didn't figure out why.

Kami:
I added some additional timing instrumentation and it seems one part of ``rpmbuild`` step just takes a very long time (~6 minutes).

Screenshot from 2021-03-28 23-12-17

Need to figure out to see if we can bump rpmbuild verbosity since it may offer more insight.

Kami:
Here is the verbose debug output if someone wants to dig in - https://circleci.com/api/v1.1/project/github/StackStorm/st2-packages/3815/output/110/3?file=true&allocation-id=6060f421bf121a40f4490233-3-build%2F203B6FE6.

Kami:
From slack (https://stackstorm-community.slack.com/archives/CSBELJ78A/p1616971134338600).

EL7 package:

```bash
rpm -q -i --provides ~/Downloads/st2-3.5dev-28.x86_64*1*.rpm
Name        : st2
Version     : 3.5dev
Release     : 28
Architecture: x86_64
Install Date: (not installed)
Group       : System/Management
Size        : 141683575
License     : ASL 2.0
Signature   : (none)
Source RPM  : st2-3.5dev-28.src.rpm
Build Date  : ned 28 mar 2021 23:30:57
Build Host  : f7917c6af893
URL         : https://github.com/StackStorm/st2
Summary     : StackStorm all components bundle
Description :
  Package is full standalone stackstorm installation including
  all components
config(st2) = 3.5dev-28
libffi-806b1a9d.so.6.0.4()(64bit)
st2 = 3.5dev-28
st2(x86-64) = 3.5dev-28
```


EL8 package:

```bash
kami ~/w/stackstorm/st2-packages (git:cache_tmp_wheelhouse)$ rpm -q -i --provides ~/Downloads/st2-3.5dev-28.x86_64.rpm
Name        : st2
Version     : 3.5dev
Release     : 28
Architecture: x86_64
Install Date: (not installed)
Group       : System/Management
Size        : 141991176
License     : ASL 2.0
Signature   : (none)
Source RPM  : st2-3.5dev-28.src.rpm
Build Date  : ned 28 mar 2021 23:37:42
Build Host  : ccdd67b8b527
URL         : https://github.com/StackStorm/st2
Summary     : StackStorm all components bundle
Description :
  Package is full standalone stackstorm installation including
  all components
config(st2) = 3.5dev-28
libffi-806b1a9d.so.6.0.4()(64bit)
python3.6dist(argcomplete) = 1.12.2
python3.6dist(click) = 7.1.2
python3.6dist(decorator) = 4.4.2
python3.6dist(filelock) = 3.0.12
python3.6dist(msgpack) = 1.0.2
python3.6dist(orjson) = 3.5.1
python3.6dist(pika) = 1.2.0
python3.6dist(prettytable) = 2.1.0
python3.6dist(prompt-toolkit) = 1.0.15
python3.6dist(python-json-logger) = 2.0.1
python3.6dist(python-statsd) = 2.1.0
python3.6dist(retrying) = 1.3.3
python3.6dist(sseclient-py) = 1.7
python3.6dist(webob) = 1.8.5
python3.6dist(wrapt) = 1.12.1
python3dist(argcomplete) = 1.12.2
python3dist(click) = 7.1.2
python3dist(decorator) = 4.4.2
python3dist(filelock) = 3.0.12
python3dist(msgpack) = 1.0.2
python3dist(orjson) = 3.5.1
python3dist(pika) = 1.2.0
python3dist(prettytable) = 2.1.0
python3dist(prompt-toolkit) = 1.0.15
python3dist(python-json-logger) = 2.0.1
python3dist(python-statsd) = 2.1.0
python3dist(retrying) = 1.3.3
python3dist(sseclient-py) = 1.7
python3dist(webob) = 1.8.5
python3dist(wrapt) = 1.12.1
st2 = 3.5dev-28
st2(x86-64) = 3.5dev-28
```

In short, I don't think we need all that "requires" Python cruft for EL 8 packages since we bundle virtualenv and that requires section looks wrong anyway (it's missing a bunch of dependencies).

Kami:
Alright, looks like a victory - from ~10-13 minutes for EL8 build step down to ~4 minutes.

It appears my workaround / fix worked. The "meat" of it is here https://github.com/StackStorm/st2-packages/pull/697/files#diff-f0d938587e6d1cdabf2257a39f1070bac432a2b3242ee1c95cdd84454549a53cR28. Basically, I re-defined those macros since we don't want to run those steps.

The build is much faster now and more in line with EL7 where all that requires cruft which we don't depend on is not there since we are not a traditional RPM for a Python library and we bundle virtual environment with all the deps.

Kami:
Do we still need to build ``cryptography`` from scratch in EL build?

I removed that change and it appears to be working fine.

I assume it may have something to do with old version of pip in the past and perhaps that version not supporting newer pre-built wheel formats used by ``cryptography`` or similar.

packages/st2/rpm/st2.spec
```diff
@@ -25,6 +25,24 @@ BuildRequires: python3-devel
...
+# Python dependencies. That process / script works by recursively scanning all the package Python
+# dependencies which is very slow (5-6 minutes).
+# Our package bundles virtualenv with all the dependendencies and doesn't rely on this metadata
+# so we skip that step to vastly speed up the build.
+# Technically we also don't Require or Provide any of those libraries auto-detected by that script
+# because those are only used internally inside a package specific virtual environment.
+# Same step also does not run on EL7.
+# See https://github.com/StackStorm/st2-packages/pull/697#issuecomment-808971874 and that PR for
+# more details.
+# That issue was found by enabling rpmbuild -vv flag.
```
armab, line 39: Can you show an example of `Requires:` record before and after this change?

There might be important dependencies in there like shared libraries that are indeed required for the package to work properly.

armab: Here is how `Requires` currently looks like:
```

Requires

    /bin/bash
    /bin/sh
    /bin/sh
    /opt/stackstorm/st2/bin/python
    /usr/bin/env
    bash
    config(st2) = 3.4.1-2
    git
    ld-linux-x86-64.so.2()(64bit)
    ld-linux-x86-64.so.2(GLIBC_2.3)(64bit)
    libc.so.6(GLIBC_2.14)(64bit)
    libcrypto.so.1.1()(64bit)
    libdl.so.2()(64bit)
    libffi-806b1a9d.so.6.0.4()(64bit)
    libffi-devel
    libgcc_s.so.1()(64bit)
    liblber-2.4.so.2()(64bit)
    libldap_r-2.4.so.2()(64bit)
    libm.so.6()(64bit)
    libpthread.so.0()(64bit)
    libpthread.so.0(GLIBC_2.2.5)(64bit)
    libpython3.6m.so.1.0()(64bit)
    libstdc++.so.6()(64bit)
    libutil.so.1()(64bit)
    libyaml-0.so.2()(64bit)
    openssh-clients
    openssh-server
    openssl-devel
    pam
    python3-devel
    rtld(GNU_HASH)
    setup

```
https://packagecloud.io/StackStorm/stable/packages/el/8/st2-3.4.1-2.x86_64.rpm

Kami: Yeah, see the comment here - https://github.com/StackStorm/st2-packages/pull/697#issuecomment-808971874 (for ``Provides``) and the output below.

Requires EL 8 before:

```bash
rpm -iqp --requires ~/Downloads/st2-3.5dev-28.x86_64.rpm
/bin/bash
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/opt/stackstorm/st2/bin/python
/usr/bin/env
bash
config(st2) = 3.5dev-28
git
ld-linux-x86-64.so.2()(64bit)
ld-linux-x86-64.so.2(GLIBC_2.3)(64bit)
libc.so.6()(64bit)
libc.so.6(GLIBC_2.13)(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.3)(64bit)
libc.so.6(GLIBC_2.3.2)(64bit)
libc.so.6(GLIBC_2.3.4)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libc.so.6(GLIBC_2.6)(64bit)
libc.so.6(GLIBC_2.7)(64bit)
libcrypto.so.1.1()(64bit)
libdl.so.2()(64bit)
libffi-806b1a9d.so.6.0.4()(64bit)
libffi-devel
libgcc_s.so.1()(64bit)
libgcc_s.so.1(GCC_3.0)(64bit)
liblber-2.4.so.2()(64bit)
libldap_r-2.4.so.2()(64bit)
libm.so.6()(64bit)
libm.so.6(GLIBC_2.2.5)(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libpython3.6m.so.1.0()(64bit)
libstdc++.so.6()(64bit)
libstdc++.so.6(CXXABI_1.3)(64bit)
libstdc++.so.6(GLIBCXX_3.4)(64bit)
libutil.so.1()(64bit)
openssh-clients
openssh-server
openssl-devel
pam
python3-devel
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
rtld(GNU_HASH)
setup
```

Requires EL 8 after:


```bash
kami ~/w/stackstorm/st2-packages (git:cache_tmp_wheelhouse)$ rpm -iqp --requires ~/Downloads/st2-3.5dev-28.x86_64*2*.rpm
/bin/bash
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/opt/stackstorm/st2/bin/python
/usr/bin/env
bash
config(st2) = 3.5dev-28
git
ld-linux-x86-64.so.2()(64bit)
ld-linux-x86-64.so.2(GLIBC_2.3)(64bit)
libc.so.6()(64bit)
libc.so.6(GLIBC_2.13)(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.3)(64bit)
libc.so.6(GLIBC_2.3.2)(64bit)
libc.so.6(GLIBC_2.3.4)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libc.so.6(GLIBC_2.6)(64bit)
libc.so.6(GLIBC_2.7)(64bit)
libcrypto.so.1.1()(64bit)
libdl.so.2()(64bit)
libffi-806b1a9d.so.6.0.4()(64bit)
libffi-devel
libgcc_s.so.1()(64bit)
libgcc_s.so.1(GCC_3.0)(64bit)
liblber-2.4.so.2()(64bit)
libldap_r-2.4.so.2()(64bit)
libm.so.6()(64bit)
libm.so.6(GLIBC_2.2.5)(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libpython3.6m.so.1.0()(64bit)
libstdc++.so.6()(64bit)
libstdc++.so.6(CXXABI_1.3)(64bit)
libstdc++.so.6(GLIBCXX_3.4)(64bit)
libutil.so.1()(64bit)
openssh-clients
openssh-server
openssl-devel
pam
python3-devel
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
rtld(GNU_HASH)
setup
```

In short, that change only affect ``Provides`` for Python packages and ``Requires`` itself is unchanged.

Kami: @armab 

> As for Provides, there's a lot of unneeded noise there, if a cleanup helps in a build speedup that's great indeed.

As per my comment here (https://github.com/StackStorm/st2-packages/pull/697#issuecomment-808974651), I don't think it's just noise.

It's actually incorrect and broken since ``Provides`` claims that package provides a bunch of other Python libraries which is not correct - those libraries are not provided by the package and are just used internally in the virtual environment.

Likely no one was affected by this bug yet since package is usually installed on isolated VMs (and as per comment above, to be "affected by it" would require some specific steps / dependencies to be installed).

packages/st2/rpm/st2.spec
```diff
@@ -25,6 +25,24 @@ BuildRequires: python3-devel
...
+%undefine __pythondist_provides
+%undefine __pythondist_requires
+%undefine __python_provides
+%undefine __python_requires
+%endif
```
armab, line 44: I'm fine with the `Provides` change, that's a great cleanup, especially if it improves the build timing! :+1: 

I'm more worried about the Requires section. If it doesn't touch the Requires, do we really need to undefine those?

```suggestion
%undefine __pythondist_provides
%undefine __python_provides
%endif
```

Kami: I don't think it should hurt anything by having it there since we don't explicitly depend on any Python dependency via yum package (we install those manually inside virtualenv).

In fact, I would argue it's probably better to have it there to detect any issues where that macro may declare dependencies which we don't want by doing "magic introspection of all the Python deps" (similar to what happened with ``Provides`` and went undetected for so long).

In short, it only affects Python dependencies and if it indeed breaks something - our end to end tests should be there to catch that. If they don't, well, we should fix that.

.circleci/config.yml
```diff
@@ -70,21 +70,57 @@ jobs:
...
+            # Copy all the files from /tmp/wheelhouse so we can cache it and substantially speed
+            # up subsequent builds
+            docker cp st2-packages-vol:/tmp/wheelhouse/. ~/st2-packages/wheelhouse/
+
+            # echo "wheelhouse cache directory content"
+            # Uncomment if you are troubleshooting wheelhouse caching issues
+            # ls -la ~/st2-packages/wheelhouse/
+            # du -hs ~/st2-packages/wheelhouse/
+      - save_cache:
+          name: Store Wheelhouse Cache
```
armab, line 120: :+1:

.circleci/config.yml
```diff
@@ -70,21 +70,57 @@ jobs:
...
+
+            # Run the build
             .circle/docker-compose2.sh build ${DISTRO}
+
             # Once build container finishes we can copy packages directly from it
-            mkdir -p ~/st2-packages/build/${DISTRO}/log/
+
+            # Copy built packages
             docker cp st2-packages-vol:/root/build/. ~/st2-packages/build/${DISTRO}
-#      # TODO: It works! (~0.5-1min speed-up) Enable CircleCI2.0 cache for pip and wheelhouse later
```
armab, line 80: Good to find this TODO comment. This takes us back in time, 3 years ago :)

packages/st2/rpm/st2.spec
```diff
@@ -25,6 +25,24 @@ BuildRequires: python3-devel
...
+# Technically we also don't Require or Provide any of those libraries auto-detected by that script
+# because those are only used internally inside a package specific virtual environment.
+# Same step also does not run on EL7.
+# See https://github.com/StackStorm/st2-packages/pull/697#issuecomment-808971874 and that PR for
+# more details.
+# That issue was found by enabling rpmbuild -vv flag.
+%undefine __pythondist_provides
+%undefine __pythondist_requires
+%undefine __python_provides
+%undefine __python_requires
```
armab, line 43:
```suggestion
%undefine __pythondist_provides
%undefine __python_provides
```
we shouldn't touch `Requires`

packages/st2/rpm/st2.spec
armab: I'm not convinced of changing the behavior for `Requires` and trying to catch that via e2e tests later.

Did `%undefine __pythondist_requires` and `%undefine __python_requires` affect the build timing or was it only `Provides` change that really helped with the speedup?

packages/st2/rpm/st2.spec
armab: If you believe how the `Provides` was generated is a bug in the package building, do we need to also include the same change for any RPM package, not just EL8?

packages/st2/rpm/st2.spec
Kami: Yes, both of the changes help.

First I accidentally only added ``requires`` and the build was faster, but not as much as when also adding provides override.

And I think end to end tests should be there exactly for reasons like that - to catch any kind of "end to end" issues with the packages. Including invalid package metadata or similar.

Technically, we could also add some quick assertions to this repo, but this could be a slippery slope since you could argue end to end tests are a better fit for catching issues like that.

packages/st2/rpm/st2.spec
Kami: Per the comment - the change only affects EL 8 and not EL 7.

I didn't dig in that deep to figure out exactly why, but I assume it's due to the differences how that script works which automatically finds all the transitive Python package dependencies and declares them under ``Provides``.

Having said that, I also added a change to use that section on EL 7 as well for just in case.

packages/st2/rpm/st2.spec
Kami: But again, if you feel strongly about it, I'm fine with adding some quick "grep based" assertions to the build job which verifies that some specific / minimum dependencies are defined as part of "Requires".

packages/st2/rpm/st2.spec
armab: That makes sense, indeed it's all good for EL7: https://packagecloud.io/StackStorm/stable/packages/el/7/st2-3.4.1-2.x86_64.rpm

packages/st2/rpm/st2.spec
armab: Alright,
Let me check the timing including the `Provides` and excluding the `Requires` change to understand your point and tradeoffs better.

armab (approved):
I'm totally fine with the `Provides` changes.

Looks like `Requires` change removes ~2mins of the build overhead for EL8.
I agree that's a reasonable price to pay for the controversial to me `%undefine __pythondist_requires` and `%undefine __python_requires` addition.

packages/st2/rpm/st2.spec
armab: Looks good to me, approved it. Thanks for research.

Kami:
@armab Thanks for the review.

PR #698, opened 2021-04-08
Add redis to install scripts

m4dcoder:
The coordinator service is required in most cases for running workflows. Install redis-server and configure redis as the default coordinator backend used by tooz. This will avoid a number of workflow issues, especially for new StackStorm users.

### DIFF
--- a/.circle/docker-compose.sh
+++ b/.circle/docker-compose.sh
@@ -11,7 +11,7 @@ set -e
 # Source the build environment defintion (details in buildenv.sh)
 . ~/.buildenv
 
-# Used for `RABBITMQHOST` `MONGODBHOST`, see docker-compose.override.yml
+# Used for `RABBITMQHOST`, `MONGODBHOST`, and `REDISHOST` see docker-compose.override.yml
 HOST_IP=$(ifconfig docker0 | grep 'inet addr' | awk -F: '{print $2}' | awk '{print $1}')
 
 set -x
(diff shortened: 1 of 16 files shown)


m4dcoder:
This PR pairs with https://github.com/StackStorm/st2/pull/5226.

armab:
Looks good :+1: 


We'll also need to include Redis in CI, Tests and docker-compose:
* https://github.com/StackStorm/st2-packages/blob/master/docker-compose.override.yml
* https://github.com/StackStorm/st2-packages/blob/master/docker-compose.yml
* https://github.com/StackStorm/st2-packages/blob/master/docker-compose.circle2.yml
* https://github.com/StackStorm/st2-packages/blob/master/docker-compose.circle.yml
* https://github.com/StackStorm/st2-packages/blob/master/scripts/generate_st2_config.sh
* https://github.com/StackStorm/st2-packages/blob/master/rake/spec/default/60-st2_all-services-ok_spec.rb#L8-L11

Kami:
As @armab said, LGTM< just need to make sure we update all other affected places (sadly there are quite a lot of those and some of those we will only stumble upon in the future - e.g. st2web also uses docker compose setup for end to end tests, etc.).

armab (approved):
Looks great with updating it in all the places! :+1: 
Additional thanks for adding CircleCI smoke tests here to run Redis in Docker.

---

We'll also need for `stackstorm/st2` to add the default Redis `[coordination]` string in the default `st2.conf`:
https://github.com/StackStorm/st2/blob/master/conf/st2.package.conf similar to other backends like Mongo and RabbitMQ.

PR #699, opened 2021-04-15
Unpin rabbitmq and use minimal erlang

### LINKED ISSUE #694: rabbitMQ version on EL8 hardcoded
The EL8 installation of ST2 pulls rabbitMQ from the packagecloud rabbitmq repository. The latest version requires an erlang version that is greater than the version that is pulled from the EL8/EPEL repositories.

This was resolved for 3.4 by pinning the rabbitMQ version to the last version that used the older erlang in PR https://github.com/StackStorm/st2-packages/pull/693.

It would be desirable to be able to pick up bug-fixes, so if rabbitMQ is available in epel then take from there, or look to see if can fetch erlang version that is compatible with latest rabbitMQ.

amanda11:
Stop pinning version of rabbitmq on EL8.

Install latest from packagecloud, and therefore as that requires newer version of erlang, install the minimal erland provided by rabbitmq - as per instructions at https://github.com/rabbitmq/erlang-rpm

Resolves #694

### DIFF
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -22,10 +22,6 @@ ST2_PKG='st2'
 ST2WEB_PKG='st2web'
 ST2CHATOPS_PKG='st2chatops'
 
-# WORKAROUND - pin rabbitmq version for the moment due to 3.8.13 requiring
-# erlang version that is not in EPEL or CentOS repos
-RABBITMQ_VERSION=3.8.12
-
 is_rhel() {
   return $(cat /etc/os-release | grep 'ID="rhel"')
 }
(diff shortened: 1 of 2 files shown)


cognifloyd (approved):

armab (approved):
Thanks for tracking this!

Kami:
LGTM.

Do we also need to update docker compose setup (if it doesn't already use the latest version)?

Kami:
And I believe st2vagrantdev may also need to be updated (I can handle that if needed).

amanda11:
> LGTM.
> 
> Do we also need to update docker compose setup (if it doesn't already use the latest version)?

I think st2-docker is ok, as it was only EL8 that we pinned rabbitmq version -as we needed a quick fix when it broke at release time.

PR #703, opened 2021-05-24
Added Password validation check

shivani-orch:
We have fixed this OWASP issue - The application accepts very weak passwords like 'test'.
Strong password policy has been implemented - min8 char password consist of letter, special char, number etc


amanda11:
Does this only enforce password rule on stdin, not via the password parameter?

Has the password policy been agreed? 

As, The password would not allow the default password that is in the docs and other places - as the default password has special char but not digit.


Therefore before this password policy was implemented, all other places in the different repos that use the current default password will need updating first - to prevent breakages.

PR #704, opened 2021-06-03
Remove xenial

amanda11:


amanda11:
Removed xenial. It is still mentioned in st2_bootstrap.sh as I believe that is global across versions, and so we can't remove it, as it would prevent installation of 3.4 on xenial.

blag (approved):
LGTM.

scripts/st2bootstrap-deb.template.sh
```diff
@@ -111,31 +111,6 @@ setup_args() {
     fi
   fi
 
-  # Python 3.6 package is not available in Ubuntu Xenial
-  # Installer can add it via 3rd party PPA based on user agreement
-  if [[ "$SUBTYPE" = 'xenial' ]]; then
```
Kami, line 116: We could leave this if here and just print an error that Xenial is not supported since v3.5.0 in case someone still tries to run new installer script on Xenial.

scripts/st2bootstrap-deb.sh
```diff
@@ -119,31 +119,6 @@ setup_args() {
...
-    sudo apt-get update > /dev/null 2>/dev/null
-    # check if python3.6 is available
-    if (! apt-cache show python3.6 2> /dev/null | grep 'Package:' > /dev/null); then
-      if [[ "$U16_ADD_INSECURE_PY3_PPA" = "0" ]]; then
-        echo ""
-        echo "WARNING!"
-        echo "The python3.6 package is a required dependency for the StackStorm st2 package but that is not installable from any of the default Ubuntu 16.04 repositories."
-        echo "We recommend switching to Ubuntu 18.04 LTS (Bionic) as a base OS. Support for Ubuntu 16.04 will be removed with future StackStorm versions."
-        echo ""
-        echo "Alternatively we'll try to add python3.6 from the 3rd party 'deadsnakes' repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa."
```
armab, line 134: Hell yeah :+1:

armab (approved):
Looks good! :+1:

scripts/st2bootstrap-deb.template.sh
amanda11: Thoughts on this comment... @Kami @armab @blag 

I don't mind either way - but we'd have to  put quite a few different - if "xenial" do this checks back in. Plus the 3.5 and "unstable" xenial packages are no-longer built daily - so what they'd get is an old version anyway.

As I understand the install process then if they install 3.4 xenial they will get the "master" st2_bootstrap - but they'd get the st2bootstrap-deb.sh from the 3.4 branch - and so would have all the stuff there.

So to me, it's safer to just remove the ifs, and say that we only support bionic/focal etc.

PR #705, opened 2021-06-07
Move to node 14

amanda11:


amanda11:
Note. The scripts will not work until the st2chatops has been promoted from staging-unstable to unstable. Tested install manually on EL7 with the staging-unstable repo.

armab (approved):
:+1:

amanda11:
> 👍
> 
> I think we'll also need upgrade notes in st2docs for the breaking change like this.

Technically st2ch

Issue #686, opened 2021-02-05
The st2 port is configured in the config file. But directly fixed in files such as service, socket, client, etc. 


huzunhao:
If you find an issue in packages, please file an issue and we'll have a look as soon as we can.
In order to expedite the process, it would be helpful to follow this checklist and provide
relevant information.

  * [ ] Operating system: `uname -a`, `./etc/lsb_release` or `cat /etc/redhat-release`
  * [ ] StackStorm version: `st2 --version`
  * [ ] Actual package versions of all packages (st2, st2web, st2chatops, nginx, mongo, rabbitmq-server)
    DEB: apt-cache policy ${PACKAGE_NAME} will give you the version of package.
    RPM: yum info ${PACKAGE_NAME} will you give the version of package.
    Note the exact name of mongo, nginx, rabbitmq changes based on OS.
  * [ ] [Contents of /etc/st2/st2.conf](url

)
  * [ ] Output of st2ctl status
  * [ ] Optional - Details about target box. E.g. vagrant box link or AWS AMI link.

#Issue details
my os is "Linux kcpm2 4.19.25-200.1.el7.bclinux.x86_64 #1 SMP Mon Aug 19 14:46:26 CST 2019 x86_64 x86_64 x86_64 GNU/Linux"
[root@kcpm2 /]# st2 --version
st2 3.3.0, on Python 2.7.5
![图片](https://user-images.githubusercontent.com/57519382/106985751-f81ed600-67a4-11eb-85eb-2a16e79288cd.png)
![图片](https://user-images.githubusercontent.com/57519382/106986086-957a0a00-67a5-11eb-999d-86fdac23b5e5.png)
![图片](https://user-images.githubusercontent.com/57519382/106986108-a32f8f80-67a5-11eb-82eb-08242d855b95.png)

armab:
Duplicate of https://github.com/StackStorm/st2/issues/3356 and https://github.com/StackStorm/st2/issues/2676

The fact that you have to configure host/port in 2 places: st2.conf and systemd service is definitely confusing.

We'd welcome any suggestions or contributions how to improve that.

luislobo:
Still cannot make it available on an Ubuntu 18.04. Where should I change the IPs?

Issue #495, opened 2017-09-07
Please run `systemctl daemon-reload` on update (systemd systems)


cognifloyd:
Originally reported as StackStorm/ansible-st2#169

I just upgraded a StackStorm install with from 2.3 to 2.4 and when I was manually looking at some issues with st2chatops, I got this:

```
# systemctl stop st2chatops
Warning: st2chatops.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

That suggests that the rpm upgrade did not run systemctl daemon-reload automatically, but it should.

```
jacob@harmony $ uname -a
Linux harmony 3.10.0-514.16.1.el7.x86_64 #1 SMP Wed Apr 12 15:04:24 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

jacob@harmony $ cat /etc/redhat-release
CentOS Linux release 7.3.1611 (Core) 

jacob@harmony $ st2 --version
st2 2.4.0

jacob@harmony $ yum info st2 st2web st2chatops st2mistral nginx mongodb-org rabbitmq-server postgresql-server
Loaded plugins: fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: mirror.us.oneandone.net
 * epel: mirror.rnet.missouri.edu
 * extras: centos.den.host-engine.com
 * nux-dextop: li.nux.ro
 * updates: mirror.dal10.us.leaseweb.net
Installed Packages
Name        : mongodb-org
Arch        : x86_64
Version     : 3.4.7
Release     : 1.el7
Size        : 0.0  
Repo        : installed
From repo   : mongodb-org-3.4
Summary     : MongoDB open source document-oriented database system (metapackage)
URL         : http://www.mongodb.org
License     : AGPL 3.0
Description : [snip]

Name        : nginx
Arch        : x86_64
Epoch       : 1
Version     : 1.12.1
Release     : 1.el7.ngx
Size        : 2.5 M
Repo        : installed
From repo   : nginx
Summary     : High performance web server
URL         : http://nginx.org/
License     : 2-clause BSD-like license
Description : [snip].

Name        : postgresql-server
Arch        : x86_64
Version     : 9.2.18
Release     : 1.el7
Size        : 16 M
Repo        : installed
From repo   : base
Summary     : The programs needed to create and run a PostgreSQL server
URL         : http://www.postgresql.org/
License     : PostgreSQL
Description : [snip]

Name        : rabbitmq-server
Arch        : noarch
[...]

PR #706, opened 2021-06-13
Use st2.conf to bind ip:port for st2auth, st2api and st2stream

nzlosh:
This pull request removes the static `.socket` units from debian and redhat packages and adds a shell based generator for `st2api`, `st2auth` and `st2stream` services.  The generator will read `st2.conf` using `crudini` to create the `.socket` file dynamically.

Fixes https://github.com/StackStorm/st2-packages/issues/686 https://github.com/StackStorm/st2/issues/3356 https://github.com/StackStorm/st2/issues/2676

There is an update to `st2ctl` that needs to be merged as part of this feature.  I'll update the issue to cross reference.


nzlosh:
This PR requires merge of https://github.com/StackStorm/st2/pull/5286 to be complete.

nzlosh:
I don't know why tests are failing on CircleCI, I run `docker-compose run --rm bionic` on my dev machine and the tests run end to end correctly.  My current guess is `st2ctl` isn't patched so the .socket files aren't being generated, but since there are no log messages as to what's failed during the CI process, it's anyones guess.

packages/st2/debian/st2api-generator
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+NORMAL_DIR=${1:-/tmp}
+EARLY_DIR=${2:-/tmp}
+LATE_DIR=${3:-/tmp}
+
+ST2SVC="st2api"
+DEFAULT_IP="127.0.0.1"
+DEFFALT_PORT="9101"
```
luislobo, line 9: Typo DEFFALT_PORT
Should be DEFAULT_PORT

packages/st2/debian/st2auth-generator
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+NORMAL_DIR=${1:-/tmp}
+EARLY_DIR=${2:-/tmp}
+LATE_DIR=${3:-/tmp}
+
+ST2SVC="st2auth"
+DEFAULT_IP="127.0.0.1"
+DEFFALT_PORT="9100"
```
luislobo, line 9: Same typo

packages/st2/debian/st2stream-generator
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+NORMAL_DIR=${1:-/tmp}
+EARLY_DIR=${2:-/tmp}
+LATE_DIR=${3:-/tmp}
+
+ST2SVC="st2stream"
+DEFAULT_IP="127.0.0.1"
+DEFFALT_PORT="9102"
+
```
luislobo, line 10: again typo

packages/st2/rpm/st2api-generator
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+NORMAL_DIR=${1:-/tmp}
+EARLY_DIR=${2:-/tmp}
+LATE_DIR=${3:-/tmp}
+
+ST2SVC="st2api"
+DEFAULT_IP="127.0.0.1"
+DEFFALT_PORT="9101"
```
luislobo, line 9: typo

luislobo (changes requested):
Thank you for your contribution. I'm one of the "affected" by this issue. 

I've found one typo repeated on several files (DEFFALT_PORT instead of DEFAULT_PORT). I'm very new at stackstorm so I don't feel confident to fully review everything works.

packages/st2/debian/st2api-generator
nzlosh: Good catch, thanks for pointing that out!

packages/st2/rpm/st2.spec
```diff
@@ -119,15 +124,15 @@ Conflicts: st2common
...
   %{_unitdir}/st2notifier.service
   %{_unitdir}/st2rulesengine.service
   %{_unitdir}/st2sensorcontainer.service
   %{_unitdir}/st2garbagecollector.service
   %{_unitdir}/st2timersengine.service
   %{_unitdir}/st2workflowengine.service
   %{_unitdir}/st2scheduler.service
+  /usr/lib/systemd/system-generators/st2api-generator
+  /usr/lib/systemd/system-generators/st2auth-generator
+  /usr/lib/systemd/system-generators/st2stream-generator
```
cognifloyd, line 138: CentOS8 has `%{_systemdgeneratordir}` but sadly, CentOS7 does not. So, this is good.

cognifloyd (approved):
Nice. Reducing hard-coded config is excellent!

rake/spec/default/60-st2_all-services-ok_spec.rb
```diff
@@ -72,7 +72,7 @@
 
 # all st2 services should work immediately after restart
 describe 'st2 services availability after restart' do
-  describe command("st2ctl restart && st2 action list") do
+  describe command("systemctl daemon-reload; st2ctl restart && st2 action list") do
```
armab, line 75: Can we add `systemctl daemon-reload` to postinst scriptlets?
And so this way the instruction won't be needed to be duplicated here or in any other installer like Ansilbe, Chef, st2ctl, etc.

Somewhat relevant https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950726

armab: Also, see https://github.com/StackStorm/st2-packages/issues/495 which could be also closed if we add it now.

armab (changes requested):
As mentioned before https://github.com/StackStorm/st2-packages/pull/706#discussion_r703713856,
we'll need to add systemctl daemon-reload to the packaging scriptlets.

That'll make the configuration automatic on install and would also fix the CI.

cognifloyd:
For CentOS, we need to add `EPEL` so installing the st2 rpms can find/install `crudini`. Should we do that in the docker images? Or should we do it in one of the scripts in this repo (maybe `scripts/install_os_packages.sh`)?

cognifloyd:
Also, I dropped the generator scripts under rpm because, as it turns out, we only need the debian copy of these generator scripts:
https://github.com/StackStorm/st2-packages/blob/82422e925d2f394c82e71832f147b9e2c6fcdf8d/rpmspec/helpers.spec#L28-L34

And `%default_install` is used in `rpm.spec` here:
https://github.com/StackStorm/st2-packages/blob/da36b56e315982b136660ae0f3276fb50a3966aa/packages/st2/rpm/st2.spec#L57-L58

`%default_install` installs things based on these files:
- `package/st2/debian/st2.dirs`
- `package/st2/debian/install` <- this already installs the generator scripts!
- `package/st2/debian/st2.links`

packages/st2/debian/control
```diff
@@ -15,7 +15,7 @@ Vcs-Browser: https://github.com/stackstorm/st2
 Package: st2
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.16), ${pre:Depends}, ${misc:Pre-Depends}, adduser
-Depends: ${shlibs:Depends}, ${misc:Depends}, sudo, adduser, ${Depends}, libssl-dev, libffi-dev, git, libpam0g, openssh-server, openssh-client, bash, netbase
+Depends: ${shlibs:Depends}, ${misc:Depends}, sudo, adduser, ${Depends}, libssl-dev, libffi-dev, git, libpam0g, openssh-server, openssh-client, bash, netbase, crudini
```
armab, line 18: Looks like `crudini` was needed in a earlier version of this PR when generators were bash-based.
In a new version of python I don't see where `crudini` is used, looks like could be removed from the dependencies.

cognifloyd: oh. good catch.

cognifloyd (approved):
Yeah! tests are passing now.

armab (commented):
Thanks, @nzlosh, and @cognifloyd. Nice work! :+1: 

I'll try the new experience end-to-end this week with the WIP packages from this PR.

The only thing I can think of that's missing is a Changelog record in the st2 repo mentioning the new feature.

packages/st2/debian/postinst
```diff
@@ -53,6 +53,7 @@ case "$1" in
       # Initially set destination files owenership (only on the first install)
       [ "$upgrading" = 1 ] || set_permissions "$SET_PERMS"
       rm -f $ST2_UPGRADESTAMP
+      systemctl daemon-reload
```
armab, line 56: would be nice to include a small code comment why we're doing this. Otherwise, it might hard to get a reason after some time.

armab (approved):
Great work! :+1: 

Just left a couple of minor cosmetic comments.

minsis:
I dont think this change was documented anywhere?

It took me a while to figure out why st2stream service wasn't picking up my .socket file anymore. Any searches to change the listen IP is still saying to update the .socket file directly.

In my case chef manages all my configs, including the deployment of the socket files. Anyways, maybe it should be documented somewhere in changelog or something so people upgrading to 3.6 know their socket files are irrelevant now.

nzlosh:
Good catch @minsis 

It was documented in the changelog https://github.com/StackStorm/st2/pull/5378/files but for some reason it's not reflected in the documentation.  @armab or @winem Can we generate the docs so people know about this change and can plan accordingly?

armab:
Changelog is in place, but looks like it would be nice to cover this important change in the st2docs `Upgrade Notes` as well:
https://docs.stackstorm.com/upgrade_notes.html#st2-v3-6

A PR would be welcome! Help wanted.

minsis:
@armab Do I make this PR into the v3.6 branch?

nzlosh:
@armab https://github.com/StackStorm/st2docs/pull/1098

armab:
Yeah, need to cherry-pick the change for v3.6 branch as well, besides of the master branch.

* `master` branch -> goes to development  https://docs.stackstorm.com/latest/
* `v3.6` branch -> goes to stable https://docs.stackstorm.com/3.6/ and https://docs.stackstorm.com/

we need it in both places

PR #708, opened 2021-10-10
Relax ini parsing to allow duplicate entries.

nzlosh:
Following failures in the end to end testing due to duplicate entries in the section, the config parser has been relaxed to allow duplicate entries.  The current default behaviour for duplicate keys is order of parsing so the last key read will be the value used.


arm4b (approved):
Thanks for the follow-up fix :+1:

cognifloyd (approved):

amanda11 (approved):
LGTM - I'd like to get this merged in before I deploy the updated st2cd package to the CI server, as this is a more long term solution.

PR #712, opened 2022-01-28
CircleCI build enhancements and fixes

armab:
CircleCI is showing some instability and different results in the packaging builds recently.
Experiment with the CircleCI environment.

* Update the outdated Docker environment
* Retry docker-pull on failure (helps with noisy build failures unable to download docker images)
* Invalidate the wheelhouse cache, where old cache might bring issues with the el8->rocky switch
* Fix corner case when the same wheelhouse cache/key could be used by all the distros as a fallback

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,9 +29,7 @@ jobs:
       - setup_remote_docker:
           reusable: true    # default - false
           exclusive: true   # default - true
-          # Temporary workaround for Circle CI issue
-          # https://discuss.circleci.com/t/setup-remote-docker-connection-failures/26434
-          version: 18.05.0-ce
+          version: 19.03.14
       - run:
           name: Ensure installation scripts are synced with their templates
           command: make .generated-files-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,8 +71,7 @@ jobs:
       - restore_cache:
           name: Restore Wheelhouse Cache
           keys:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -69,12 +67,11 @@ jobs:
             docker cp ~/st2-packages st2-packages-vol:/root
       - run:
           name: Pull dependent Docker Images
-          command: .circle/docker-compose2.sh pull ${DISTRO}
+          command: .circle/docker-compose2.sh pull ${DISTRO} || .circle/docker-compose2.sh pull ${DISTRO}
       - restore_cache:
           name: Restore Wheelhouse Cache
           keys:
-            - wheelhouse-v2-{{ .Branch }}-{{ checksum "/tmp/distro-version" }}-{{ checksum "/tmp/st2-version" }}
-            - wheelhouse-v2-{{ .Branch }}-
```
armab, line 77: this one is a potentially dangerous fallback as a cache produced by distro winning the first build race might be used then by all the future builds

amanda11 (approved):
Looks good to me.

armab:
Update: removed the wheelhouse cache to be consistent with the st2 repo environment: https://github.com/StackStorm/st2/blob/master/.circleci/config.yml

With wheelhouse cache (cache hit):
image

Without wheelhouse cache:
image

Looks like it just provides the 30sec speedup, potentially bringing corner cases with the complex configuration and cache/invalidation.

amanda11:
@armab Looks great, do you want to merge.

PR #713, opened 2022-01-28
Reorder in Makefile and restrict setuptools

amanda11:
Need to check rocky-release before centos-release as on Rocky systems, rocky-release is link to centos-release.
Also restrict setuptools when it gets installed on EL8 Makefile to < 60 as will fail until we upgrade python to 3.8.

### DIFF
--- a/packages/st2/component.makefile
+++ b/packages/st2/component.makefile
@@ -88,8 +88,7 @@ bdist_wheel: .stamp-bdist_wheel
 	cat requirements.txt
 # We need to install these python packages to handle rpmbuild 4.14 in EL8
 ifeq ($(EL_VERSION),8)
-        # Setuptools 60.x requires > python 3.6
-	$(PIP_BINARY) install wheel 'setuptools<60' virtualenv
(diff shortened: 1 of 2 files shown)


packages/st2/component.makefile
```diff
@@ -88,7 +88,8 @@ bdist_wheel: .stamp-bdist_wheel
 	cat requirements.txt
 # We need to install these python packages to handle rpmbuild 4.14 in EL8
 ifeq ($(EL_VERSION),8)
-	$(PIP_BINARY) install wheel setuptools virtualenv
+        # Setuptools 60.x requires > python 3.6
+	$(PIP_BINARY) install wheel 'setuptools<60' virtualenv
```
armab, line 92: As the builds are green, I think we don't need the pinning anymore.
Especially, if we're planning to switch EL8 to py3.8 in the upcoming PRs.

cognifloyd (approved):

armab (approved):
Thanks! :+1:

PR #714, opened 2022-02-28
Add overrides directory for pack overriding of meta data

amanda11:
Create overrides directory for overriding pack meta data, associated with https://github.com/StackStorm/st2/pull/5506


amanda11:
Build errors seem to be related to https://github.com/aws/aws-sam-cli/issues/3661, will try and fix versions to combat the error due to upgrade of MarkupSafe.

armab (approved):
Looks :+1: 

Would be also good to include a small test to check if dir created after the package install,
here's an example: 
https://github.com/StackStorm/st2-packages/blob/b2f0edb08e07f5fc0dbb7aa0e531a01567558b69/rake/spec/spec_helper.rb#L57-L61

PR #715, opened 2022-03-14
Initial py3.8 files

amanda11:
Update EL8 to use python 3.8


armab (approved):
Nice work handling the corner cases with the el8 env :+1:

PR #717, opened 2022-05-19
Use erlang < 25 on EL8

amanda11:
Modify El8 to use Erlang 25 to avoid issues with RabbitMQ 3.10 which only has preview support for erlang 25.

Partially resolves: https://github.com/StackStorm/st2/issues/5651

### DIFF
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -512,7 +512,7 @@ install_st2_dependencies() {
 install_rabbitmq() {
   # Install erlang from rabbitmq/erlang as need newer version
   # than available in epel.
-  # Use Erlang 24 has Erlang 25 has only preview support with RabbitMQ 3.10
+  # Use Erlang 24 as RabbitMQ 3.10 only has preview support for Erlang 25.
   curl -sL https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | sudo bash
   sudo yum -y install erlang-24*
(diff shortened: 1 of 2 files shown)


scripts/st2bootstrap-el8.sh
```diff
@@ -511,9 +511,10 @@ install_st2_dependencies() {
 
 install_rabbitmq() {
   # Install erlang from rabbitmq/erlang as need newer version
-  # than available in epel
+  # than available in epel.
+  # Use Erlang 24 has Erlang 25 has only preview support with RabbitMQ 3.10
```
cognifloyd, line 515:
```suggestion
  # Use Erlang 24 as RabbitMQ 3.10 only has preview support for Erlang 25.
```

scripts/st2bootstrap-el8.template.sh
```diff
@@ -147,9 +147,10 @@ install_st2_dependencies() {
 
 install_rabbitmq() {
   # Install erlang from rabbitmq/erlang as need newer version
-  # than available in epel
+  # than available in epel.
+  # Use Erlang 24 has Erlang 25 has only preview support with RabbitMQ 3.10
```
cognifloyd, line 151:
```suggestion
  # Use Erlang 24 as RabbitMQ 3.10 only has preview support for Erlang 25.
```

cognifloyd (approved):
comment nitpicks. Otherwise looks good.

PR #718, opened 2022-05-23
Use erlang < 25 on EL8

amanda11:
Update Bash installer for 3.7 to use erlang24*

RabbitMQ have released a newer version 3.10.2 which now manages to start with erlang 25. But they still report on https://www.rabbitmq.com/which-erlang.html that the erlang 25 support is preview. 
Therefore probably safer to stick with using 24*

### DIFF
--- a/scripts/st2bootstrap-el8.sh
+++ b/scripts/st2bootstrap-el8.sh
@@ -511,9 +511,10 @@ install_st2_dependencies() {
 
 install_rabbitmq() {
   # Install erlang from rabbitmq/erlang as need newer version
-  # than available in epel
+  # than available in epel.
+  # Use Erlang 24 has Erlang 25 has only preview support with RabbitMQ 3.10
   curl -sL https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | sudo bash
-  sudo yum -y install erlang
+  sudo yum -y install erlang-24*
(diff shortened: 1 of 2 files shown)


amanda11:
@armab I've cherrypicked the st2-packages for the 3.7 branch in this PR, but I'd appreciate your check as to process to what we need to do to cherry pick for 3.7. We shouldn't need new packages built as its just the bash installer we're updateing.

As per @bishopbm1 comment in slack, the new RabbitMQ 3.10.2 will start with erlang 24 so even without this change we should be able to install ok now. But given that erlang 25 is "preview" support, would it still be better to use 24.* on erlang anyway.

armab (approved):
Thanks!

armab:
The stable branch is referenced in the `st2_bootstrap.sh` script which is an entry point for the `curl|bash` installer.
https://github.com/StackStorm/st2-packages/blob/ed1cfe29a6202c444a5626773bdf973e22605b92/scripts/st2_bootstrap.sh#L18-L19

From that, the installer will use the `v3.7` branch here:
https://github.com/StackStorm/st2-packages/blob/ed1cfe29a6202c444a5626773bdf973e22605b92/scripts/st2_bootstrap.sh#L166

cognifloyd (approved):
Yeah, let's stick with erlang 24 for now.

PR #719, opened 2022-06-06
Use Erlang 24 on Ubuntu

amanda11:
Erlang 25 is now available on Debian RabbitMQ repositories, but current RabbitMQ only has preview support for Erlang 25.

Move to CloudSmith so that Erlang 24 can be picked up, as not available at LaunchPad.

Use pin preferences to make Erlang 24 be installed in preference to Erlang 25.


nzlosh (approved):
LGTM

cognifloyd (approved):

PR #721, opened 2022-07-15
generate unit file for st2api, st2auth and st2stream

nzlosh:
Move api, auth and stream service unit files to generator so st2.conf variables are used consistently for .socket and .service files.


packages/st2/debian/st2auth-generator
```diff
@@ -4,9 +4,9 @@ import logging
 import time
 import sys
 
-ST2SVC="st2auth"
-DEFAULT_IP="127.0.0.1"
-DEFAULT_PORT="9100"
+ST2SVC = "st2auth"
+DEFAULT_IP = "127.0.0.1"
+DEFAULT_PORT = "9101"
```
cognifloyd, line 9:
```suggestion
DEFAULT_PORT = "9100"
```

packages/st2/debian/st2stream-generator
```diff
@@ -4,9 +4,9 @@ import logging
 import time
 import sys
 
-ST2SVC="st2stream"
-DEFAULT_IP="127.0.0.1"
-DEFAULT_PORT="9102"
+ST2SVC = "st2stream"
+DEFAULT_IP = "127.0.0.1"
+DEFAULT_PORT = "9101"
```
cognifloyd, line 9:
```suggestion
DEFAULT_PORT = "9102"
```

cognifloyd (changes requested):
Looks like a bit of copy pasta. I left suggestions to fix the ports.

Also, I wonder if we could combine these so one generator handles all 3 services. But it's probably not worth the effort.

cognifloyd (approved):

packages/st2/debian/st2api.service
```diff
@@ -1,19 +0,0 @@
...
-Description=StackStorm service st2api
-After=network.target st2api.socket
-Requires=st2api.socket
-
-[Service]
-Type=simple
-User=st2
-Group=st2
-Environment="DAEMON_ARGS=-k eventlet -b 127.0.0.1:9101 --workers 1 --threads 1 --graceful-timeout 10 --timeout 30 --log-config /etc/st2/logging.api.gunicorn.conf --error-logfile /var/log/st2/st2api.log"
-EnvironmentFile=-/etc/default/st2api
```
cognifloyd, line 11: I just noticed that the generator does not include `EnvironmentFile` any more.

One good thing, is that makes the generator truly agnostic between debian vs EL.

One potential gotcha is that if anyone is relying on this, then that file will no longer be sourced. They can, however, add `EnvironmentFile` to a systemd service override file `/etc/systemd/system/st2api.conf.d/some-file-name-here.conf`. That can also be used to override anything else if desired. (The same applies to st2auth and st2stream)

nzlosh: As far as I know, the `/etc/default/st2api` configuration file is how Debian based distributions allowed overriding sysv and upstart service scripts.  As you've rightly pointed out, people should use the built-in systemd override mechanism and all currently supported OS' are systemd based.  We can highlight this in the v3.8 upgrade notes.

packages/st2/rpm/postinst_script.spec
```diff
@@ -1,4 +1,8 @@
 set -e
 
-# make sure that our socket generators run
+# Reload systemd to run generators for unit and socket files.
 systemctl daemon-reload >/dev/null 2>&1 || true
+
+# Enable services created by systemd generator
+systemctl enable st2api st2auth st2stream >/dev/null 2>&1  || true
+systemctl start st2api st2auth st2stream >/dev/null 2>&1  || true
```
cognifloyd, line 8: Looking at the test error for EL distros:

I think this is handled by the `%post` macro which uses the `%service_post` helper. That runs:
https://github.com/StackStorm/st2-packages/blob/23f100cbe709678cbdae81055b8c7dd244c2b883/rpmspec/helpers.spec#L86

What if you remove the output redirection in the `%service_post` macro helper? Then maybe we'll see if there's an error. And maybe remove it here too.

cognifloyd (commented):
I'm pretty sure tests are failing because this line is having an error that has been silenced:
https://github.com/StackStorm/st2-packages/blob/75a0e5e06fc911c932da616aefac2065c4f38970/packages/st2/rpm/st2.spec#L89

That line lists the 6 services that aren't getting enabled correctly.

packages/st2/rpm/postinst_script.spec
nzlosh: There appears to be a worker timeout for gunicorn on EL7, but I don't think that's the cause of the services failing to start for the RPM based distros.  I still don't see any errors in the logs that indicate why the service is not enabled and started.

cognifloyd:
aargh. There's still no helpful output during `yum install`. I don't know which verbosity flag(s) to use ([`--verbose`](https://www.mankier.com/8/yum#--verbose) or [`--rpmverbosity`](https://www.mankier.com/8/yum#--rpmverbosity) or both?), but what if we add some verbose settings on this line:

https://github.com/StackStorm/st2-packages/blob/6fba8554f6a46992db49f7bc6a2a7b429005d1d8/scripts/install_os_packages.sh#L15

`scripts/install_os_packages.sh` gets called in the package tests here:
https://github.com/StackStorm/st2-packages/blob/739d24ec47d1889670e57581c482a5bbcf8a60a8/rake/build/setup.rake#L13

We need to figure out why the `%post` bits aren't running like we expect. ☹️

nzlosh:
I'm going to debug this on a VM, pushing and waiting for the CI to fail is a hindrance and unreasonably slow.

PR #724, opened 2022-09-29
Use erlang 25

amanda11:
Investigate if install will work with erlang 25 on all repos.


armab (approved):
Thanks for the PR :+1:

PR #728, opened 2023-10-27
Fix EL7 build for Orquesta v1.6.0 update with 'python_version' markers

armab:
pip resolver failed processing the dependencies when using the `python_version` dependent environment markers in orquesta requirements.txt https://github.com/StackStorm/orquesta/pull/255.

Build failure: https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940?invite=true#step-109-978547_123

Relying on `--use-deprecated=legacy-resolver` masks the problem and makes the build succeed.

See https://github.com/StackStorm/st2/pull/6050 for more context.

### DIFF
--- a/rpmspec/package_venv.spec
+++ b/rpmspec/package_venv.spec
@@ -17,7 +17,7 @@
 
 %define venv_python %{venv_bin}/%{python_binname}
 # https://github.com/StackStorm/st2/wiki/Where-all-to-update-pip-and-or-virtualenv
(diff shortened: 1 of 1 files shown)


armab:
This PR is RFR, but will hold with merging before we fix the e2e build failures.

cognifloyd (approved):

PR #733, opened 2023-12-12
Install net-tools automatically on deb-distributions

winem:
We already install net-tools on CentOS and Rocky Linux. Now, we also do it for installations using the deb-installer.  The package manager will detect on it's own if it's actually needed to install or update the net-tools or if the latest version is already installed and it does not have to do anything else.

This also addresses an error when running the installer without having netstat available on Ubuntu 20.04 for example: 
```
20231212T161937+0000 ################################################################
20231212T161937+0000 ### Installing from staging repos!!! USE AT YOUR OWN RISK!!! ###
20231212T161937+0000 ################################################################
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 sudo: netstat: command not found
20231212T161937+0000 Hit:1 http://mirrors.digitalocean.com/ubuntu focal InRelease
20231212T161937+0000 Hit:2 http://mirrors.digitalocean.com/ubuntu focal-updates InRelease
```

### DIFF
--- a/scripts/st2bootstrap-deb.sh
+++ b/scripts/st2bootstrap-deb.sh
@@ -430,6 +430,10 @@ fail() {
 }
 
 
+install_net_tools() {
+  # Install netstat
+  sudo apt install -y net-tools
+}
 
 install_st2_dependencies() {
   # Silence debconf prompt, raised during some dep installations. This will be passed to sudo via 'env_keep'.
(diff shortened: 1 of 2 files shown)


armab (approved):
Good find! Thanks 👍

PR #743, opened 2024-02-19
Add Jammy

### LINKED ISSUE #737: Add Ubuntu 22.04 support

mamercad:
Fixes #737, needs [#123](https://github.com/StackStorm/st2packaging-dockerfiles/pull/123).

### DIFF
--- a/packages/st2/debian/rules
+++ b/packages/st2/debian/rules
@@ -20,10 +20,7 @@ BUILD_PRE_DEPENDS := python3 (>= 3.8)
 BUILD_DEPENDS := python3-distutils, python3-dev
 
 %:
-
-ifeq ($(DEB_DISTRO),focal)
-	dh $@ --with python-virtualenv --python /usr/bin/python3.8
-endif
+	dh $@ --with python-virtualenv --python /usr/bin/python3
 
 override_dh_installdirs:
 	dh_installdirs


Vagrantfile
```diff
@@ -5,6 +5,11 @@
 VAGRANTFILE_API_VERSION = "2"
 
 VIRTUAL_MACHINES = {
+  :jammy => {
+    :hostname => 'st2-packages-jammy',
+    :box => 'ubuntu/jammy64',
+    :ip => '192.168.16.26',
```
cognifloyd, line 11:
```suggestion
    :ip => '192.168.16.27',
```
The IPs should be unique.

cognifloyd (commented):
It looks like CI is trying to use Python 3.10 on Ubuntu Jammy. I don't think we've adjusted requirements or the code to support that yet. Can we get python 3.9 installed?

mamercad:
> It looks like CI is trying to use Python 3.10 on Ubuntu Jammy. I don't think we've adjusted requirements or the code to support that yet. Can we get python 3.9 installed?

With Jammy, Python 3.10.12 is in base:

```shell
root@bf0289b2692f:/# grep VERSION= /etc/os-release 
VERSION="22.04.4 LTS (Jammy Jellyfish)"

root@bf0289b2692f:/# python3 -V
Python 3.10.12
```

Are you saying that we should support Python 3.9 in Jammy?

nzlosh (approved):

PR #744, opened 2024-02-22
Add rocky9 support and update to using docker compose v2

nzlosh:


README.md
```diff
@@ -106,7 +106,7 @@ Current community packages are hosted on https://packagecloud.io/StackStorm. For
 
 If you are adding support for a new distribution for which ``packagingbuild`` and ``packagingtest``
 images are not yet published to Docker Hub and you want to test the build pipeline locally, you
-need to update ``docker-compose.yml`` file to use locally built Docker images.
+need to update ``docker compose.yml`` file to use locally built Docker images.
```
mamercad, line 109: ``docker-compose.yml``

mamercad (changes requested):
Small search and replace problem, other than that, excited that we're getting there for new OSs.

README.md
nzlosh: Oh good spotting.  The entire sentence should be re-written.  docker-compose is v1 and is no longer functional with circle ci deprecating docker <1.20.  The docker compose plugin is required for v2 support, which available from the docker repository.

mamercad (approved):
🚢

PR #747, opened 2024-10-07
Fix package build for rocky9

nzlosh:
Use the same pip and cryptograph version as st2 project for both rocky 8 and 9.


nzlosh:
It seems `python3-virtualenv` (unavailable on Rocky9 https://bugs.almalinux.org/view.php?id=281) was causing the exception in `venvctrl`.  Moving off `virtualenv` to `python -m venv` allowed `venvctrl` to work correctly.

As a side note, Debian is unaffected because it uses `dh-virtualenv` to handle python virtual environments during packaging.  However, `dh-virtualenv` looks like it's no longer maintained (last update 3rd Jan 2022), so we might have to investigate using `venvctrl` for deb based solutions too.

cognifloyd (approved):

PR #749, opened 2024-10-26
Fix circleci deployment pipeline

nzlosh:
- format shell scripts with 4 spaces.
- use sed regex to extract package metadata from file.
- update distribution lists for package processing.
- use regex for flexible matching in packagecloud metadata package version.


rush-skills (approved):

rush-skills (approved):
LGTM

cognifloyd (approved):
Thank you!

PR #750, opened 2025-02-04
Following the guidelines in the file, refreshing this

skiedude:
The dist_utils.py in this repo according to the comments should be a near copy of the one in the st2 repo. I've copied it over (removed some comments specific to the st2 version) so that it can also stop breaking on requirements.txt that use the `;` to separate python markers.

This should fix the circleci tests on https://github.com/StackStorm/st2/pull/6296 and future PRs once merged.

This is a bit of a catch22 situation.

This build pulls master of st2, which does not have my dist_utils fixes yet, so it fails. The st2 pulls the current master of this repo which doesn't have my changes yet and fails.

One of these probably has to be forced merged so the other tests with the changes?

### DIFF
--- a/packages/st2/dist_utils.py
+++ b/packages/st2/dist_utils.py
@@ -2,6 +2,7 @@
 # NOTE: This file is auto-generated - DO NOT EDIT MANUALLY
 #       Instead copy from https://github.com/StackStorm/st2/blob/master/scripts/dist_utils.py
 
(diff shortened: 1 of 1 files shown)


nzlosh (approved):

nzlosh:
Applying patch with circleci failure due to circular dependencies between st2-packages and st2 repos.

PR #752, opened 2025-05-16
Update bootstrap scripts for supported operating systems for v3.9

nzlosh:


scripts/st2_bootstrap.sh
```diff
@@ -115,82 +135,67 @@ if [[ "$REPO_TYPE" == 'staging' ]]; then
...
-    echo "Unsupported ubuntu codename ${SUBTYPE}. Please use Ubuntu 20.04 (focal) as base system!"
-    exit 2
-  fi
-  ST2BOOTSTRAP="${BASE_PATH}/${BRANCH}/scripts/st2bootstrap-deb.sh"
-  BOOTSTRAP_FILE="st2bootstrap-deb.sh"
+echo "*** Detected distribution is ${OS_PRETTY_NAME} ***"
+if is_rpm; then
+    # Rocky Linux versions 8.x or 9.x
+    MAJOR_VERSION=$(cut -d. -f1 <<<"$OS_VERSION_ID")
+    if [[ ! "$MAJOR_VERSION" =~ ^[89] ]]; then
```
amanda11, line 176: Ist2 bootstrap used for all versions, rather than being version specific.
So it has to support all the o/s versions that stackstorm support, as it's the common thread before it call the version specific bootstrap calls.

So as stackstorm 3.8 is still current, and unless dropped the usual premise as we support the last 2 stackstorm versions, it must still support all the o/s of 3.8 and 3.9.

So we can't drop redhat 7 yet from the list, and same for any Ubuntu's that 3.8 supports.

nzlosh: You are quite right, this PR is intended for use with the st2 3.9 release.  I'll mark it as draft so it won't be merged until the release process is started, which should be this week or the next.

amanda11: It's more than that. You still need to keep the o/s versions in here that 3.8 supports until stackstorm drops support for 3.8. As this file is the installer that is used for all supported stackstorm versions. 

You will get the master version when you download the installer for ANY stackstorm versions. So it always has to keep the o/s versions of latest and previous version.

amanda11: I am just pointing out how it works now. I agree it would be better to do what you say, but you will need to update the documents for old and new, and to update the e2e tests as well that use that link in the e2e runs if you want to do that.
So unless you do those changes first, then this has to be generic for all.
It would also mean that any users who may have their own automations that use our current method would need to be given advance warning thay they can't use the same method as they currently use.

cognifloyd (approved):

PR #758, opened 2026-06-23
pip and python changes for 3.10

guzzijones:


nzlosh (approved):

guzzijones:
circle ci expects 4 builds.  I will have to see what it takes to add `focal` and `el10`. for now i have put `jammy jammy el9 el9`

nzlosh:
Set the parallelism to 2 if there are only two to make.  I'll revert the debug distro changes that were merged without review so we can actually discuss what's being changed and why.

